PR #24, opened 2017-05-10
Fix broken tests

kusma:
This fixes the broken tests after updating uno and merging release-1.0


kusma:
This fixes the broken master after updating uno.

PR #25, opened 2017-05-10
Fuse.Reactive.JavaScript: remove Duktape support

kusma:
This JavaScript VM has not been in active use for a long time, and
is unmaintained and not actively tested. So let's get rid of it.


kusma:
This is just a proposal. It quite drastically reduce the size of our working-directories, where Duktape is currently the bulk of the code-base, as 41% of the code.

On the other hand, Duktape is a very easy to bring up JavaScript VM, so perhaps it'd be useful for bringing up other platforms if needed...

ollef:
I've tried to keep the Duktape code up-to-date with the other JS VMs since it's useful as a small but slow alternative to V8 and possible to turn on with a flag, so I'm a bit reluctant to throw it away. Last I checked it worked OK. If fuselibs code size is the issue we could perhaps consider shipping the external code some other way.

`Fuse.Scripting.Jurassic`, on the other hand, is not maintained AFAIK. I'd be surprised if it worked at all.

kusma:
@ollef: OK, then I think we need to start running tests using `-DUSE_DUKTAPE`. In its current form, it's easy to regress something.

ollef:
Yeah, that's a good idea. I know there's at least one failing test already due to limited unicode support in Duktape, but perhaps that can be conditionally ignored.

kusma:
@ollef: Yeah, we can conditionally ignore based on defines, so that's no problem. But perhaps it's also time to consider updating the duktape code (or if there's no working update, perhaps we can fix it)...?

ollef:
@kusma: Yeah, it looks like there has been a few Duktape releases since we last updated so that would be good to do.

kusma:
OK, closing this in favor of #46

PR #29, opened 2017-05-11
travis: use builtin ccache-caching

kusma:
This is just an experiment, I'm having some problems building on the public TravisCI instance.


kusma:
This didn't pan out.

PR #30, opened 2017-05-11
DotNet support for base64 operations in ImageTools 

eeue56:
Now this will work on DotNet platforms: 

```
<App>
	<JavaScript>
		var ImageTools = require("FuseJS/ImageTools");
		var Observable = require("FuseJS/Observable");

		var imagePath = Observable();
		var base64Image =	"iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l" +
							"EQVQI12MAAoMHIFLAAYSEwIiJgYGZASrI38AAAwBamgM5VF7xgwAAAABJRU5ErkJggg==";
		ImageTools.getImageFromBase64(base64Image)
		.then(function(image) {
			imagePath.value = image.path;
			ImageTools.getBase64FromImage(image).then(function(b64){
				console.log(b64);
			});
		});

		module.exports = { test: new Date().toString(), image: imagePath };
	</JavaScript>
	<Image File="{image}" />
</App>
```

Previously this was unsupported. This is a first PR to add full DotNet compatibility for image tools

### DIFF
--- a/Source/Fuse.ImageTools/Fuse.ImageTools.unoproj
+++ b/Source/Fuse.ImageTools/Fuse.ImageTools.unoproj
@@ -9,7 +9,10 @@
   },
   "InternalsVisibleTo":[
     "Fuse.Camera",
-    "Fuse.CameraRoll"
+    "Fuse.CameraRoll",
+    "FuseTest",
+    "Fuse.Testing",
+    "Fuse.ImageTools.Test"
   ],
   "Packages": [
     "Uno.Threading",
@@ -32,5 +35,8 @@
     "iOS/ImagePicker.m:CSource:iOS",
     "iOS/ImageHelper.h:CHeader:iOS",
     "iOS/ImageHelper.m:CSource:iOS"
+  ],
+  "Excludes": [
+    "Tests/",
   ]
 }
--- a/Source/Fuse.ImageTools/Fuse.ImageTools.unoproj
+++ b/Source/Fuse.ImageTools/Fuse.ImageTools.unoproj
@@ -10,8 +10,6 @@
   "InternalsVisibleTo":[
     "Fuse.Camera",
     "Fuse.CameraRoll",
-    "FuseTest",
-    "Fuse.Testing",
     "Fuse.ImageTools.Test"
   ],
   "Packages": [
--- a/Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
+++ b/Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
@@ -1,0 +1,68 @@
+using Uno;
+using Uno.Collections;
+using Uno.Testing;
+using Fuse;
+using FuseTest;
+
+namespace Fuse.ImageTools.Test
+{
+	public class ImageToolsTest : TestBase
+	{
+		private Object _image = null;
+
+
+		[Test]
+		public void CreatingAnImageSucceeds()
+		{
+			if defined(!DOTNET) return;
+
+			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
+			using (var root = new TestRootPanel()){
+				var future = ImageTools.ImageFromBase64("iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l" + "EQVQI12MAAoMHIFLAAYSEwIiJgYGZASrI38AAAwBamgM5VF7xgwAAAABJRU5ErkJggg==");
+				future.Then(Print, Fail);
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+
+				Assert.IsFalse(_image == null);
+			}
+			_image = null;
+		}
+
+		[Test]
+		public void CreatingAnImageFails()
+		{
+			if defined(!DOTNET) return;
+			
+			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
+			using (var root = new TestRootPanel()){
+				try 
+				{
+					var future = ImageTools.ImageFromBase64("iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l");
+					future.Then(Print, Fail);
+				} catch (FormatException e)
+				{
+					// expected error case
+				}
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+
+				Assert.IsTrue(_image == null);
+			}
+			_image = null;
+		}
+
+		private void Fail(Exception e)
+		{
+			_image = null;
+		}
+
+		private void Print(Object image)
+		{
+			_image = image;
+		}
+	}
+}
--- a/Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
+++ b/Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
@@ -10,49 +10,31 @@ namespace Fuse.ImageTools.Test
 	{
 		private Object _image = null;
 
-
 		[Test]
+		[Ignore("Only supported on dotnet", "!DOTNET")]
 		public void CreatingAnImageSucceeds()
 		{
-			if defined(!DOTNET) return;
-
 			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
 			using (var root = new TestRootPanel()){
 				var future = ImageTools.ImageFromBase64("iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l" + "EQVQI12MAAoMHIFLAAYSEwIiJgYGZASrI38AAAwBamgM5VF7xgwAAAABJRU5ErkJggg==");
 				future.Then(Print, Fail);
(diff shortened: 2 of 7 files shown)


kusma:
@eeue56: maybe squash the last three commits?

eeue56:
@kusma done!

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -0,0 +1,68 @@
...
+			
+			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
+			using (var root = new TestRootPanel()){
+				try 
+				{
+					var future = ImageTools.ImageFromBase64("iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l");
+					future.Then(Print, Fail);
+				} catch (FormatException e)
+				{
+					// expected error case
```
kusma: We use `Assert.Throws` for this.

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -0,0 +1,68 @@
...
+				}
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+
+				Assert.IsTrue(_image == null);
```
kusma: If we expect an error, I don't think we need any of this StepFrame-stuff.

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -0,0 +1,68 @@
...
+{
+	public class ImageToolsTest : TestBase
+	{
+		private Object _image = null;
+
+
+		[Test]
+		public void CreatingAnImageSucceeds()
+		{
+			if defined(!DOTNET) return;
```
kusma: use `[Ignore("Some explanation here", "!DOTNET")]` instead

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -0,0 +1,68 @@
...
+
+				Assert.IsFalse(_image == null);
+			}
+			_image = null;
+		}
+
+		[Test]
+		public void CreatingAnImageFails()
+		{
+			if defined(!DOTNET) return;
```
kusma: use `[Ignore("Some explanation here", "!DOTNET")]` instead

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -0,0 +1,68 @@
...
+
+			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
+			using (var root = new TestRootPanel()){
+				var future = ImageTools.ImageFromBase64("iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l" + "EQVQI12MAAoMHIFLAAYSEwIiJgYGZASrI38AAAwBamgM5VF7xgwAAAABJRU5ErkJggg==");
+				future.Then(Print, Fail);
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
+				root.StepFrame();
```
kusma: If you make a closure-object instead, we can step until we either fail, or pass here instead, so we don't need to know the timing...

Source/Fuse.ImageTools/Fuse.ImageTools.unoproj
```diff
@@ -9,7 +9,10 @@
   },
   "InternalsVisibleTo":[
     "Fuse.Camera",
-    "Fuse.CameraRoll"
+    "Fuse.CameraRoll",
+    "FuseTest",
+    "Fuse.Testing",
```
kusma: Why are those two needed?

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -10,49 +10,31 @@ namespace Fuse.ImageTools.Test
...
-
 			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
 			using (var root = new TestRootPanel()){
 				var future = ImageTools.ImageFromBase64("iVBORw0KGgoAAAANSUhEUgAAAA8AAAAPAQMAAAABGAcJAAAABlBMVEX9//wAAQATpOzaAAAAH0l" + "EQVQI12MAAoMHIFLAAYSEwIiJgYGZASrI38AAAwBamgM5VF7xgwAAAABJRU5ErkJggg==");
 				future.Then(Print, Fail);
-				root.StepFrame();
-				root.StepFrame();
-				root.StepFrame();
-				root.StepFrame();
-
```
kusma: Huh, is this safe?

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -10,49 +10,31 @@ namespace Fuse.ImageTools.Test
...
 				Assert.IsFalse(_image == null);
```
kusma: This just proves that `void Fail(Exception e)` hasn't been called (potentially... yet)

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -10,49 +10,31 @@ namespace Fuse.ImageTools.Test
...
 		private Object _image = null;
 
-
 		[Test]
+		[Ignore("Only supported on dotnet", "!DOTNET")]
 		public void CreatingAnImageSucceeds()
 		{
-			if defined(!DOTNET) return;
-
 			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
```
kusma: This is unused, no?

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -10,49 +10,31 @@ namespace Fuse.ImageTools.Test
...
 			_image = null;
 		}
 
-		[Test]
-		public void CreatingAnImageFails()
+		void MakeInvalidImage()
 		{
-			if defined(!DOTNET) return;
-			
 			Fuse.ImageTools.ImageTools image = new Fuse.ImageTools.ImageTools();
```
kusma: Same as above, this is unused, no?

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -10,49 +10,31 @@ namespace Fuse.ImageTools.Test
 	{
 		private Object _image = null;
 
-
 		[Test]
+		[Ignore("Only supported on dotnet", "!DOTNET")]
```
kusma: How can this be? `ImageFromBase64` does `closure.Reject("Unsupported platform");` when we're not on Android nor iOS, as far as I can tell...

eeue56: Do we run Uno tests on Android and iOS too? I thought it was only DotNet + Native?

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
eeue56: Yep

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
kusma: @eeue56: We currently have to run them manually.

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
```diff
@@ -0,0 +1,48 @@
...
+using FuseTest;
+
+namespace Fuse.ImageTools.Test
+{
+	public class ImageToolsTest : TestBase
+	{
+		private Object _image = null;
+
+		[Test]
+		[Ignore("Only supported on dotnet, Android or iOS", "native")]
```
kusma: Now the comment and condition are a mismatch.... "native" means "CMake" or "MSVC" on macOS and Windows respectively.

I think what you want is `[Ignore("Only supported on mobile platforms", "MOBILE")]`

eeue56: Possibly, a little confused 😅  This is supported on DotNet + Android + iOS, not just mobile. So it should be `"MOBILE || DOTNET"` I suppose? It's currently not supported on CMake nor MSVC - it's supported on every other target

kusma: Yeah, right. I just looked at the current version of the implementation. YEah. `MOBILE || DOTNET` seems correct

kusma: Oh, wait. Maybe "Native" would be right. We want to *ignore* on native...

eeue56: So it's good like this?

kusma: @eeue56: Yeah, I think so.

Source/Fuse.ImageTools/Tests/ImageTools.Test.uno
kusma: Ah, you're right.

PR #31, opened 2017-05-11
Feature toggle for implicit GraphicsView

nekronos:
While profiling apps using `NativeViewHost` in the root I observed that a significant amount of cpu time was spent in [`EndDraw`](https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Controls.Native/iOS/GraphicsView.uno#L108-L112). I guess we are stuck waiting for swap buffers.

Disabling the implicit `GraphicsView` yielded visible performance improvements on both iOS and Android.

This feature is in our backlog and will at some point be an option in `unoproj`

### DIFF
--- a/Source/Fuse.iOS/iOSApp.uno
+++ b/Source/Fuse.iOS/iOSApp.uno
@@ -19,7 +19,21 @@ namespace Fuse
 		}
 
 		TreeRendererPanel _renderPanel;
-		Fuse.Controls.GraphicsView _graphicsView;
+
+		extern(!DISABLE_IMPLICIT_GRAPHICSVIEW)
+		Fuse.Controls.GraphicsView _graphicsView = new Fuse.Controls.GraphicsView();
+
+		Visual RootVisual
+		{
+			get
+			{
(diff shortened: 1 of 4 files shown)


Source/Fuse.iOS/iOSApp.uno
```diff
@@ -126,7 +141,7 @@ namespace Fuse
  			if (_prevStatusBarOrientation != o)
  			{
  				_prevStatusBarOrientation = o;
-				UpdateManager.PerformNextFrame(_graphicsView.InvalidateVisual);
+				UpdateManager.PerformNextFrame(RootVisual.InvalidateVisual);
```
kusma: I think this regress the iOS screen-rotation stuff, doesn't it?

nekronos: No, but I changed to code to make it clearer https://github.com/fusetools/fuselibs-public/pull/31/commits/47bc5bf8444de9db6396d071e35f94af1f003249#diff-4146b1a7d97819d8c81051849608276fR144

PR #38, opened 2017-05-16
Fix for Atlas issue

mortoray:
fixes https://github.com/fusetools/fuselibs/issues/4176

All I did was use the full transform to translate the points of the atlas item. This seems like the right thing to do, but it seems unlikely this was forgotten previously -- am I missing something?  I'm also uncertain how the previous items managed to be rotated at all without this transform multiplication.

This will also break if there are non-flat transforms, but I think the caching excludes those anyway (though it doesn't really have to, since the element's visuals are still flat). Note, this would be broken before as well, so no further problem here.

### DIFF
--- a/Source/Fuse.Elements/Caching/ElementBatch.uno
+++ b/Source/Fuse.Elements/Caching/ElementBatch.uno
@@ -321,7 +321,9 @@ namespace Fuse.Elements
 				var opacity = entry._opacity;
 
 				var transform = entry._elm.LocalTransform;
-				float2 positionOrigin = transform[3].XY + (float2)cachingRect.Minimum * densityScale;
+				//this calculation assumes the transform is flat (a precondition to caching the element)
+				float2 localOrigin = (float2)cachingRect.Minimum * densityScale;
+				float2 positionOrigin = transform[3].XY + localOrigin.X * transform[0].XY + localOrigin.Y * transform[1].XY;
 				float2 size = (float2)cachingRect.Size * densityScale;
 				float2 right = transform[0].XY * size.X;
 				float2 up = transform[1].XY * size.Y;


kusma:
@mortoray: The old code was rotated because it picked out the principal vectors of the transform (which are already rotated). I don't at first glance understand why this helps, and the code does look a bit less efficient. Hmm...

Source/Fuse.Elements/Caching/ElementBatch.uno
```diff
@@ -321,14 +321,16 @@ namespace Fuse.Elements
...
+				var org = (float2)cachingRect.Minimum * densityScale;
+				var sz = (float2)cachingRect.Size * densityScale;
+				var ul = Vector.Transform( org, transform );
+				var ur = Vector.Transform( org + float2(sz.X,0), transform );
+				var br = Vector.Transform( org + sz, transform );
+				var bl = Vector.Transform( org + float2(0,sz.Y), transform);
+				vertexPositions.Set((i * 4 + 0) * _positionInfo.BufferStride + _positionInfo.BufferOffset, float3(ul.XY, opacity));
+				vertexPositions.Set((i * 4 + 1) * _positionInfo.BufferStride + _positionInfo.BufferOffset, float3(ur.XY, opacity));
+				vertexPositions.Set((i * 4 + 2) * _positionInfo.BufferStride + _positionInfo.BufferOffset, float3(br.XY, opacity));
+				vertexPositions.Set((i * 4 + 3) * _positionInfo.BufferStride + _positionInfo.BufferOffset, float3(bl.XY, opacity));
```
kusma: Looks like the ".XY"-swizzles here are needless.

kusma:
@mortoray: Is this sufficient to fix the problem?
```patch
diff --git a/Source/Fuse.Elements/Caching/ElementBatch.uno b/Source/Fuse.Elements/Caching/ElementBatch.uno
index 906e64c..50f8fb6 100644
--- a/Source/Fuse.Elements/Caching/ElementBatch.uno
+++ b/Source/Fuse.Elements/Caching/ElementBatch.uno
@@ -321,7 +321,7 @@ namespace Fuse.Elements
 				var opacity = entry._opacity;
 
 				var transform = entry._elm.LocalTransform;
-				float2 positionOrigin = transform[3].XY + (float2)cachingRect.Minimum * densityScale;
+				float2 positionOrigin = Vector.Transform((float2)cachingRect.Minimum * densityScale, transform);
 				float2 size = (float2)cachingRect.Size * densityScale;
 				float2 right = transform[0].XY * size.X;
 				float2 up = transform[1].XY * size.Y;
```

Source/Fuse.Elements/Caching/ElementBatch.uno
mortoray: Transform is returning a float4, so I need the swizzle.

kusma: Oh. Right.

mortoray:
@kusma Yes, that transform fixes it. I'm just trying a combination of Scaling, Rotation and Shear now to make sure it's okay.

kusma:
@mortoray: The above seems to fix the issue for me. But, it can be done even tighter:

```uno
diff --git a/Source/Fuse.Elements/Caching/ElementBatch.uno b/Source/Fuse.Elements/Caching/ElementBatch.uno
index 906e64c..c89f468 100644
--- a/Source/Fuse.Elements/Caching/ElementBatch.uno
+++ b/Source/Fuse.Elements/Caching/ElementBatch.uno
@@ -321,7 +321,8 @@ namespace Fuse.Elements
 				var opacity = entry._opacity;
 
 				var transform = entry._elm.LocalTransform;
-				float2 positionOrigin = transform[3].XY + (float2)cachingRect.Minimum * densityScale;
+				float2 localOrigin = (float2)cachingRect.Minimum * densityScale;
+				float2 positionOrigin = transform[3].XY + localOrigin.X * transform[0].XY + localOrigin.Y * transform[1].XY;
 				float2 size = (float2)cachingRect.Size * densityScale;
 				float2 right = transform[0].XY * size.X;
 				float2 up = transform[1].XY * size.Y;
```

So, the problem is that the origin was applied in the wrong coordinate-system.

kusma:
(The reason why I care about the "tightness" here, is that this is performance critical code)

PR #41, opened 2017-05-17
fixes x,y,width,height functions for already rooted nodes

mortoray:
fixes https://github.com/fusetools/fuselibs/issues/4189


Duckers:
Looks good, please merge

PR #45, opened 2017-05-18
Made ValueForwarder.IValueListener internal

kristianhasselknippe:
The docs generator tried to create links to this interface because it is
implemented by some public classes (like `With`, although they are
explicitly implemented). I guess this should be considered a bug in the
docs generator, but if this is an OK change anyway, its an easier fix
for now.

### DIFF
--- a/Source/Fuse.Reactive/ValueObserver.uno
+++ b/Source/Fuse.Reactive/ValueObserver.uno
@@ -76,7 +76,7 @@ namespace Fuse.Reactive
 
 	class ValueForwarder: ValueObserver
 	{
-		public interface IValueListener { void NewValue(object value); }
+		internal interface IValueListener { void NewValue(object value); }
 
 		IValueListener _listener;
 		public ValueForwarder(IObservable obs, IValueListener listener)


kusma:
@kristianhasselknippe: This is definitely a bug in the docs generator. The current code is sane as-is, and I don't want to have to live with limitations like this forever, as we're going to end up breaking similar things in the future, and playing catch-up like this is frustrating.

So while this *should* be OK for now, we need to _make sure_ this gets fixed in the docs generator. This would include adding a uno-issue for it.

Thanks for a thorough commit message. One small nit, though; we prefer imperative, present tense for commit messages. So that would be "Make ValueForwarder.IValueListener internal". But I would also name the motivation in the subject rather than the mechanism. So how about "Work around docs-generator bug". And we should also have a link to the uno-issue.

kusma:
@kristianhasselknippe: By the way, this code seems to be identical on release-1.0, didn't docs-generation work for that branch? Or is newer uno/docs-generator more picky somehow?

kristianhasselknippe:
@kusma: i'm actually not quite sure. I'll take a closer look at the generator

kristianhasselknippe:
We don't need to merge this now that i think of it. I'll just use it for docs generation and then i'll look into fixing it in the generator instead.

PR #47, opened 2017-05-19
Feature Android Native Surface impl

nekronos:
Native support for the `Fuse.Drawing.Surface` API on Android.


Source/Fuse.Drawing.Surface/Android/GraphicsSurface.uno
```diff
@@ -0,0 +1,161 @@
...
+		}
+
+		[Foreign(Language.Java)]
+		static Java.Object LoadImage(int glTextureId, int width, int height)
+		@{
+			int size = width * height * 4;
+			int[] pixels = new int[size];
+
+			IntBuffer pixelData = IntBuffer.wrap(pixels);
+			GLES20.glReadPixels(0, 0, width,height, GLES20.GL_RGBA, GLES20.GL_UNSIGNED_BYTE, pixelData);
```
kusma: You need to do `GLES20.glPixelStorei(GLES20.GL_UNPACK_ALIGNMENT, 1);` before this, otherwise you risk writing past the end of `pixels` (and also getting garbage) on some image sizes. We already have this problems other places, but please don't repeat the mistake ;)

nekronos: Fixed in 65fcbf0

PR #50, opened 2017-05-23
Index, Trigonometry, and math function extravaganza

mortoray:
You wanted functions you get functions. Lots of them.

Should be safe for a 1.0.1 release as it doesn't really change anything, mainly adds new stuff.

Fixes https://github.com/fusetools/fuselibs/issues/4190


Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -36,7 +36,7 @@ namespace Fuse
 			}
 			return null;
 		}
-
+
```
mortoray: Something came and went from that file during my branch. I'll remove the remnant whitespace.

Source/Fuse.Common/Tests/FuseTest/TestBase.uno
```diff
@@ -105,6 +105,57 @@ namespace FuseTest
...
+			for (int i=0; i < v.Children.Count; ++i)
+			{
+				var q = v.Children[i] as T;
+				if (q != null)
+					list.Add(q);
+			}
+			return list.ToArray();
+		}
+		
+		static string ConcatList(string a, string b)
```
mortoray: It's creating a comma-separated list, thus the name. (It's just a test helper function so I didn't think too long about it)

Source/Fuse.Reactive.Bindings/InstantiatorFunction.uno
```diff
@@ -0,0 +1,142 @@
...
+		}
+		
+		class InstantiatorSubscription : Subscription
+		{
+			InstantiatorFunction _expr;
+			Instantiator _instantiator;
+			Node _instance;
+			Selector _item;
+			IListener _listener;
+			IContext _context; //TODO: in base?
```
mortoray: Yeah, I'll check (I probaby already have that in the base class)

Source/Fuse.Common/Tests/FuseTest/TestBase.uno
kusma: OK, I guess that's fair enough... I'd kind of expect the input to be a list as well (or even varargs), but whatever, this'll do for now ;)

Source/Fuse.Marshal/Marshal.Cast.uno
```diff
@@ -114,6 +114,75 @@ namespace Fuse
...
 		}
+		
+		/**
+			Converts a value to a float4
+			
+			@return true if converted successfully, false if no suitable conversion exists
+			@param value the result value (0 padded as necessary)
+			@param size the size of the result
+		*/
+		public static bool TryMarshalToFloat4(object o, out float4 value, out int size)
```
mortoray: The name is changed now to reflect that it actually does a slightly different conversion.

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,376 @@ namespace Fuse.Reactive
...
+	[UXFunction("even")]
+	/** True if the rounded value is even, false otherwise*/
+	public sealed class Even : UnaryOperator
+	{
+		[UXConstructor]
+		public Even([UXParameter("Operand")] Expression operand): base(operand) {}
+		protected override object Compute(object operand)
+		{
+			//this rounds floats automatically it seems
+			var q = (int)Math.Floor(Marshal.ToType<float>(operand)+0.5f);
```
mortoray: Yes it is. It does round-to-positive-infinity on 0.5 values. Math.Round does this as well, but seeing as though it's the same I'll just use Math.Round instead.

Source/Fuse.Marshal/Marshal.Cast.uno
kusma: What's the difference?

mortoray: The `ToFloat4` takes a single value and does `float4(v)` whereas the new `TryToZeroFloat4` (new name) does `float4(v,0,0,0)`. Similarly for float2/3

kusma: Right, makes sense from an implementation point of view.

But, do we really want different behavior for different usages here? This sounds pretty confusing to the end-user to me...

mortoray: For the `attract` function, where this new code originate, it's a large difference. You need to zeros to animate correctly.

For the math functions it won't make a difference, the result is the same either way. But I prefer this new function since it gives out the size explicitly. The math functions don't change the size of the input then, a float2 remains a float2. In a way this function isn't really producing a `float4` it's just using it for encoding.

However, for value assignment, to a property, the current ToFloat4 makes sense. Since it provides the standard shortcuts of `Margin="4"` and `Margin="4,5"` for example. This function  is actually producing a `float4` for comsumption.

kusma:
@mortoray: We should IMO take this to master first, and if a lot of time passes until we get to ship master (and maybe get some usage with it?) we can consider back-porting to the release-branch.

kusma:
@mortoray: Looks like you'll have to rebase this.

mortoray:
@kusma I did rebase this from `release-1.0` to `master`, and the changeset appears correct. I have no idea how it's detecting code from other people in it.

kusma:
@mortoray: Seems we just needed to trigger a recheck :)

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+				return null;
+			int size = Math.Max(asize, bsize);
+			
+			if (size == 1)
+				return Math.Lerp(av.X, bv.X, tv);
+			if (size == 2)
+				return Math.Lerp(av.XY, bv.XY, tv);
+			if (size == 3)
+				return Math.Lerp(av.XYZ, bv.XYZ, tv);
+			if (size == 4)
```
kusma: switch instead, perhaps?

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+				!Marshal.TryToType<float>(mx, out mxv))
+				return null;
+			
+			if (size == 1)
+				return Math.Clamp(av.X, mnv, mxv);
+			if (size == 2)
+				return Math.Clamp(av.XY, mnv, mxv);
+			if (size == 3)
+				return Math.Clamp(av.XYZ, mnv, mxv);
+			if (size == 4)
```
kusma: Switch instead perhaps?

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+			int size;
+			if (Marshal.TryToZeroFloat4(operand, out v, out size))
+			{	
+				if (size == 1)
+					return _op(v[0]);
+				if (size == 2)
+					return float2((float)_op(v[0]),(float)_op(v[1]));
+				if (size == 3)
+					return float3((float)_op(v[0]),(float)_op(v[1]),(float)_op(v[2]));
+				if (size == 4)
```
kusma: Switch instead, perhaps?

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+		{
+			_name = name;
+			_op = op;
+		}
+		protected override object Compute(object left, object right)
+		{
+			double lv = 0;
+			double rv = 0;
+			if (!Marshal.TryToType<double>(left, out lv) ||
+				!Marshal.TryToType<double>(right, out rv))
```
kusma: The indentation here looks funky, see other comment ;)

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+			base(first, second, third) 
+		{ }
+		protected override object Compute(object a, object b, object t)
+		{
+			float4 av = float4(0), bv = float4(0);
+			int asize = 0, bsize = 0;
+			float tv = 0;
+			if (!Marshal.TryToZeroFloat4(a, out av, out asize) ||	
+				!Marshal.TryToZeroFloat4(b, out bv, out bsize) ||
+				!Marshal.TryToType<float>(t, out tv))
```
kusma: The indentation here looks funky, see other comment ;)

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+			base(first, second, third) 
+		{ }
+		protected override object Compute(object a, object mn, object mx)
+		{
+			float4 av = float4(0);
+			float mxv = 0, mnv = 0;
+			int size = 0;
+			if (!Marshal.TryToZeroFloat4(a, out av, out size) ||
+				!Marshal.TryToType<float>(mn, out mnv) ||
+				!Marshal.TryToType<float>(mx, out mxv))
```
kusma: The indentation here looks funky, see other comment ;)

Source/Fuse.Reactive.Expressions/Tests/MathFunctions.Test.uno
```diff
@@ -0,0 +1,231 @@
...
+					Assert.AreEqual( Math.Floor(v), p.floor.Float );
+					Assert.AreEqual( Math.DegreesToRadians(v), p.degreesToRadians.Float );
+					Assert.AreEqual( Math.RadiansToDegrees(v), p.radiansToDegrees.Float );
+					Assert.AreEqual( Math.Exp(v), p.exp.Float );
+					Assert.AreEqual( Math.Exp2(v), p.exp2.Float );
+					Assert.AreEqual( Math.Fract(v), p.fract.Float );
+					Assert.AreEqual( Math.Sign(v), p.sign.Float );
+					Assert.AreEqual( Math.Round(v), p.round.Float );
+					Assert.AreEqual( Trunc.Op(v), p.trunc.Float );
+					//some have special exceptions where the Inf/NaN value doesn't compare right on some platforms
```
kusma: Huh, really? What platforms are that? This sounds kinda serious...

Source/Fuse.Reactive.Expressions/Tests/MathFunctions.Test.uno
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+		
+		[Test]
+		public void TruncOp()
+		{
+			Assert.AreEqual(0, Trunc.Op(0.5));
+			Assert.AreEqual(0, Trunc.Op(-0.7));
+			Assert.AreEqual(5, Trunc.Op(5.9));
+			Assert.AreEqual(-110, Trunc.Op(-110.7));
```
kusma: I would also test stuff like `1.0 - 1e-20` vs `1.0` here (and the same for negatives)

Source/Fuse.Marshal/Marshal.Convert.uno
```diff
@@ -76,13 +76,31 @@ namespace Fuse
 			return false;
 		}
 
+		/**
+			Be aware this function may throw a NullReferenceException if the type cannot be converted to the desired one. It is advised to use TryToType or TryConvertTo instead.
```
mortoray: It was throwing NullReferenceException since we convert to null and then attempt to cast, which results in NRE. I did't wish to change the odd behaviour in this PR.

Source/Fuse.Nodes/Node.uno
```diff
@@ -133,5 +133,28 @@ namespace Fuse
 			return this;
 		}
 
+		//We want `: Behavior` here. This is an attempt to workaround a Uno/DotNet issue on some platforms
```
mortoray: I didn't isolte a standalone test-case and thus didn't record an issue. Shall I remove this comment, or what shall I do instead?

Source/Fuse.Reactive.Bindings/Instance.uno
```diff
@@ -324,12 +324,36 @@ namespace Fuse.Reactive
...
 		}
 		
-		//contains only the added nodes (not all data items)
+		/**
+			A list of all items which have been added to the Instantiator. This only includes the items
+			within the range of Offset+Limit.
+			
+			This list should only be modified via the InsertNew, RemoveAt, and RemoveAll functions.
+		*/
 		List<WindowItem> _windowItems = new List<WindowItem>();
```
mortoray: Encapsulation would make it harder to write those other functions. This is more of a note for that if you need to modify this value you shouldn't do it unless you know what those other functions already do. I'm never clear on where/how to make such notes.

Source/Fuse.Reactive.Bindings/Instance.uno
```diff
@@ -700,8 +731,9 @@ namespace Fuse.Reactive
 			//find last node prior to where we want to introduce
 			var lastNode = GetLastNodeFromIndex(windowIndex-1);
 
-			Parent.InsertNodes( Parent.Children.IndexOf(lastNode) + 1, newElements.GetEnumerator() );
+			//assign first for lookups during the rooting in Insert can find the data
 			wi.Nodes = newElements;
+			Parent.InsertNodes( Parent.Children.IndexOf(lastNode) + 1, newElements.GetEnumerator() );
```
mortoray: The ordering of the two statements is changed. It's the new `index` expression that requires this. Previously it didn't make a difference since nothing would callback into `Each` during the insertion.

It is a bugfix, but it only affects the new features.

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -33,4 +34,462 @@ namespace Fuse.Reactive
...
+	public sealed class Alternate : BinaryOperator
+	{
+		[UXConstructor]
+		public Alternate([UXParameter("Left")] Expression left, [UXParameter("Right")] Expression right): base(left, right) {}
+		protected override object Compute(object left, object right)
+		{
+			float fvalue = 0;
+			float fgroup = 0;
+			if (!Marshal.TryToType<float>(left, out fvalue) ||
+				!Marshal.TryToType<float>(right, out fgroup))
```
mortoray: The indentation is normal here. It's one additional tab from the previous line, since it's still part of that line. The other option would be to start with `if(` and have all conditions on subsequent lines for alignment. I'm not sure that's better though.

mortoray: multi-line conditionals are just always a formatting problem. :/

Source/Fuse.Reactive.Expressions/MathFunctions.uno
mortoray: Sure.

Source/Fuse.Reactive.Expressions/Tests/MathFunctions.Test.uno
mortoray: I think it was DotNet and also Windows. I didn't think we'd be able to influence NaN handling on any target anyway, plus since the results make no sense in these test conditions, I didn't bother looking further. I don't know which platform is correct, in theory NaN shouldn't compare valid on any of them, but since there's object wrappers involved in some... 

The conditions limit the tests for proper values, skipping NaN tests then.

Source/Fuse.Marshal/Marshal.Convert.uno
kusma: OK, makes sense.

Source/Fuse.Nodes/Node.uno
kusma: I'd just file a ticket with out a full set of information (or maybe to a branch of fuselibs wich reproduces it).

Source/Fuse.Reactive.Bindings/Instance.uno
kusma: Fair enough.

Source/Fuse.Reactive.Bindings/Instance.uno
kusma: OK, thanks for clarifying.

Source/Fuse.Reactive.Expressions/Tests/MathFunctions.Test.uno
kusma: Fair enough.

PR #54, opened 2017-05-24
Add fix and test for issue fuselibs-public #53

### LINKED ISSUE #53: Regression in Fuse 1.0.0 - Containers with no subtree nodes throw NRE
Reproduced with https://github.com/Duckers/Fuse.MaterialDesign

Causes NRE in `Container.uno`

Duckers:
Fixes #53 - NRE in Container.uno

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+## Bug in Container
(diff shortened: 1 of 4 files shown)


Duckers:
@kusma Can you please help me understand why this travis-ci thing fails? I can't see any details about it

kusma:
@Duckers: it was a random failure, and re-triggering it made the problem go away. A little bit disappointing, but we have some stack-trace logging going on for master, that will hopefully let us find the culprit.

PR #55, opened 2017-05-26
Fix: Marshal.TryConvertTo would throw exception instead of fail gracefully

Duckers:
This fixes and adds tests for cases where Marshal.TryConvertTo would throw exception instead of returning false as expected when an object is attempted converted to an incompatible type.

This caused a crashbug in the FuseCloud app


cbaggers:
Detail. It's causing a bug in my modified version of FuseCloud, which is likely triggered by my changes. As far as I know it is not affecting our shipping FuseCloud app.

kusma:
Sent from wrong repo

PR #58, opened 2017-05-26
Adding MapTile as a Polyfill for MapView, instead of boring NativeVie…

bolav:
…wHost message.

Just a first stab to show something. Next things to add would be:

- Pointer interaction, like drag, pinch to zoom, and rotate
- Support for other sizes than quadratic
- Support for other grid sizes than 3x3

This first effort is barebones, but IMO still a real help when developing map-related apps

This PR contains:
- [X] Changelog
- [X] Documentation
- [ ] Tests


Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,285 @@
...
+{
+	/** A polyfill for Native @MapView
+
+	This is not meant to be a full implementation, but just something that will help you when developing on local preview.
+
+	This should also work as a standalone component, so it can be used on native.block
+
+	This uses the OSM tile server, and that has both a usage policy https://operations.osmfoundation.org/policies/tiles/ and a license https://www.openstreetmap.org/copyright .
+
+	The mappin is from wikipedia https://commons.wikimedia.org/wiki/File:Map_pin_icon.svg and is Creative Commons Attribution-Share Alike 3.0 Unported.
```
kusma: How do we respect the "Attribution"-part here? Do you think this link to the wikipedia-file is sufficient?

bolav: That was my thoughts. The reason is:

- This is not production ready yet. So it will only be used in development.
- If you release a high traffic app, you would need either your own mapserver (something like https://github.com/ncareol/osm-tiles-docker ), or a service (like http://www.geofabrik.de/ ), both of which would change the attribution requirement.

kusma: I was thinking about the mappin asset, not the mapserver.

bolav: Sorry about that. I'm unsure about the attribute, since it's for now only meant for development. Best case would of course be if you (Fuse) had a gfx artist that could create a pin for us.

kusma:
If I add this to manual test app, like so:
```diff
diff --git a/Tests/ManualTests/ManualTestingApp/MainView.ux b/Tests/ManualTests/ManualTestingApp/MainView.ux
index 4bcaabd..e2d0dfe 100644
--- a/Tests/ManualTests/ManualTestingApp/MainView.ux
+++ b/Tests/ManualTests/ManualTestingApp/MainView.ux
@@ -138,13 +138,13 @@
 			<HitTestPage/>
 			<DefaultFonts/>
 			<Pages.SurfaceMisc/>
-			
+			<MapViewPage/>
+
 			<Android>
 				<WebViewPage/>
 				<BackButtonView/>
 			</Android>
 			<iOS>
-				<MapViewPage/>
 				<WebViewPage/>
 			</iOS>
 		</PageControl>
```

And compile and run, like so:

`./Stuff/uno build -tdotnet -r Tests/ManualTests/ManualTestingApp/ManualTestingApp.unoproj`

I get this:

```
The application has crashed because of an unhandled exception:

System.InvalidProgramException: Common Language Runtime detected an invalid program.
   at Fuse.Controls.MapView.UpdateCameraNextFrame()
   at Fuse.Controls.MapView.SetLatitude(Double value, IPropertyListener origin) in C:\Users\erik\src\fuselibs\Source\Fuse.Maps\MapView.uno:line 511
   at ManualTestingApp_FuseControlsMapView_Latitude_Property.Set(PropertyObject obj, Double v, IPropertyListener origin) in C:\Users\erik\src\fuselibs\Tests\ManualTests\ManualTestingApp\.uno\ux11\ManualTestingApp.unoproj.g.uno:line 418
   at Uno.UX.Property`1.SetAsObject(PropertyObject obj, Object value, IPropertyListener origin) in C:\Users\erik\src\fuselibs\Stuff\lib\UnoCore\1.0.11-master-67495cd\source\uno\ux\$.uno:line 306
   at Uno.UX.Property.SetAsObject(Object value, IPropertyListener origin) in C:\Users\erik\src\fuselibs\Stuff\lib\UnoCore\1.0.11-master-67495cd\source\uno\ux\$.uno:line 283
   at Fuse.Reactive.DataBinding.SetTarget(Object value) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\DataBinding.uno:line 163
   at Fuse.Reactive.DataBinding.TryPushAsValue(Object newValue) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\DataBinding.uno:line 239
   at Fuse.Reactive.DataBinding.PushValue(Object newValue) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\DataBinding.uno:line 229
   at Fuse.Reactive.DataBinding.NewValue(Object value) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\DataBinding.uno:line 207
   at Fuse.Reactive.ExpressionBinding.Fuse.Reactive.IListener.OnNewData(IExpression source, Object obj) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\ExpressionBinding.uno:line 51
   at Fuse.Reactive.DataSubscription.Resolve(Object data) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\Subscription\DataSubscription.uno:line 46
   at Fuse.Node.DataFinder.NextData(Object data) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.DataContext.uno:line 52
   at Fuse.Node.EnumerateData(IDataEnumerator e) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.DataContext.uno:line 96
   at Fuse.Reactive.DataSubscription.FindData() in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\Subscription\DataSubscription.uno:line 32
   at Fuse.Reactive.DataSubscription..ctor(IExpression source, Binding origin, String key, IListener listener) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\Subscription\DataSubscription.uno:line 21
   at Fuse.Reactive.ExpressionBinding.Fuse.Reactive.IContext.Subscribe(IExpression source, String key, IListener listener) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\ExpressionBinding.uno:line 34
   at Fuse.Reactive.Data.Subscribe(IContext context, IListener listener) in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Expressions\Expression.uno:line 85
   at Fuse.Reactive.ExpressionBinding.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\ExpressionBinding.uno:line 29
   at Fuse.Reactive.DataBinding.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.Bindings\DataBinding.uno:line 126
   at Fuse.Binding.Root(Node parent) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Binding.uno:line 19
   at Fuse.Node.RootBindings() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.Bindings.uno:line 48
   at Fuse.Node.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.Rooting.uno:line 129
   at Fuse.Visual.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.uno:line 91
   at Fuse.Elements.Element.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Node.uno:line 30
   at Fuse.Controls.Control.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Controls\Control.Background.uno:line 78
   at Fuse.Controls.LayoutControl.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Panels\LayoutControl.uno:line 117
   at Fuse.Controls.Panel.OnRooted() in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Panels\Panel.uno:line 73
   at Fuse.Node.RootInternalImpl(Visual parent) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.Rooting.uno:line 110
   at Fuse.Node.RootInternal(Visual parent) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.Rooting.uno:line 80
   at Fuse.Node.Relate(Visual parent, Node child) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Node.Rooting.uno:line 171
   at Fuse.Visual.OnAdded(Node b) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.Children.uno:line 106
   at Fuse.Visual.InsertNodes(Int32 index, IEnumerator`1 items) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.Children.uno:line 209
   at Fuse.NodeGroupBase.AddContent() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\NodeGroup.uno:line 165
   at Fuse.NodeGroupBase.set_UseContent(Boolean value) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\NodeGroup.uno:line 47
   at Fuse.Triggers.Trigger.Start() in C:\Users\erik\src\fuselibs\Source\Fuse.Triggers\Trigger.uno:line 328
   at Fuse.Triggers.Trigger.PlayEnd(Boolean on, Action done) in C:\Users\erik\src\fuselibs\Source\Fuse.Triggers\Trigger.uno:line 497
   at Fuse.Triggers.Trigger.DirectActivate(Action done) in C:\Users\erik\src\fuselibs\Source\Fuse.Triggers\Trigger.uno:line 416
   at Fuse.Triggers.Trigger.Activate(Action done) in C:\Users\erik\src\fuselibs\Source\Fuse.Triggers\Trigger.uno:line 404
   at Fuse.Triggers.WhileTrigger.SetActive(Boolean on) in C:\Users\erik\src\fuselibs\Source\Fuse.Triggers\WhileTrigger.uno:line 19
   at Fuse.Triggers.WhileVisible.OnIsVisibleChanged(Object sender, EventArgs args) in C:\Users\erik\src\fuselibs\Source\Fuse.Triggers\WhileVisible.uno:line 27
   at Fuse.Visual.InvokeEventHandler(Object obj, Object args) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.EventStorage.uno:line 40
   at Fuse.Properties.ForeachInList(PropertyHandle handle, Action`2 action, Object state) in C:\Users\erik\src\fuselibs\Source\Fuse.Common\Properties.uno:line 125
   at Fuse.Visual.RaiseEvent(PropertyHandle ph, VisualBits ne) in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.EventStorage.uno:line 35
   at Fuse.Visual.OnIsVisibleChanged() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.IsVisible.uno:line 63
   at Fuse.Elements.Element.OnIsVisibleChanged() in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.ITreeRenderer.uno:line 35
   at Fuse.Visual.UpdateIsVisibleCache() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.IsVisible.uno:line 47
   at Fuse.Visual.UpdateIsVisibleCache() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.IsVisible.uno:line 51
   at Fuse.Visual.UpdateIsVisibleCache() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.IsVisible.uno:line 51
   at Fuse.Visual.OnLocalVisibleChanged() in C:\Users\erik\src\fuselibs\Source\Fuse.Nodes\Visual.IsVisible.uno:line 37
   at Fuse.Elements.Element.OnVisibilityChanged(Visibility oldVisibility, IPropertyListener origin) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Layout.Properties.uno:line 201
   at Fuse.Elements.Element.SetVisibility(Visibility value, IPropertyListener origin) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Layout.Properties.uno:line 194
   at Fuse.Elements.Element.set_Visibility(Visibility value) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Layout.Properties.uno:line 187
   at Fuse.Controls.PageControl.UpdateProgress(Element page, NavigationPageState state, PageData pd) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Navigation\PageControl.uno:line 207
   at Fuse.Controls.NavigationControl.OnPageProgressChanged(Object page, NavigationArgs args) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Navigation\NavigationControl.uno:line 229
   at Fuse.Navigation.NavigationHandler.Invoke(Object sender, NavigationArgs state)
   at Fuse.Navigation.VisualNavigation.OnPageProgressChanged(Double current, Double prev, NavigationMode mode) in C:\Users\erik\src\fuselibs\Source\Fuse.Navigation\VisualNavigation.uno:line 84
   at Fuse.Navigation.StructuredNavigation.ChangeProgress(Single prev, Single next, NavigationMode mode) in C:\Users\erik\src\fuselibs\Source\Fuse.Navigation\StructuredNavigation.uno:line 252
   at Fuse.Navigation.StructuredNavigation.OnUpdated() in C:\Users\erik\src\fuselibs\Source\Fuse.Navigation\StructuredNavigation.uno:line 214
   at Fuse.UpdateListener.Invoke() in C:\Users\erik\src\fuselibs\Source\Fuse.Common\UpdateManager.uno:line 52
   at Fuse.UpdateManager.ProcessListeners(Stage stage, List`1& _exceptions) in C:\Users\erik\src\fuselibs\Source\Fuse.Common\UpdateManager.uno:line 359
```

Seems we're generating some illegal .NET bytecode here :(

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,293 @@
...
+				}
+				else
+				{
+					UpdateMap();
+				}
+			}
+		}
+
+		float2 _tilepos = float2(0);
+		float2 _cornertile = float2(0);
```
kusma: `int2`, perhaps?

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,293 @@
...
+			{
+				for (var x = (int)p.X - 1; x < (int)p.X + 2; x++)
+				{
+					maps[i].Url = MakeUrl((int)Zoom, x, y);
+					i++;
+				}
+			}
+			var adj = float2(0.5f);
+			adj.X -= p.X - Math.Floor(p.X);
+			adj.Y -= p.Y - Math.Floor(p.Y);
```
kusma: This makes me want to rewrite this as something like this:

```diff
diff --git a/Source/Fuse.Maps/MapTile.uno b/Source/Fuse.Maps/MapTile.uno
index 88cd761..b2b6311 100644
--- a/Source/Fuse.Maps/MapTile.uno
+++ b/Source/Fuse.Maps/MapTile.uno
@@ -250,9 +250,12 @@ namespace Fuse.Controls
 			if (m == null) return;
 			var p = WorldToTilePos(Longitude, Latitude, (int)Zoom);
 			_tilepos = p;
+
+			var tmp = Math.Floor(p);
+			_cornertile = (int2)tmp - 1;
+			var adj = 0.5f - p - tmp;
+
 			var i = 0;
-			_cornertile.X = Math.Floor(p.X - 1);
-			_cornertile.Y = Math.Floor(p.Y - 1);
 			for (var y = (int)p.Y - 1; y < (int)p.Y + 2; y++)
 			{
 				for (var x = (int)p.X - 1; x < (int)p.X + 2; x++)
@@ -261,9 +264,7 @@ namespace Fuse.Controls
 					i++;
 				}
 			}
-			var adj = float2(0.5f);
-			adj.X -= p.X - Math.Floor(p.X);
-			adj.Y -= p.Y - Math.Floor(p.Y);
+
 			adj = adj * maps[4].ActualSize; 
 			trans.XY = adj;
 			UpdateMarkers();
```

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,293 @@
...
+			}
+			set
+			{
+				_lng = value;
+				MapView m = Parent as MapView;
+				if (m != null)
+				{
+					m.Longitude = value; 
+				}
+				else {
```
kusma: Please no egyptian

Source/Fuse.Maps/MapView.uno
```diff
@@ -241,6 +238,10 @@ namespace Fuse.Controls
 		bool _willUpdateCameraNextFrame;
 		internal void UpdateCameraNextFrame()
 		{
+			if defined(!MOBILE) {
+			        map.UpdateMap();
+			        return;
+			}
```
kusma: This fixed the .NET error:
```diff
diff --git a/Source/Fuse.Maps/MapView.uno b/Source/Fuse.Maps/MapView.uno
index 4fcc6cc..6643e58 100644
--- a/Source/Fuse.Maps/MapView.uno
+++ b/Source/Fuse.Maps/MapView.uno
@@ -238,13 +238,18 @@ namespace Fuse.Controls
 		bool _willUpdateCameraNextFrame;
 		internal void UpdateCameraNextFrame()
 		{
-			if defined(!MOBILE) {
-			        map.UpdateMap();
-			        return;
+			if defined(!MOBILE)
+			{
+				map.UpdateMap();
+			}
+			else
+			{
+				if(!MapIsReady || _willUpdateCameraNextFrame)
+					return;
+
+				UpdateManager.PerformNextFrame(ApplyCameraState, UpdateStage.Primary);
+				_willUpdateCameraNextFrame = true;
 			}
-			if(!MapIsReady || _willUpdateCameraNextFrame) return;
-			UpdateManager.PerformNextFrame(ApplyCameraState, UpdateStage.Primary);
-			_willUpdateCameraNextFrame = true;
 		}
 
 		void ApplyCameraState()
```

The problem here, is that .NET doesn't like it when we have dead code after an unconditional return. Uno is not so clever at optimizing this away.

kusma:
I get these errors when trying this out in the manual test app:

InternalError: ImageSource error: 'HttpImageSource-failed-status: 404 ' in Fuse.Resources.HttpImageSourceImpl<C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Resources\ImageSource.uno:111>
InternalError: ImageSource error: 'HttpImageSource-failed-status: 404 ' in Fuse.Resources.HttpImageSourceImpl<C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Resources\ImageSource.uno:111>
InternalError: ImageSource error: 'HttpImageSource-failed-status: 404 ' in Fuse.Resources.HttpImageSourceImpl<C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Resources\ImageSource.uno:111>

It seems we're asking for out-of-bounds map-tiles when we're all zoomed out.

kusma:
And just to be clear: This is really cool stuff :D

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,293 @@
...
+		}
+
+		// http://wiki.openstreetmap.org/wiki/Zoom_levels
+		// http://wiki.openstreetmap.org/wiki/Slippy_map_tilenames
+		String _tileserver = "http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png";
+		public String TileServer
+		{
+			get { return _tileserver; }
+			set { _tileserver = value; }
+		}
```
bolav: Depends on the usecase of the class. If it is to be used in production I think it is valuable to be able to override the servers.

kusma: This is a polyfill for better preview-experience, not a production-feature. All our production-targets already have "real" mapview implementations.

Besides, this is not currently possible to override through the normal interface, because the MapView doesn't provide an alternative URL configuration. So at the moment it's more or less dead code.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -257,6 +257,12 @@ namespace Fuse.Controls
 			var i = 0;
 			var tmp = Math.Floor(p);
 			_cornertile = (int2)tmp - 1; // Have the corner one to the left and above the center of the map
+
+			// Calculate offset (adjustment) of tile, compared to center of map:
+			var adj = (float2(0.5f) - (p - tmp));
```
kusma: You don't need the outer parens here

Source/Fuse.Maps/MapTile.uno
bolav: Yes. My idea is that `MapTile` can be used directly, instead of `MapView` when it get's good enough. This is good if you want to use selfmade OSM tiles. That is my main reason for wanting to keep it.

kusma: OK, fair enough.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+			}
+			UpdateMarkers();
+		}
+
+		public float2 WorldToTilePos(double lon, double lat, int zoom)
+		{
+			float2 p = float2(0);
+			p.X = (float)((lon + 180.0) / 360.0 * (1 << zoom));
+			p.Y = (float)((1.0 - Math.Log(Math.Tan(lat * Math.PI / 180.0) + 
+				1.0 / Math.Cos(lat * Math.PI / 180.0)) / Math.PI) / 2.0 * (1 << zoom));
```
mortoray: Please use `Math.Pow(2,zoom)` instead of `1 << zoom` when using floats.
Use the `Math.DegreessToRadians` and RadiansToDegrees functions to clarify this code.

mortoray: As `Math.Cos` can be zero this `1/Cos` part looks undefined for some latitudes. You should always check the divisors.

mortoray: I'd also like some reference or comment as to what formula this is. Just a name or link is fine.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+	This should also work as a standalone component, so it can be used on native.block
+
+	This uses the OSM tile server, and that has both a usage policy https://operations.osmfoundation.org/policies/tiles/ and a license https://www.openstreetmap.org/copyright .
+
+	The mappin is from wikipedia https://commons.wikimedia.org/wiki/File:Map_pin_icon.svg and is Creative Commons Attribution-Share Alike 3.0 Unported. 
+
+	@seealso Fuse.Controls.MapView
+	*/
+
+	public class MapTile : Panel
```
mortoray: If this is only meant as a polyfill perhaps it can be `internal` and not `public`. Or is there any value in a user instantiating it on their own?

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+			trans = new Fuse.Translation();
+			var scaling = new Fuse.Scaling()
+			{
+				Factor=1.5f
+			};
+			Children.Add(grid);
+			Children.Add(trans);
+			Children.Add(scaling);
+			Placed += OnPlaced;
+			_markerGraphicsCache = new MarkerIconCache(UpdateMarkers);
```
mortoray: This likely creates a memory loop, _markerGraphicsCache -> this -> _markerGraphicsCache

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+			{
+				Children.Remove(i);
+			}
+			_markers.Clear();
+		}
+
+		List<Fuse.Controls.Image> _markers = new List<Fuse.Controls.Image>();
+		public void UpdateMarkers()
+		{
+			ClearMarkers();
```
mortoray: Clearing the markers and recreating them on any update is inefficient. Instead of recreating the `Image` obejcts each time could you not reuse the existing ones?

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+				var _trans = new Fuse.Translation()
+				{
+					RelativeNode = i,
+					RelativeTo   = Fuse.TranslationModes.Size,
+					X = m.IconAnchorX - 0.5f,
+					Y = m.IconAnchorY * -1
+				};
+
+				Children.Add(i);
+				BringToFront(i);
```
mortoray: It's better to just set a `ZOffset` on the children instead of relying on modifying the ordering this way.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+			{
+				FileSource fs = m.IconFile;
+				if (fs == null)
+				{
+					fs = new global::Uno.UX.BundleFileSource(import global::Uno.IO.BundleFile("./mappin.png"));
+					m.IconAnchorY = 1;
+					m.IconAnchorX = 0;
+				}
+				// The diff between the panel and the grid. Offset this to center the marker
+				var diff = (ActualSize - _grid_size) / 2;
```
mortoray: Is this calculation actually necessary? Can't you use realtive positioning on the child element iself and let it calculate it's position?

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+				var p = WorldToTilePos(m.Longitude, m.Latitude, (int)Zoom);
+
+				// P is now relative to the top corner of the grid
+				p -= _cornertile;
+				// P is now in pixels
```
mortoray: It's still in Points I believe.

Source/Fuse.Maps/MapTile.uno
kusma: @mortoray: Why is `Math.Pow(2, zoom)` better? Does it have something to do with integer overflow?

mortoray: In the case of zoom I guess that's unlikely, but in general I don't like using integer ops when it's a floating point operation.

Note that `Zoom` is a floating point value. The casting to `int` in several places also makes this suspicious. `Math.Pow` works fine with float, but of course `<<` would not.

Source/Fuse.Maps/MapTile.uno
bolav: My thoughts are that this could be good enough for the user to want to instantiate it on their own. That is why both `TileServer` is available, and this is available.

Source/Fuse.Maps/MapTile.uno
bolav: Then I need help in how to solve this. I was following the pattern of the original code:
https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Maps/iOS/MapView.uno#L172

			_markerGraphicsCache = new MarkerIconCache(UpdateMarkers);

https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Maps/Android/MapView.uno#L39

			_markerGraphicsCache = new MarkerIconCache(UpdateMarkers);

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+			Children.Add(grid);
+			Children.Add(trans);
+			Children.Add(scaling);
+			Placed += OnPlaced;
+			_markerGraphicsCache = new MarkerIconCache(UpdateMarkers);
+		}
+
+		// http://wiki.openstreetmap.org/wiki/Zoom_levels
+		// http://wiki.openstreetmap.org/wiki/Slippy_map_tilenames
+		String _tileserver = "http://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png";
```
bolav: I was hoping that this was enough to solve that:

https://github.com/fusetools/fuselibs-public/pull/58/files#diff-3d31341ced45520565b9eeb0a7507703R15

Source/Fuse.Maps/MapTile.uno
bolav: I agree. My focus on this first effort was just to make something that works.

Source/Fuse.Maps/MapTile.uno
bolav: Probably, but I was struggling a lot to get the positioning correct. I'm not even sure it's correct yet. Seems to me that the center of the map jitters a bit when I zoom. But long story short; if we want better positioning I need help. This is the best I managed after using a lot of time.

Source/Fuse.Maps/MapTile.uno
bolav: Yes. The comment is pixels/points in contrast to ratio.

Source/Fuse.Maps/MapTile.uno
bolav: Example please.

mortoray: `i.ZOffset = 1`

Source/Fuse.Maps/MapTile.uno
mortoray: Without a solid use-case it'd be better to keep it internal. It provides a valuable service for MapView, but on it's own it's uncertain. 

If it's public then we'd have to go through an API review, require examples, more docuemtnation, and also be supported in future version. All things we'd like to avoid where possible.

Source/Fuse.Maps/MapTile.uno
mortoray: Those unfortunately have the same memory leak. I'll file an issue for them. You might be able to fix them at the same time though. 

The general approach for messaging, and this cache is one of them, is to only add a listener during rooting, and then remove it during unrooting. You don't care about changes outside of rooting; you'll just be iterating over everything anyway at rooting time.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+				_markers.Add(i);
+			}
+		}
+
+		bool ready = false;
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+			MapView m = Parent as MapView;
+			m.ClipToBounds = true;
```
bolav: I have to clip to the parents bounds, and not to my bounds, because of the zooming done on self. 

https://github.com/fusetools/fuselibs-public/pull/58/files#diff-3d31341ced45520565b9eeb0a7507703R49

Source/Fuse.Maps/MapTile.uno
mortoray: The tiles have an attribution requirement. I think this means we need to display text on the screen saying where the tiles are from.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+			ready = true;
+		}
+
+		float _grid_size = 0;
+		public void OnPlaced(object sender, PlacedArgs args)
+		{
+			var s = Math.Min(ActualSize.X, ActualSize.Y);
+			grid.Width = s;
```
bolav: Yes. That looks really promising, although I need to figure out how to calculate the marker placements.

Source/Fuse.Maps/MapTile.uno
mortoray: It's still an OSM tile viewer, not a general purpose one. You couldn't for example hook into Google or Apple maps with it.

Source/Fuse.Maps/MapTile.uno
```diff
@@ -0,0 +1,298 @@
...
+				{
+					_lng = m.Longitude; 
+				}
+				return _lng;
+			}
+			set
+			{
+				_lng = value;
+				MapView m = Parent as MapView;
+				if (m != null)
```
bolav: The idea is that if we have a MapView as a parent, there is no need to set lat and lng on this component, the MapView's values will be the ones to use, so it doesn't seem necessary to set these on the parent during rooting.

Source/Fuse.Maps/MapTile.uno
mortoray: All the tiles are still children of this panel. If this just fills the to parent the clipping wouldn't be any different if placed on this element, even with zooming.

ClipToBounds clips to the ActualSize bounds, not the rendering bounds. Zoom has no impact on those bounds.

mortoray:
This license requirement concerns me: "Heavy use (e.g. distributing an app that uses tiles from openstreetmap.org) is forbidden without prior permission from the System Administrators. See below for alternatives."

I think this would be a really nice feature to preview MapView, but we'd need a way to actually prevent this feature from being distributed, at least by default.  I think if the tile class itself is just `internal`, at least for now, that might be enough.

In either case, based on that requirement, I think we'll unfortunately need legal review before we can even use this in our desktop app.

Source/Fuse.Maps/MapTile.uno
bolav: zoom is an integer. Do you still want `Math.Pow`?

mortoray: I just prefer semantically relevant operations. Bitops are not floating point ops, so it looks odd in this equation -- like an unneeded optimization.  When I see a bitop I assume there's some reason for it, but in this case there isn't.

This is of course a very minor point, and I wouldn't block a PR because of such things.

Source/Fuse.Maps/MapTile.uno
mortoray: If having `MapView` as a parent is necessary then remove the setters. If the setters remain then it's important to keep the state synchronized with the `MapView` (which is problematic).

bolav:
Stamen has OSM tileservers: http://maps.stamen.com/#howto

Carto: https://carto.com/location-data-services/basemaps/

Source/Fuse.Maps/MapTile.uno
bolav: Okay. I will make the changes to make this internal.

Source/Fuse.Maps/MapTile.uno
bolav: This works as I want when I have `ClipToBounds` on the parent, and it doesn't work when I put it on this Panel itself. The reason is that I'm adding a `Translation` on the Panel itself, to move it, so that the wanted center of the map is centered. And then I clip with the parent, to not show the map outside the bounds. This is a workaround, but it was how I made it work.

bolav:
@mortoray If the feature will be added based on that, maybe it helps if I contact the system administrators of OSM tileserver?

Source/Fuse.Maps/OSMMapTile.uno
```diff
@@ -0,0 +1,255 @@
...
+
+		MapView _mapview_parent = null;
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+			Placed += OnPlaced;
+			_mapview_parent = Parent as MapView;
+			if (_mapview_parent == null) {
+				_mapview_parent.ClipToBounds = true;
+				Fuse.Diagnostics.UserError( "OSMapTile needs a MapView as it's parent", Parent );
```
Sunjammer: "its" (sorry, pet peeve)

Source/Fuse.Maps/Fuse.Maps.unoproj
```diff
@@ -30,6 +31,7 @@
     "iOS/MapViewController.m:CSource:iOS",
     "Android/*",
     "Android/FuseMap.java:Java:Android",
-    "MarkerIconCache.uno"
+    "MarkerIconCache.uno",
+    "mappin.png:Bundle"
```
Sunjammer: Only bundle this if not mobile target

bolav: Can you please give me the correct syntax, so I don't need to hunt for it?

Source/Fuse.Maps/OSMMapTile.uno
bolav: Sorry about that.

bolav:
@Sunjammer are the documentation provided here: https://github.com/fusetools/fuselibs-public/pull/58/files#diff-e4d0372cef320c93beb16b6ea445fee7R9 not enough? If that's the case, please give me an example of documentation that you think would suffice.

Source/Fuse.Maps/Fuse.Maps.unoproj
kusma: I believe it should be `"mappin.png:Bundle:!Mobile"`

Sunjammer:
Please update Fuse.Maps/Docs/Brief.md to inform how desktop/preview behavior differs from device

Source/Fuse.Maps/Docs/Brief.md
```diff
@@ -1,6 +1,6 @@
 The `MapView` allows you to present annotated, interactive world-wide maps to the user using the mapping APIs native to the platform: Google Maps on Android and Apple Maps on iOS.
 
-The `MapView` is a native control, and thus needs to be contained in a @NativeViewHost to be displayed. As with other native mobile controls, there currently isn't a `MapView` available for desktop targets.
+The `MapView` is a native control, and thus needs to be contained in a @NativeViewHost to be displayed. As with other native mobile controls. There is also a barebones `MapView` available for desktop, but it doesn't support interaction or gestures yet.
```
Sunjammer: I'd just remove the whole line `As with other native mobile controls` here.

Source/Fuse.Maps/Docs/Brief.md
```diff
@@ -39,3 +39,7 @@ Google Maps requires a valid Google Maps API key. Follow [Google's documentation
     }
 }
 ```
+
+# Maps on Desktop
+
+The desktop uses OSM tiles to display a placeholder map, so map development is possible using only local preview. It's still early days for the functionality, so there are no interactions possible, and some of the properties are not functional yet. (Tilt, Style and Bearing.)
```
Sunjammer: Maybe link to OSM here for reference?

mortoray:
Closing as abandoned, unfortunately. Recorded in issue https://github.com/fusetools/fuselibs-public/issues/821  hopefully somebody decides to finish this.

PR #59, opened 2017-05-28
JavaScript Dependency Injection

Duckers:
> Requires new Uno from `feature-JSDI` to work: https://github.com/fusetools/uno/pull/1159

Added support for injecting UX expressions into `<JavaScript>` tags using the `dep` XML namespace. S

Example:
```
	<JavaScript>
		exports.foo = 123
	</JavaScript>
	<JavaScript dep:foo="{foo}">
		foo // this is now 123
	</JavaScript>
```
A script is not executed until all of its dependencies are available. If any of the dependencies change, the script is re-executed.

This has multiple use-cases:

* Accessing data from data context `dep:foo="{foo}"`
* Accessing properties synchronously `dep:SomeProp="{Property SomeProp}"`
* Dependency injection in tests

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


kusma:
Sent from wrong repo.

PR #60, opened 2017-05-28
JavaScript: optional explicit require() of UX symbols

Duckers:
> This branch builds upon #59 - so please merge that one first

Symbols declared with `ux:Name`, `ux:Dependency` or `dep` are in this PR also available to `require()` for `<JavaScript>` modules. 

This allows us to write code that plays nicer with transpilers and linters. Using require for names declared in UX is optional, but may make the code more readable and maintainable, e.g. `var router = require("router")` over just using `router` with no declaration.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


kusma:
Sent from wrong repo.

PR #61, opened 2017-05-28
Implemented array lookup in UX expressions

Duckers:
UX expressions now support arbitrary array lookups, e.g. `{someArray[index+3]}`. The same syntax can also be used with string keys, e.g `{someObject[someKey]}`. The lookup is fully reactive - both the collection and the key/index can change.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Reactive.Expressions/LookUp.uno
+++ b/Source/Fuse.Reactive.Expressions/LookUp.uno
@@ -1,0 +1,213 @@
+using Uno;
+using Uno.UX;
+using Uno.Collections;
+
+namespace Fuse.Reactive
+{
+	/** Represents a reactive object-index look-up operation, with a computed index.
+		
+		Index can be either a number (for IArray lookups) or a string (for IObject lookups).
+	 */
+	public sealed class LookUp: Expression
+	{
+		public Expression Collection { get; private set; }
+		public Expression Index { get; private set; }
+
+		[UXConstructor]
+		public LookUp([UXParameter("Collection")] Expression collection, [UXParameter("Index")] Expression index)
+		{
+			Collection = collection;
+			Index = index;
+		}
+
+		public override IDisposable Subscribe(IContext context, IListener listener)
+		{
+			return new LookUpSubscription(this, context, listener);
+		}
+
+		sealed class LookUpSubscription: IDisposable, IObserver, IListener, ValueForwarder.IValueListener
+		{
+			IListener _listener;
+			LookUp _lu;
+
+			IDisposable _colSub;
+			IDisposable _indexSub;
+
+			public LookUpSubscription(LookUp lu, IContext context, IListener listener)
+			{
+				_listener = listener;
+				_lu = lu;
+				_colSub = _lu.Index.Subscribe(context, this);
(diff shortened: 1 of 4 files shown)


kusma:
@Duckers: please also fill in the "** Briefly describe changes and the motivation behind them here **"-part above...

Duckers:
@kusma done

Source/Fuse.Reactive.Expressions/LookUp.uno
```diff
@@ -0,0 +1,213 @@
+using Uno;
+using Uno.UX;
+using Uno.Collections;
+
+namespace Fuse.Reactive
+{
+	/** Represents a reactive object-index look-up operation, with a computed index.
+		
+		Index can be either a number (for IArray lookups) or a string (for IObject lookups).
```
kusma: Could you document what happens with out-of-bounds accesses? And maybe what happens when a fractional number is passed?

Source/Fuse.Reactive.JavaScript/Tests/UXExpressionsTest.uno
```diff
@@ -38,6 +38,89 @@ namespace Fuse.Reactive.Test
...
+			Assert.AreEqual("Hello foo : FOO!", e.t1.Value);
+
+			if defined (FUSELIBS_NO_TOASTS)
+			{
+				Diagnostics.DiagnosticReported += OnDiagnosticReported;
+				Diagnostics.DiagnosticDismissed += OnDiagnosticDismissed;
+
+				Assert.AreEqual(0, _currentDiagnostics.Count);
+
+				e.CallDec.Perform(); // index = -2 now, so that's = -1 and should give errro
```
kusma: typo: "error"

Source/Fuse.Reactive.JavaScript/Tests/UXExpressionsTest.uno
```diff
@@ -38,6 +38,89 @@ namespace Fuse.Reactive.Test
...
+
+		void OnDiagnosticReported(Diagnostic d)
+		{
+			_currentDiagnostics.Add(d);
+		}
+
+		void OnDiagnosticDismissed(Diagnostic d)
+		{
+			_currentDiagnostics.Remove(d);
+		}
```
kusma: This seems like it just repeats the logic of `RecordDiagnosticGuard`, except less robust. Why can't you just use that?

Source/Fuse.Reactive.Expressions/LookUp.uno
```diff
@@ -0,0 +1,213 @@
...
+						PushNewData(obj[key]);
+					}
+					else
+					{
+						SetDiagnostic("Object does not contain the given key '" + key + "'", _lu.Index);
+					}
+					return;
+				}
+
+				SetDiagnostic("Unsupported collection type: " + _collection, _lu.Collection);
```
kusma: Perhaps this should include some more information... How about "Trying to index an unsupported collection type: "... instead? Just saying "unsupported" is a bit hard to know exactly what went wrong...

kusma:
Sent from wrong repo

PR #62, opened 2017-05-28
Added internal parser in fuselibs that parses JSON into an IObject tree

Duckers:
This will be needed by future features. Internal only so no changelog/docs needed


kusma:
Without any users, I don't think this is useful. Can't we just introduce it when we have a user for it?

Duckers:
It is needed by #63, but splitting the PRs to keep features minimal

kusma:
That's not how pull-requests work. #63 already contains this stuff, so we don't need this PR.

PR #64, opened 2017-05-29
Handle iOS status bar changes properly

eeue56:
This fixes an issue where the "in call" statusbar would not correctly resize the window. It only resolves the issue if the user uses a `ClientPanel` - normal panels or raw elements do not scale accordingly, as they lack the TopFrame handlers.

## Test

There are two cases for testing, both only apply to iOS apps that have `ClientPanel`

- Start an application with a `ClientPanel` without being in-call
- toggle in-call on and off

- Start an application with `ClientPanel` with being in-call
- toggle in-call on and off

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
+++ b/Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
@@ -49,7 +49,13 @@ namespace Fuse.Controls
 
 		protected override float2 GetContentSize(LayoutParams lp)
 		{
-			if defined(iOS || Android)
+			if defined(iOS) {
+				// on iOS, we always treat the bar as though it's 40 points
+				var height = 40 / AppBase.Current.PixelsPerPoint;
+				var x = SystemUI.TopFrame.Size.X / AppBase.Current.PixelsPerPoint;
+				return float2(x, height);
+			}
+			else if defined(Android)
 			{
 				var x = SystemUI.TopFrame.Size / AppBase.Current.PixelsPerPoint;
 				return x;
--- a/Source/Fuse.Platform/iOS/SystemUI.uno
+++ b/Source/Fuse.Platform/iOS/SystemUI.uno
@@ -184,6 +184,7 @@ namespace Fuse.Platform
 				density);
 		}
 
+		static private float _modifier = 0;
 		static void _statusBarWillChangeFrame(Uno.Platform.iOS.uCGRect _endFrame, double animationDuration)
 		{
 			if (Lifecycle.State == ApplicationState.Uninitialized)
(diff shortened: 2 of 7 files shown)


Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
```diff
@@ -49,7 +49,13 @@ namespace Fuse.Controls
 
 		protected override float2 GetContentSize(LayoutParams lp)
 		{
-			if defined(iOS || Android)
+			if defined(iOS) {
+				// on iOS, we always treat the bar as though it's 40 points
+				var height = 40 / AppBase.Current.PixelsPerPoint;
```
mortoray: Why? This seems odd to have a hard-coded value here.

Source/Fuse.Platform/iOS/SystemUI.uno
```diff
@@ -211,6 +212,23 @@ namespace Fuse.Platform
 
 			var args = new SystemUIWillResizeEventArgs(SystemUIID.TopFrame, reason, endFrame, startFrame, animationDuration, 1);
 
+			Rect newFrameRect;
+			// when the endFrame size is > 40, we are streching to fill the space
+			if (endFrame.Size.Y > 40)
```
mortoray: The magic number 40 again. Use a `const` with a name, and explanation, or why we have this value.

Tests/ManualTests/ManualTestingApp/Singles/StatusBar.ux
```diff
@@ -0,0 +1,29 @@
+<Page Title="Status bar" ux:Class="StatusBar">
+	<InfoStack ux:Key="Info">
+		<p>This page is intended to test the status bar handling when a phone call is recieved with the app open.</p>
+	</InfoStack>
+	<StackPanel Margin="10">
```
mortoray: I think it'd be good to have something aligned to all the corners of the page. I guess the top frame already has the app header/buttons, but maybe put too circles in the bottom left/right corners to ensure they aren't getting clipped and/or distorted.

Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
eeue56: This is also the issue that got flagged last time - the tricky issue is that it's device specific. We can calculate it by getting the height of the top frame, but there is no consistent measurement we can get at runtime that way which will mean this functions as expected.

mortoray: It just seems very odd that a PR which is supposed to adjust to a dynamic change in the status bar is using a static size for it?!  This also seems not to be future proof, as I suspect Apple isn't guaranteeing a permanent 40pixel height to the status bar.

mortoray: Wouldn't fixing it at 40 mean we always lose that top bit of the screen, even when it's only 20 in size?

eeue56: Yep - this PR is very much work in progress. I was exploring alternative ways to calculate the right compensation but hadn't quite discovered it yet

mortoray:
Closing as abandoned.  Other status changes have come since then, so I can't isolate the issue this PR may still address.

PR #65, opened 2017-05-29
README: add section on how to integrate with fuse

kusma:
This adds a short section about how to configure fuse/uno to always pick up the checked-out fuselibs.

I'm not in love with this, but we need this information in some form. Some feedback would be greatly appreciated.

One of the big lacks, is that this doesn't really mention where to put the unoconfig-file, might be confusing to people who aren't used to dealing with unoconfig-files (which is probably most users). But I don't really know what the best advice would be.

We used to have something like this at the root of the repo, which made things work when using `Stuff/uno` from inside the fuselibs-repo. But that was removed by @mortend when the package-manager was introduced. We could consider adding it back...

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,20 @@ the UI framework used in [Fuse](https://www.fusetools.com/) apps.
 * `build.sh` runs stuff (which downloads uno if needed), and builds all packages.
 * `test.sh` runs all tests.
 
+
+### Fuse
+
+If you want to use a self-built fuselibs installation with an installed
+Fuse, you can do this through a `.unoconfig`-file, something like this:
+
+```
+Packages.SourcePaths += <path-to-fuselibs>/Source
+```
+
+You'll need to replace `<path-to-fuselibs>` above with the actual path to
+your fuselibs checkout.
+
+
 ## Contributing
 
 Please read [CONTRIBUTING.md](CONTRIBUTING.md) for details on our code of conduct, and the process for submitting pull requests to us.
--- a/README.md
+++ b/README.md
@@ -22,8 +22,11 @@ the UI framework used in [Fuse](https://www.fusetools.com/) apps.
 
 ### Fuse
 
-If you want to use a self-built fuselibs installation with an installed
-Fuse, you can do this through a `.unoconfig`-file, something like this:
+You may use a locally built copy of fuselibs with an installed copy of
+Fuse. This is done by creating a file named `.unoconfig` in either a Fuse
+project directory (applies to that project only), or in your home
+directory (applies to all projects). It should contain something like the
+following:
 
 ```
 Packages.SourcePaths += <path-to-fuselibs>/Source


README.md
```diff
@@ -19,6 +19,18 @@ the UI framework used in [Fuse](https://www.fusetools.com/) apps.
...
+
+### Fuse
+
+If you want to use a self-built fuselibs installation with an installed Fuse, you can do this through a `.unoconfig`-file, something like this:
+
+```
+Packages.SourcePaths += <path-to-fuselibs>/Source
+```
+
+You (obviously) need to replace `<path-to-fuselibs>` above with the actual path to your fuselibs checkout.
```
sebbert: The sarcasm here is (obviously) not needed 😉

README.md
```diff
@@ -19,6 +19,18 @@ the UI framework used in [Fuse](https://www.fusetools.com/) apps.
 * `build.sh` runs stuff (which downloads uno if needed), and builds all packages.
 * `test.sh` runs all tests.
 
+
+### Fuse
+
+If you want to use a self-built fuselibs installation with an installed Fuse, you can do this through a `.unoconfig`-file, something like this:
```
sebbert: This doesn't mention where to place the .unoconfig file. For instance:

If you want to use a locally built copy of fuselibs with an installed Fuse, create a file named `.unoconfig` in your home directory and add the following:

README.md
kusma: Hehe, good point ;)

README.md
kusma: @sebbert: Yeah, I wrote some rant about that in the cover-letter of this PR (above). I'm still a little bit torn if we should advice people about the home-directory, as that affects all projects. We could also advice them to put the `.unoconfig` in their project folder. And finally, we could just reinstate the top-level .unoconfig in the fuselibs repo, and tell people to run uno from inside there.

I'm not sure which option is better.

README.md
mortoray: Don't you need to somehow remove the old fuselibs path? If you simply do `+=` don't you get the checked out fuselibs and the installed fuselibs? I don't know which is resolved in that case.

kusma: @mortoray: I believe fuse doesn't use Packages.**Source**Path, and I believe uno prefers using the source-path. This is reported to work for users...

README.md
sebbert: @kusma Oh sorry, didn't see that – skipped right to the diff on this one.

I think we should document both options, as they can both be useful in different cases. We can always add a sentence about running uno from the repo directory if we decide to reinstate the repo-level .unoconfig.

How about something like this?

---

You may use a locally built copy of fuselibs with an installed Fuse. This is done by creating a file named `.unoconfig` in either a Fuse project directory (applies to that project only), or in your home directory (applies to all projects). It should contain something like the following:

kusma: @sebbert: Yeah, that sounds good. I was afraid of this quickly turning into a `.unoconfig`-tutorial, but your suggestion balances this nicely.

kusma:
@sebbert: Pushed fixes for your input :)

PR #68, opened 2017-05-29
UX parameter() function

Duckers:
UX expressions improvements. Added parameter(page) function which returns the routing parameter of the page parsed as an JSON string.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Common/Json.Parse.uno
+++ b/Source/Fuse.Common/Json.Parse.uno
@@ -1,0 +1,120 @@
+using Uno;
+using Uno.Data.Json;
+
+namespace Fuse
+{
+	public static partial class Json
+	{
+		/** Parses the given JSON string into a heirarchy of appropriate Uno objects.
+
+			The returned object can be a single value, or a tree composed of `IObject`, `IArray`, `double`, `string` and `bool` objects.
+		*/
+		internal static object Parse(string json)
+		{
+			return Convert(JsonReader.Parse(json));
+		}
+
+		/** Parses a list of JSON strings into a list of objects, using `Parse(string)`. */
+		internal static object[] Parse(string[] json)
+		{
+			var res = new object[json.Length];
+			for (int i = 0; i < res.Length; i++)
+			{
+				res[i] = Parse(json[i]);
+			}
+			return res;
+		}
+
+		static object Convert(JsonReader r)
+		{
+			switch (r.JsonDataType)
+			{
+				case JsonDataType.Object:
+					{
+						var keys = r.Keys;
+						var values = new object[keys.Length];
+						for (int i = 0; i < keys.Length; i++)
+						{
+							values[i] = Convert(r[keys[i]]);
+						}
+						return new Object(keys, values);
+					}
+				case JsonDataType.Array:
+					{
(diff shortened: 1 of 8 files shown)


Source/Fuse.Common/Json.Parse.uno
```diff
@@ -0,0 +1,120 @@
...
+				case JsonDataType.String: return (string)r;
+				case JsonDataType.Boolean: return (bool)r;
+			}
+			return null;
+		}
+
+		class Object : IObject
+		{
+			readonly string[] _keys;
+			readonly object[] _values;
```
mortoray: Why not a Dictionary?

Source/Fuse.Reactive.Expressions/Parameter.uno
```diff
@@ -0,0 +1,94 @@
...
+
+			protected override void OnNewOperand(object obj)
+			{
+				ClearDiagnostic();
+
+				UnsubscribeVisual();
+				
+				try
+				{
+					_visual = (Visual)obj;
```
mortoray: Doesn't this need to silently handle the `null` case, since it comes up in bindings and other functions?

Setting the value ot `null` I believe is the correct behaviour now... or do we just not update the value?

Source/Fuse.Reactive.JavaScript/Tests/UX/Expressions.Parameter.ux
```diff
@@ -0,0 +1,32 @@
...
+
+		function step2() {
+			router.goto("bar", {param: 300})
+		}
+
+		function step3() {
+			router.goto("foo", {param: 400, bar:"yeah"})
+		}
+
+		module.exports = { step1, step2, step3 }
```
mortoray: This JS syntax is unfortuantely supported on all our platforms. You must put `step1: step1, step2: step2, step3: step3`

Source/Fuse.Reactive.Expressions/Parameter.uno
Duckers: Notice the `try..catch`? It already "silently" handles it and converts it to a diagnostic.

Source/Fuse.Reactive.JavaScript/Tests/UX/Expressions.Parameter.ux
Duckers: This is test code

mortoray: I believe we have a perpetual plan to run the tests on iOS and Android though. Plus, test sample inevitably get copied into production code/examples.

Source/Fuse.Reactive.Expressions/Parameter.uno
mortoray: If the diagnostic actually silent? The `null` will show up often during rooting/unrooting.

kusma:
> Failed:  Fuse.Reactive.Test.UXExpressionsTest.ParameterFunction

Duckers:
Tests should be fixed now

PR #70, opened 2017-05-29
Fix: Marshal.TryConvertTo would throw exception instead of fail gracefully

Duckers:
This fixes and adds tests for cases where Marshal.TryConvertTo would throw exception instead of returning false as expected when an object is attempted converted to an incompatible type.

This caused a crashbug in the FuseCloud app


CHANGELOG.md
```diff
@@ -31,6 +31,8 @@
 
 ## ImageTools
 - Added supported for encoding/decoding images to/from base64 on DotNet platforms, including Windows and Mac OS X.
+## Bugfixes
```
kusma: You're missing a newline before this line here

PR #71, opened 2017-05-29
JavaScript Dependency Injection

Duckers:
> Requires new Uno from `feature-JSDI` to work: https://github.com/fusetools/uno/pull/1159

Added support for injecting UX expressions into `<JavaScript>` tags using the `dep` XML namespace. S

Example:
```
	<JavaScript>
		exports.foo = 123
	</JavaScript>
	<JavaScript dep:foo="{foo}">
		foo // this is now 123
	</JavaScript>
```
A script is not executed until all of its dependencies are available. If any of the dependencies change, the script is re-executed.

This has multiple use-cases:

* Accessing data from data context `dep:foo="{foo}"`
* Accessing properties synchronously `dep:SomeProp="{Property SomeProp}"`
* Dependency injection in tests

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


Source/Fuse.Reactive.JavaScript/RootableScriptModule.uno
```diff
@@ -31,10 +31,18 @@ namespace Fuse.Reactive
...
 
+		internal Dictionary<string, object> Dependencies;
+
 		protected override string GenerateArgs(Context c, ModuleResult result, List<object> args)
 		{
 			var argsString = base.GenerateArgs(c, result, args);
 
+			foreach (var dep in Dependencies) 
+			{
+				argsString += ", " + dep.Key;
```
mortoray: Don't we need a name check here to avoid producing somethign JS will break on?

Source/Fuse.Reactive.JavaScript/JavaScript.Dependencies.uno
```diff
@@ -0,0 +1,155 @@
...
+		}
+
+		void DisposeDependencySubscriptions()
+		{
+			if (_dependencies != null) 
+				for (var i = 0; i < _dependencies.Count; i++)
+					_dependencies[i].Unsubscribe();
+		}
+
+		void DispatchEvaluateIfDependenciesReady()
```
mortoray: I think you need to defer this otherwise you'll end up evaluating the module several times needlessly.

Consider a `<JavaScript deps:a="{a}" deps:b="{b}">` If both `a` and `b` change you'll end up instantiating the module twice, rather than just once. Since the values to a module are often related I think it's very likely that multiple dependencies will change at one time.

Source/Fuse.Reactive.JavaScript/ModuleInstance.uno
```diff
@@ -0,0 +1,110 @@
...
+
+		// UI thread
+		void SetDataContext()
+		{
+			_js.SetDataContext(_dc);
+		}
+
+		ModuleResult _moduleResult;
+
+		public void Dispose()
```
mortoray: Dipose needs to somehow block `Evaluate` from posting results. With the dependencies it's possible to have the same module evaluating in parallel multiple times if the deps change quickly. Only the final, current, evaluation of the module will be expected.

Source/Fuse.Reactive.JavaScript/RootableScriptModule.uno
Duckers: Stricly speaking yes, but there are tons of cases like this already and should be handled more generally, not in this site specifically.

The behavior of the current implementation will be acceptable imho, you will get an exception from trying to compile the script as expected. The error message might not be the friendliest, that's all

Source/Fuse.Reactive.JavaScript/JavaScript.Dependencies.uno
Duckers: It waits until all dependencies have yielded at least one value, so I don't understand your concern here

mortoray: It's when all dependencies get a new value in the same frame. The evaluation will be redone for each of the new values, not just once. This will lead to a significant over evaluation when multiple dependencies are used.

Duckers: > This will lead to a significant over evaluation when multiple dependencies are used.

In what use cases? I don't see that

Duckers: Note that each operator waits until all operands are ready before yielding its result. Hence, you will only get inappropriate double-evaluation if one or more of the operators double-yield without intending to do so.

mortoray: I'm talking about multiple variables in the same frame, which happens often.  So `dep:a="{a}" dep:b={b} dep:c={c}"`  if all of a, b, and c update in one frame this will revaluate three times in that frame.

Duckers:
This one seems to be stuck on travis-ci - anything we can do to unblock this @kusma ?

kusma:
@Duckers: It's building now, but it takes a while. The backlog on travis is big right now, it seems.

PR #73, opened 2017-05-29
JavaScript: optional explicit require() of UX symbols

Duckers:
Symbols declared with ux:Name, ux:Dependency or dep are in this PR also available to require() for <JavaScript> modules.

This allows us to write code that plays nicer with transpilers and linters. Using require for names declared in UX is optional, but may make the code more readable and maintainable, e.g. var router = require("router") over just using router with no declaration.

This PR contains:

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


mortoray:
I thought we were going to put these in some path to minimize conflicts and improve clarity, like `require("UX/foo")`. I think it'd be disconcerting that `require("email")` could refer to either a variable or the email module.

I wonder if that wouldn't also be better for linters et al to have a single pattern they can recognize.

Source/Fuse.Reactive.JavaScript/Tests/UX/OptionalExplicit.ux
```diff
@@ -0,0 +1,21 @@
+<Panel ux:Class="UX.OptionalExplicit">
+	<Router ux:Name="router" />
+	<JavaScript>
+		exports.foo = 123
+		// should not be reached below, as there is a bundled file that matches
+		// this path
+		exports.OptionalExplicit = 777
```
mortoray: Something seem off on this test. Do you mean that exports from a previous JS can be require'd in the next one? If yes, maybe you meant to do `require("foo")` in the next bit to show that it's working. Otherwise nothing shows you can `require` the exports at all here.

Source/Fuse.Reactive.JavaScript/Tests/DITest.uno
```diff
@@ -0,0 +1,31 @@
...
+using Fuse.Navigation;
+using FuseTest;
+
+namespace Fuse.Reactive.Test
+{
+	public class DITest : TestBase
+	{
+		[Test]
+		public void DIBasics()
+        {
```
mortoray: The indentation went wacky here (infiltration of spaces maybe).

Consider `using (var root = TestRootPanel.CreateWithChild(e))` instead. This form will include cleanup tests.

Duckers:
> I thought we were going to put these in some path to minimize conflicts and improve clarity, like require("UX/foo")

Well, I'm in a tango here, because what you are suggesting would be backwards incompatible for users who put their UX files in a `/UX` folder, while ntroducing a non-path-character as a delimiter would severly violate the CommonJS standard an possibly create problems with linters.

Source/Fuse.Reactive.JavaScript/Tests/UX/OptionalExplicit.ux
Duckers: It is visible below because of dependency injection:  `<JavaScript dep:bar="{foo}">`

mortoray:
I know adding a path might be backwards incompatible, but I think adding them in the root namepsace will cause us more grief in the long term. The clash between simple variable names and simple path names seems highly likely to happen.

Plus, as a code reviewer it'll be unclear what these random `require(foo)` statements do. With a prefix it makes it very clear where they are coming from.

Duckers:
Well, in that case I'm leaning towards a prefix instead, e.g. `ux:router`. This might confuse linters but at least there are no chances of backwards incompatible clashing.

mortoray:
I think `ux:` is a fine prefix. I think linters/checkers should be okay with this as `:` is a common path symbol and CommonJS doesn't actually define what's allowed.

mortoray:
Replaced by https://github.com/fusetools/fuselibs-public/pull/462/files

PR #74, opened 2017-05-30
Feature better bundle jsapi

Sunjammer:
The `FuseJS/Bundle` API is currently very limited and only really allows you to read text files. This PR adds features for listing, extracting and reading Bundles to JS in a more meaningful way.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


Source/FuseJS/Bundle.uno
```diff
@@ -120,6 +227,31 @@ namespace FuseJS
...
+			public ExtractClosure(string searchPath, string destinationPath, bool overwriteIfExists)
+			{
+				_searchPath = searchPath;
+				_destPath = destinationPath;
+				_overwrite = overwriteIfExists;
+			}
+			public string Invoke()
+			{
+				BundleFile bfile;
+				if(TryGetBundleFile(_searchPath, out bfile)){
```
mortoray: Bracket style

Source/FuseJS/Bundle.uno
```diff
@@ -134,5 +266,23 @@ namespace FuseJS
...
+			readonly string _filename;
+
+			public ReadBufferClosure(string filename)
+			{
+				_filename = filename;
+			}
+			public byte[] Invoke()
+			{
+				BundleFile bfile;
+				if(TryGetBundleFile(_filename, out bfile)){
```
mortoray: bracket style

Source/FuseJS/Tests/BundleTests.uno
```diff
@@ -0,0 +1,159 @@
...
+	*/
+
+	[extern(DOTNET) DotNetType("System.Reflection.Assembly")]
+	internal extern(DOTNET) abstract class BclAssembly
+	{
+		public extern static BclAssembly GetExecutingAssembly();
+		public extern string Location { get; }
+	}
+
+	class Utils{
```
kusma: space before the opening curly

Sunjammer:
Added ye olde tests, fixed the formatting

Source/FuseJS/Tests/BundleTests.uno
```diff
@@ -0,0 +1,159 @@
...
+using Uno.Testing;
+using Uno.Threading;
+using Uno.Text;
+using FuseTest;
+using Uno.Compiler.ExportTargetInterop;
+
+namespace FuseJS.Test
+{
+	/*
+		File path stuff cribbed from Fuse.FileSystem. That stuff should probably be made public/simpler for dogfooding.
```
kusma: Can't we use InternalsVisibleTo to get this without duplicating it?

Sunjammer: That'd expand this PR to affect FileSystem as well, that didn't seem good

mortoray: It's better to expand the PR than duplicate code. Duplicated code is pure evil. 😈

Source/FuseJS/Bundle.uno
```diff
@@ -53,14 +53,15 @@ namespace FuseJS
 			}
 			```
 		*/
-		static Future<byte[]> GetBuffer(object[] args)
+		public static Future<byte[]> ReadBuffer(object[] args)
```
mortoray: Don't make these public for testing, make them `internal` and use InternalsVisibleTo. The person that needs these visible for public use can make them public then.

Source/FuseJS/Tests/BundleTests.uno
```diff
@@ -0,0 +1,160 @@
...
+
+	
+	public class BundleTests : TestBase
+	{
+
+		[Test]
+		public void TestList()
+		{
+			var result = Utils.WaitForSuccessAndGetValue(FuseJS.Bundle.GetList());
+			Assert.IsTrue(result.Count()>0, "No bundles found");
```
mortoray: I'd really like to see that it checks for some expected bundle files. Bundling the wrong files is something that has happened before. :(

Source/FuseJS/Tests/BundleTests.uno
```diff
@@ -0,0 +1,160 @@
...
+				break;
+			}
+		}
+
+		[Test]
+		public void TestReadSync()
+		{
+			var files = Utils.WaitForSuccessAndGetValue(FuseJS.Bundle.GetList());
+			foreach(var f in files){
+				var result = FuseJS.Bundle.ReadSync(f.SourcePath);
```
mortoray: It'd be nice to see a byte comparison for at least one of the bundle files. Reading garbage by accident is not unheard of.

Source/FuseJS/Tests/BundleTests.uno
kusma: I agree with @mortoray here. It's a tiny change to the FileSystem-package, and we do similar stuff all the time...

Source/FuseJS/Tests/BundleTests.uno
Sunjammer: This API should really just return a list of every bundled file, no matter where it came from. As long as it returns a list of at least 1 file, the test is sound IMO.

mortoray: It does test something, but I don't think it tests enough. consider I ahve bundle files `one`, `two`, and `three`. This test will pass if the API returns just `two`, for example. It'd be a nice improvement for a limited implementation cost.

You don't have to check if the entire list is correct. Seed some files in the test project and ensure they are in the list.

Sunjammer:
Rewriting all the tests

mortoray:
@Sunjammer Sorry if I seem extra picky on the tests, it's just that JS interfaces are something that we've had many problems with silent regressions. As we go forward our test coverage must be further expanded. As I indicated though, if you can do high-level UX/JS tests they are more valuable than low-level unit tests. They usually require less maintenance as well since change the high-level API far less frequently.

Sunjammer:
I appreciate the pickiness, but the only reason I haven't done UX/JS tests is because I have no idea how and I can't seem to find a good approach in our existing tests...

Sunjammer:
Made nicer tests now. Will rebuild this PR to be much cleaner.

Source/FuseJS/Bundle.uno
Sunjammer: I don't see a strong rationale for this. InternalsVisibleTo is targeted, this opens for more widespread use and thus dogfooding. I'm not convinced making these anything but public does any good.

mortoray: It's a general rule we're following: everything is internal/private unless we have an actual use-case that requires it. Every `public` member increase our public API coverage and limits how we can refactor, and increases what we need to support.  This PR is about a jsAPI and thus having public Uno methods is not part of the use-case.

PR #77, opened 2017-05-31
Use a linked-list for Visual.Children with non-copying version locked iteration

mortoray:
This changes `Visual.Children` to use the new `ObjectList` (via a slightly improved `MiniList`). It has a few key improvements over a `List`:

- It's a linked list reducing the cost of inserting/removing children at specific locations
- It offers a locked version of the list for iteration, avoiding the need to do `ToArray()` or otherwise take a copy
- (minor) It reduces assigning objects during manipulation to avoid memory management overhead

This PR contains:
- [ ] ~Changelog~ - no user-visible changes
- [x] Documentation - on internal APIs, no user visible features
- [x] Tests


mortoray:
I forgot to migrate the tests on the final change. I'll do that soon.
Vegard at least confirms it doesn't make things slower (my profiling indicates at least far less memory activity in key locations).

Source/Fuse.Common/Internal/MiniList.uno
```diff
@@ -3,6 +3,15 @@ using Uno.Collections;
 
 namespace Fuse.Internal
 {
+	//An error with Uno handling the Enumerator<T> generic is preventing this from being a private enum to MiniList
+	//E2047: No implicit cast from Fuse.Internal.MiniList<T>.Mode to Fuse.Internal.MiniList<T>.Mode
```
kusma: Do we have a ticket for this?

Source/Fuse.Common/Internal/MiniList.uno
```diff
@@ -11,16 +20,26 @@ namespace Fuse.Internal
 	struct MiniList<T> : IList<T> where T : class
 	{
 		object _list;
+		ObjectList<T> AsList { get { return (ObjectList<T>)_list; } }
+		
+		MiniListMode _mode = MiniListMode.Empty;
```
kusma: This makes this container twice as large, and no longer very "Mini"... Why do we need this?

mortoray: The immediate reason was to fix the 1-item mode, which wasn't working correctly and wasn't clear to fix without the enum. (It could support null again now, but I haven't lifted the restriction yet)

The second reason is for performance. The dynamic type checking is costing us time. This is showing up in loops of the form `for (int i=0; i < list.Count; ++i) { foo(list[i]); }` as each iteraction involved several dynamic type casts and comparison. These lists and loops are occuring in critical code paths, so the chioce of speed outweighs the increase in size (which as a whole isn't much for the entire app: several thousand lists are created in total).

Source/Fuse.Common/Internal/MiniList.uno
mortoray: No. I didn't find the problem significant enough (doesn't touch a public API, minor annoyance) to be side-tracked coming up with a smaller reproduction at this time.

kusma: I think it would be better to file a ticket even if we don't have a minimal reproduction, in case we bother in the future. That way it would be possible to have everything related point to the same place.

Source/Fuse.Common/Internal/MiniList.uno
kusma: Well, this list is supposed to trade performance for size. You undo this here, rendering the semantics pointless. IMO, it would be better to switch away from MiniList<T> and to a normal List<T> in this case.

mortoray: It was never the goal to reduce total memory size. It's meant to efficiently handle the cases where 0 and 1 children are common, with an ever-so-slight performance loss for the 2+ cases. The efficiency is primarily in allocation overhead, not necessarily size, though in the 0 and 1 case it is still smaller than a List object. Don't forget it is a `struct` meaning the 0 and 1 case are supported without any allocation whatsoever.

kusma: Right, you're right, going back to the old discussion, this was about allocations, not size. But that makes the type confusingly named, and that gets slightly worse with this change. But that's not a functional problem, so maybe we can deal with that at some later point instead.

Source/Fuse.Common/Internal/MiniList.uno
mortoray: Added the issue

Source/Fuse.Common/Internal/MiniList.uno
mortoray: I've added a comment to the class.
If you can think of a better name it'd be easy enough to change as it's not a public API.

mortoray:
Rebased.

Source/Fuse.Common/Tests/Fuse.Common.Test/UnoEnumerable.Test.uno
```diff
@@ -0,0 +1,103 @@
...
+	public class UnoEnumerableTest : TestBase
+	{
+		[Test]
+		[Ignore("https://github.com/fusetools/uno/issues/1148")]
+		public void ForeachDispose()
+		{
+			var q = new SomeEnumerable<string>();
+			foreach (var s in q) { }
+			Assert.AreEqual(0, q.EnumCount);
+		}
```
kusma: These tests should be added to Uno instead, no? (the whole file)

Source/Fuse.Reactive.Bindings/Instance.uno
```diff
@@ -330,7 +330,7 @@ namespace Fuse.Reactive
 			
 			This list should only be modified via the InsertNew, RemoveAt, and RemoveAll functions.
 		*/
-		List<WindowItem> _windowItems = new List<WindowItem>();
+		Fuse.Internal.ObjectList<WindowItem> _windowItems = new Fuse.Internal.ObjectList<WindowItem>();
```
kusma: `using Fuse.Internal;` to get rid of the full namespaces?

Source/Fuse.Common/Tests/Fuse.Common.Test/UnoEnumerable.Test.uno
mortoray: Should I merge them with the Ignore to Uno?   I wrote them here originally since this was critically important to the `ObjectList`.

Source/Fuse.Reactive.Bindings/Instance.uno
mortoray: Maybe, I'm becoming reluctant to pull in namespaces for a single use lately as we have several name clashes across packages now, especially in the `Reactive.*` namespaces (where Instance is).

Source/Fuse.Common/Tests/Fuse.Common.Test/UnoEnumerable.Test.uno
mortoray: I've removed it. Added the test to the Uno issue for now.

Source/Fuse.Reactive.Bindings/Instance.uno
kusma: `Fuse.Internal.ObjectList` is getting repeated three times. I think that's enough to warrant a using-statement. Unless it actually cause issues, of course.

mortoray: It works with using Fuse.Internal.

PR #83, opened 2017-06-01
Scrollview invalidate visual while a pointer is pressed to avoid choppy behavior

Duckers:
When pointer input is uneven while scrolling, this would erraticaly flip caching preferences, leading to poor perceived performance in scrollviews. This fix makes sure visual is always invalidated while scrolling to keep cache preference consistent (and thus framerate consistent).

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests


Source/Fuse.Controls.ScrollView/Scroller.uno
```diff
@@ -276,6 +296,8 @@ namespace Fuse.Gestures
 
 		GestureRequest IGesture.OnPointerReleased(PointerReleasedArgs args)
 		{
+			StopInvalidateVisual();
```
mortoray: This is not necessary, the one in `OnLostCapture` is enough (it is always called).

Source/Fuse.Controls.ScrollView/Scroller.uno
```diff
@@ -198,9 +200,26 @@ namespace Fuse.Gestures
...
+			if (_pressed)
+				UpdateManager.RemoveAction(_scrollable.InvalidateVisual);
+			_pressed = false;
+		}
+
+		
 		
 		GestureRequest IGesture.OnPointerPressed(PointerPressedArgs args)
 		{
+			StartInvalidateVisual();
```
mortoray: This should go into the `OnCapture` function instead. There's no guarantee the scrollview gets capture when `OnPointerPressed` is called.

Source/Fuse.Controls.ScrollView/Scroller.uno
```diff
@@ -198,9 +200,26 @@ namespace Fuse.Gestures
 		{
 			get { return _scrollable == null ? GesturePriority.Lower : _scrollable.GesturePriority; }
 		}
+
+		bool _pressed;
+		void StartInvalidateVisual()
+		{
+			if (!_pressed)
+				UpdateManager.AddAction(_scrollable.InvalidateVisual);
```
mortoray: Hmm, don't we have a way to just prevent caching instead? Forcing invalidation could actually hurt performance if the user is idling.

mortoray:
Is this the cause of the mysterious jump in the animation?

Perhaps that's why I couldn't reproduce it in simpler tests, it has to interact with caching in a particular way.

Source/Fuse.Controls.ScrollView/Scroller.uno
kusma: @mortoray: We don't, not one that propagates up the tree like we would need here. I think we should make a mechanism like that instead. But maybe something like this could be an OK stop-gap solution.

kusma: @mortoray: I actually don't think this would hurt performance. We'll still be able to cache the contents of the scroll-viewer, which is a much better candidate IMO.

Maybe another alternative would be to make CachingMode=Never the default for ScrollView?

mortoray: CachingMode=Never might be too much. There are plenty of use-cases for ScrollView that don't involve the whole screen, or where the cotent remains static for a long-time. I think this stop-gap is probably better for now. 

If we do see this problem in other places we might consider attaching this to the interacction state of a control. Any control where `IsInteracting` is true would turn off caching.

Duckers:
I don't think this one is the cause of the mysterious jump, this is to avoid unwanted caching on frames where you accidentially get no pointer events (although doing a smooth motion). this happens on Android and DotNet

Source/Fuse.Controls.ScrollView/Scroller.uno
Duckers: Wontfix, @mortoray please approve PR

CHANGELOG.md
```diff
@@ -25,6 +25,10 @@
 - Updated the bundled libjpeg, libpng, Freetype, and SDL2 libaries for macOS to not use AVX instructions, since they are incompatible with the CPUs in some older Mac models. This fixes an issue with SIGILLs in native builds.
 
 ### Native
+## ScrollView performance 
+- Improved perceived ScrollView performance by preventing caching while pointers are pressed on them, avoiding inconsistent framerates.
+
+## Native
```
kusma: This looks like a bad merge.

Source/Fuse.Controls.ScrollView/Scroller.uno
```diff
@@ -198,7 +200,21 @@ namespace Fuse.Gestures
...
 		}
-		
+
+		bool _pressed;
+		void StartInvalidateVisual()
+		{
+			if (!_pressed)
+				UpdateManager.AddAction(_scrollable.InvalidateVisual);
+			_pressed = true;
+		}
```
kusma: Please add a newline between these two methods...

CHANGELOG.md
kusma: This section should be moved to the top of the changelog now that 1.1 is branched off.

CHANGELOG.md
```diff
@@ -42,6 +42,8 @@
 - Updated the bundled libjpeg, libpng, Freetype, and SDL2 libaries for macOS to not use AVX instructions, since they are incompatible with the CPUs in some older Mac models. This fixes an issue with SIGILLs in native builds.
 
 ### Native
+## ScrollView performance 
+- Improved perceived ScrollView performance by preventing caching while pointers are pressed on them, avoiding inconsistent framerates.
```
kusma: This is *still* a bad merge. Please actually fix it. This change should go in the top of the changelog, not down in the old part of some previous release.

CHANGELOG.md
```diff
@@ -41,8 +44,6 @@
 - Updated the bundled Freetype library on macOS to now (again) include both 32-bit and 64-bit symbols, which fixes an issue where .NET and preview builds would crash with a SIGILL at startup when running on older Mac models.
 - Updated the bundled libjpeg, libpng, Freetype, and SDL2 libaries for macOS to not use AVX instructions, since they are incompatible with the CPUs in some older Mac models. This fixes an issue with SIGILLs in native builds.
 
-### Native
-
 ## Native
```
kusma: You left the wrong version of "### Native" here.

nekronos:
@Duckers please fix the changelog, then we can merge this asap 🚨

PR #86, opened 2017-06-02
Feature observable pull

Sunjammer:
While reactive operators are lovely, there are scenarios when you need to aggregate the latest values from observables to use in callbacks responding to user interactions. An example is keeping the username and password in a sign-in form, but when the sign-in button is clicked you just have to trust username.value and password.value to be populated. 

This feels more fragile than, say, a `combineLatest`. However, `combineLatest` and other operators need a subscription. 
This creates a situation where creating temporary subscriptions or even storing values in separate fields in an onValueChanged handler becomes the boilerplate. Boring.

This PR adds `Observable.pull(module, function callback(observable){ ... })`, which formalizes the creation of such temporary subscriptions and results in (IMO) much nicer looking code for some of these scenarios:

```JS
const Observable = require('FuseJS/Observable')
const App = require('./src/App')
const username = Observable("")
const password = Observable("")
const pageIndex = Observable(0)

function onNextButton()
{
    switch(pageIndex.value)
    {
        case 0:
            pageIndex.value++
            break
        case 1:
            username.combineLatest(password, (username, password) => {
                    return {
                        username : username.trim(),
                        password : password.trim()
                    }
                })
                .pull(module, obs => App.signIn(obs.value))
            break
    }
}

module.exports = {
    username,
    password,
    onNextButton,
    pageIndex,
    nextButtonText : pageIndex.map(index => {
        switch(index){
            case 0:
                return "Next"
            default:
                return "Finish"        
        }
    }),
    canGoNext : pageIndex.combineLatest(username, password, (index, username, password) => {
        switch(index)
        {
            case 0:
                return App.validateUserName(username)
            case 1:
                return App.validatePassword(password)
        }
        return false
    })
}
```

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Reactive.JavaScript/FuseJS/Observable.js
+++ b/Source/Fuse.Reactive.JavaScript/FuseJS/Observable.js
@@ -1679,14 +1679,14 @@ Observable._getDataObserver = function(node, key)
 	return res;
 }
 
-Observable.prototype.pull = function(onResult){
+Observable.prototype.pull = function(module, onResult){
 	var self = this;
 	function resolver(obs)
 	{
 		self.removeSubscriber(this);
 		onResult(obs);
 	}
-	self.addSubscriber(resolver);
+	self._addDisposableSubscriber(module, resolver);
 }
 
 module.exports = Observable;


Source/Fuse.Reactive.JavaScript/FuseJS/Observable.js
```diff
@@ -1679,4 +1679,14 @@ Observable._getDataObserver = function(node, key)
 	return res;
 }
 
+Observable.prototype.pull = function(onResult){
+	var self = this;
+	function resolver(obs)
+	{
+		onResult(obs);
```
mortoray: Remove the subscriber first, then call the callback. This avoids an unintentional feedback loop.

Source/Fuse.Reactive.JavaScript/FuseJS/Observable.js
```diff
@@ -1679,4 +1679,14 @@ Observable._getDataObserver = function(node, key)
...
+		self.removeSubscriber(this);
+	}
+	self.addSubscriber(resolver);
```
mortoray: This needs to be attached to the `module` somehow to ensure it is cleanedup even if resolve is not called.

mortoray:
I'm not opposed to the idea but I don't see this use-case as supporting the need for pull. You could grab the `value` of the password and username observables directly with much less code than this.

I'd want to see an example where the snapshot is of something where the source OBservables are not available and where a snapshot actually makes sense.

Sunjammer:
IMO interacting with `.value` in general is a bad pattern. It feels like manual unboxing in a world where we have APIs that generally handle that for you. Fetching value directly without a temporal context is an inherently unsafe pattern.

mortoray:
I don't see a problem with interacting with the `value` of a source Observable. There's no way to use Observables without doing this. The Observable chain has to start somewhere, and on that source you have to modify the `value`. Something like `<TextInput Value="{username}">` also sets that value.

I want to see a use-case where the manual extraction is not possible and the `pull` behaviour still makes sense. Since at the moment you won't convince me your syntax is cleaner than:

```
App.signIn( {username.value, password.value} )
```

Sunjammer:
Just like you won't convince me `App.signIn( {username.value, password.value} )` looks safe to a user after we've educated them at length that you cannot guarantee observables contain values :)

Sunjammer:
Added a minor transform to the `combineLatest` observable. Already at this point we are doing things to value that will kill the app if value is undefined

mortoray:
But we do provide guarantees about values in source observables. We can't not provide guarantees about those. If you have `<TextInput Value="{username}">` we guarantee that the source Observable will get the input value.

Sunjammer:
That presumes knowledge about how the observable will be bound. That value could be coming from anywhere, an async file-reading api that fetches cached credentials, a textinput, whatever. We just don't know this. By using `pull` you have an absolute guarantee for access safety you can design around, rather than making presumptions about the surrounding codebase.

Sunjammer:
Note that this PR does not fix anything other than convenience and readability. This is not a missing _feature_ of Fuse, but it will aid learning and readability and IMO increase user confidence in their codebase.

mortoray:
That's my point, I want a use-case that my approach won't work for.

If you intend on this waiting for undefined items you'll need to test that it actually doesn't trigger until they are defined -- an empty value can be propagated.

Sunjammer:
Here's a strict and safe version without pull and with direct value access:

```JS
var un = username.value
if(un)
  un = un.trim()

var pwd = password.value
if(pwd)
  pwd = pwd.trim()

if(pwd&&un)
  App.signIn({ username : un, password : pwd})
else
  //Handle error

```

Sunjammer:
That makes presumptions about what the user wants. By pulling on a combineLatest you have guaranteed values.

mortoray:
Closing as abandoned. I'm not recording an issue for now as we believe the Model system should obviate the need for new Observable features.

PR #87, opened 2017-06-02
Gesture migration

mortoray:
This migrates several of the gesture and behaviors to use the gesture system. It allows these behaviours to overlap and behave together better.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
 # Unreleased
 
+## Gesture
+- The experimental `IGesture` interface has changed. 
+	* The `Significance`, `Priority` and `PriotityAdjustment` have been merged into the single `GetPriority` function.
+	* `OnCapture` is changed to `OnCaptureChanged` and provides the previous capture state
```
kusma: Other multi-level lists use two spaces for indent, please use that in the future instead...

Source/Fuse.Gestures/Internal/TwoFinger.uno
```diff
@@ -77,17 +75,16 @@ namespace Fuse.Gestures.Internal
...
 			_point[0].Down = _point[1].Down = -1;
-			Pointer.ReleaseCapture(this);
 			if (_begun)
 			{
-				_node.EndInteraction(this);
 				if (Ended != null)
 					Ended();
 			}
 			_begun = false;
+			_trackingKeyboard = false;
```
kusma: Should this be wrapped in a try-finally block, so we know that `_trackingKeyboard` becomes `false` even if an `Ended`-handler throws?

Thinking about it a bit more, events kinda feel a bit crappy for this, as one event-handler throwing is going to ruin the fun for all the others, leaving them in a limbo-state... Why does programming have to be so awful?

Source/Fuse.Gestures/Swipe.uno
```diff
@@ -210,11 +210,11 @@ namespace Fuse.Gestures
 		}
 		
 		internal static Selector GesturePriorityName = "GesturePriority";
-		GesturePriority _gesturePriority = GesturePriority.Lower;
+		GesturePriority _gesturePriority = GesturePriority.Low;
 		/**
 			The priority of the swiping gesture when competing with other gestures.
 			
-			The default is Higher. 
+			The default is Lower.
```
kusma: This change of behavior should be mentioned in the changelog.

Source/Fuse.Gestures/Internal/TwoFinger.uno
mortoray: I'll reorder. In general I want to call  the event handlers last. I don't know if the entire handling is exception safe though, Gesture as a whole might be an issue, let me check a bit there.

mortoray: I did something minimal, but looking at the overall code I can't really "fix" it here.

https://github.com/fusetools/fuselibs/issues/4236

PR #88, opened 2017-06-05
Made overall rendering less prone to frame drops during interaction 

Duckers:
Made overall rendering less prone to frame drops during interaction by preventing caching to bitmap of visuals while they are interacted with. Sometimes pointer input doesn't arrive each frame during interaction. This lead to the caching heuristic incorrectly thinking the visual was appropriate to cache, which in some scenarios could lead to expensive framebuffer operations, and no more 60 FPS.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Improved smoothness of interaction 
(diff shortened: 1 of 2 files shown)


Duckers:
Or, wait. Typically an element being interacted with is just animating its own transform. Its fine to cache in that scenario, the visual is not invalidated.

Actually, NOT caching could be disasterous

PR #89, opened 2017-06-06
ScrollingAnimation: Update animation if From or To changes

nekronos:
We have to update the animation state if From or To changes, if not the animation can get out of sync since it only updates when the scrollview is scrolls.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 - Added `reload` and `retry` JavaScript functions on `Image` to allow reloading failed images.
 - Fixed infinite recursion bug that could happen if a MemoryPolicy is set on a MultiDensityImageSource
 
+## ScrollingAnimation
+- Fixed issue where the animation could become out of sync if the properties on ScrollingAnimation were updated.
+
 ## macOS SIGILL problems
 - Updated the bundled Freetype library on macOS to now (again) include both 32-bit and 64-bit symbols, which fixes an issue where .NET and preview builds would crash with a SIGILL at startup when running on older Mac models.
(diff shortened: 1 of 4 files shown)


Source/Fuse.Controls.ScrollView/Triggers/ScrollingAnimation.uno
```diff
@@ -93,6 +93,8 @@ namespace Fuse.Triggers
 				_hasFrom = true;
 				if (!_hasRange)
 					_range = ScrollingAnimationRange.Explicit;
+
+				BypassSeek(OffsetScrollProgress);
```
mortoray: This needs a `if (_scrollable != null)`, which can only be true if the control is rooted and has a scrollview.

Source/Fuse.Controls.ScrollView/Triggers/ScrollingAnimation.uno
```diff
@@ -105,6 +107,8 @@ namespace Fuse.Triggers
 				_hasTo = true;
 				if (!_hasRange)
 					_range = ScrollingAnimationRange.Explicit;
+
+				BypassSeek(OffsetScrollProgress);
```
mortoray: As in other comment.

Source/Fuse.Controls.ScrollView/Triggers/ScrollingAnimation.uno
nekronos: Fixed

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+## ScrollingAnimation: Fixed issue where the animation could become out of sync if the properties on ScrollingAnimation were updated.
```
kusma: We usually have a short heading, and a bullet-list under it. Why should this entry be different?

nekronos:
Test added @kusma

PR #90, opened 2017-06-07
Reload failed images

mortoray:
Fixes https://github.com/fusetools/fuselibs/issues/4227
Adds Image.reload/retry functions

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests - except 4227 as there is no easy way to test that in an automated fashion

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,12 @@
 # Unreleased
 
+## Image
+- A failed to load Image with a Url will now try again when the Url is used again in a new Image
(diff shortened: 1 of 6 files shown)


kusma:
One question springs to mind; why do we even store failures in the cache?

mortoray:
@kusma This cache serves two purposes, where caching was not actually the most important part. It's the connecting of multiple images to the same source, ensuring if they use the same URL, or file, in multiple places the actual same image source is used. I don't think there's actually anything "cached" at the http level for these images. The source is simply marked failed, and since it's s shared source any new user will just see the failed state (previously, this corrects that).

PR #102, opened 2017-06-12
Scroller: Fix nullref in OnPointerMoved

nekronos:
If there is no gesture present, I assume that the GestureRequest should be Ignore

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.ScrollView/Scroller.uno
+++ b/Source/Fuse.Controls.ScrollView/Scroller.uno
@@ -245,6 +245,9 @@ namespace Fuse.Gestures
 
 		GestureRequest IGesture.OnPointerMoved(PointerMovedArgs args)
 		{
+			if (_gesture == null)
+				return GestureRequest.Ignore;
+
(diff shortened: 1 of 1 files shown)


mortoray:
Hmm, I think this fixes it in the wrong place. If the gesture has been disposed then the callback shouldn't be called at all (the code points to that as well). The fix would be in the Gesture code itself, to block the callback (otherwise this error repeats in all gesture users).  Maybe a flag in `Dispose` which is checked in the event handlers.

I wonder if this is fixed on my gestures branch. Doing a fix on the main branch would likely conflict.  Hmm, maybe just go with this for now and I'll fix this in the branch?

kusma:
@nekronos: Should this be against release-1.0 instead?

kusma:
I just realized, this should also have a changelog-entry. @nekronos, care to write one?

nekronos:
Yup!

PR #104, opened 2017-06-12
Attempt at unbreaking the map documentation

petterroea:


kusma:
I don't understand this. It's an *attempt*? Do we not know if it works or not?

kusma:
By the way, the change looks completely bogus to me...

petterroea:
Since this is in the fuselibs repo, i have no way as far as i know of verifying that it works.

kusma:
@petterroea: Surely you can run the docs-generator yourself, no?

petterroea:
Yes, but the way the docs-generator is set up, it pulls in precrunched data from fuselibs. I will investigate the possibility of building fuselibs-docs locally.

sebbert:
@kusma Verified locally in https://github.com/fusetools/fuselibs-public/pull/116

kusma:
@sebbert: I'm not sure I understand. #116 is different than this. Does that mean that this one did not work as-is?

sebbert:
@kusma This pull request (#104) produces invalid output. Since I can't push directly to the branch this is based on, I fixed the issue separately, verified the result, and put it in #106

kusma:
OK, then I'll close this one.

PR #105, opened 2017-06-13
Adjust batched element rect padding and some small cleanups/optimizations

yupferris:
While working to fix https://github.com/fusetools/fuselibs/issues/4213 (fixed proper in https://github.com/fusetools/fuselibs/issues/4213), @kusma and I found a bunch of code that could use some clarification/small optimizations.

Particularly we renamed a `_rect` to `AtlasRect` on `Fuse.Elements.ElementBatchEntry` for clarity, introduced the `DrawingOffset` field, and moved some duplicated calculation out of `ElementBatch`, as we already had done these calculations when placing elements inside their atlases, and could just pass the data along with these fields.

We also realized that we had an extra half-pixel more padding around all of the batched elements than we needed, so I went ahead and choked up on that. This is both a minor speed optimization and an additional safety net in case we run into issues of elements drawing outside of their allocated area in the batch atlas in the future.

### DIFF
--- a/Source/Fuse.Elements/Caching/ElementAtlas.uno
+++ b/Source/Fuse.Elements/Caching/ElementAtlas.uno
@@ -51,9 +51,8 @@ namespace Fuse.Elements
 
 		public bool AddElement(Element elm)
 		{
-			Recti cacheRect;
-			if (!Cache.GetCachingRect(elm, out cacheRect))
-				return false;
+			Recti cacheRect = ElementBatch.GetCachingRect(elm);
+
 			Recti rect;
 			if (!_rectPacker.TryAdd(cacheRect.Size, out rect))
 				return false;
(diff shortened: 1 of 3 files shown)


kusma:
@yupferris: You're supposed to actually fill out that PR-template ;)

yupferris:
@kusma: I went ahead and filled in the info portion of the template. As for the checkboxes, I'm not sure what our threshold is for putting things in the changelog - are these changes something that should go in there? If so, what might I write there in this case, as an example? Re documentation I don't think there's anything public-facing that changes, so that doesn't make much sense. And I'm just not sure how to test this smaller perimeter.

kusma:
@yupferris: As for the checkboxes, I usually remove those who doesn't apply, and check those who does (and is fixed). It's meant as a reminder to consider these things, really. In your case, I think none of these apply.

yupferris:
@kusma: sounds good. Removed them in this case.

PR #106, opened 2017-06-13
Fuse.Elements: set scissor-box if ClipToBounds

kusma:
If a batched element has ClipToBounds=True, but contents that overflow
it, we didn't set up the scissor-box. This lead to rendering outside the
allocated part of the element-atlas, and could corrupt the rendering of
other elements.

Add a test-case to make sure this doesn't regress in the future.

This PR contains:
- [x] Changelog
- [x] Tests


kusma:
This fixes fusetools/fuselibs#4213.

kusma:
@yupferris: In case you're wondering why I pushed a change, it was just about getting the changelog-stuff merge cleanly with another change on release-1.0

Source/Fuse.Elements/Tests/Caching.Test.uno
```diff
@@ -48,5 +48,30 @@ namespace Fuse.Test
...
 		}
+
+		[Test]
+		public void BorderIssue()
+		{
+			var p = new UX.Caching.BorderIssue();
+			var root = TestRootPanel.CreateWithChild(p, int2(32, 32));
+			root.CaptureDraw();
+
+			// check outside red rectangle
```
kusma: ugh, this comment is wrong :P

PR #113, opened 2017-06-14
Fixtests

mortoray:
Fixes some intermittendly failing tests.

NOTE: there's also a TODO comment added to Navigation. It's not really related, but i don't really want a PR just for it and I don't want to lose it either. 

This PR contains:
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Common/Tests/FuseTest/TestRootNode.uno
+++ b/Source/Fuse.Common/Tests/FuseTest/TestRootNode.uno
@@ -311,7 +311,7 @@ namespace FuseTest
 			one is connected to an Observable in another. The single StepFrameJS only takes care of
 			a single JS propagation, not the followup to the other element.
 			
-			This calls StepFrameJS multiple times to be correct in those situations.  THe number should
+			This calls StepFrameJS multiple times to be correct in those situations.  The number should
 			be the number of modules that need to be traversed. Don't just increase the number
 			until it works.
 		*/
--- a/Source/Fuse.Common/Tests/FuseTest/TestRootNode.uno
+++ b/Source/Fuse.Common/Tests/FuseTest/TestRootNode.uno
@@ -306,6 +306,22 @@ namespace FuseTest
 		}
 		
 		/**
+			If the test code involves multiple JavaScript elements that need to communicate between
+			each other, a single StepFrameJS may not be enough. This happens when an Observable in
+			one is connected to an Observable in another. The single StepFrameJS only takes care of
+			a single JS propagation, not the followup to the other element.
+			
+			This calls StepFrameJS multiple times to be correct in those situations.  THe number should
+			be the number of modules that need to be traversed. Don't just increase the number
+			until it works.
+		*/
+		public void MultiStepFrameJS(int count)
+		{
+			for (int i=0; i < count; ++i)
+				StepFrameJS();
+		}
+		
+		/**
 			Steps frames until the Deferred actions are all cleared. Guaranteed to step at least one frame.
 		*/
 		public void StepFrameDeferred()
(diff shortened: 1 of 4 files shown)


Source/Fuse.Navigation/WhileCanGoBack.uno
```diff
@@ -11,6 +11,8 @@ namespace Fuse.Navigation
 	public abstract class WhileHistoryTrigger : WhileTrigger
 	{
 		IBaseNavigation _context;
+		//TODO: This won't work due to being overidden in `OnUnrooted`. A good fix would be to remove this `public` part
+		//and alter the `Navigation` override to work with a `Node` (if that's even possible)
```
kusma: This seems unrelated to the rest of this patch...

Source/Fuse.Reactive.JavaScript/Tests/ObservableTest.uno
```diff
@@ -388,34 +388,34 @@ namespace Fuse.Reactive.Test
...
 		[Test]
-		[Ignore("https://github.com/fusetools/fuselibs-public/issues/107")]
 		public void MapTwoWayFloat3()
 		{
 			var p = new UX.Observable.MapTwoWayFloat3();
 			using (var root = TestRootPanel.CreateWithChild(p))
 			{
-				root.StepFrameJS();
+				//multi-step an Observable linked between two JavaScript elements involved
+				root.MultiStepFrameJS(2);
```
kusma: Maybe you could explain briefly in the commit message why two steps are needed?

Source/Fuse.Navigation/WhileCanGoBack.uno
mortoray: Ah, was it a combined commit? CAn we leave this one for now. Reversion is unlikely since it's just a comment.

Source/Fuse.Reactive.JavaScript/Tests/ObservableTest.uno
mortoray: I assumed the extended comment on MultiStepFrameJS would be sufficient. :)

Source/Fuse.Navigation/WhileCanGoBack.uno
kusma: I meant the other way around, so we'd keep the comment if we needed to revert the other part. But sure, it's not a big deal.

Source/Fuse.Common/Tests/FuseTest/TestRootNode.uno
```diff
@@ -306,6 +306,22 @@ namespace FuseTest
 		}
 		
 		/**
+			If the test code involves multiple JavaScript elements that need to communicate between
+			each other, a single StepFrameJS may not be enough. This happens when an Observable in
+			one is connected to an Observable in another. The single StepFrameJS only takes care of
+			a single JS propagation, not the followup to the other element.
+			
+			This calls StepFrameJS multiple times to be correct in those situations.  THe number should
```
kusma: `THe` -> `The`

Source/Fuse.Reactive.JavaScript/Tests/ObservableTest.uno
kusma: OK, fair enough.

mortoray:
@kusma Do you still need changes here, or can we merge this?

kusma:
@mortoray: We should fix that typo before merging, at least. The other stuff is OK as-is.

mortoray:
Yes, I thought I did, but I guess I didn't. Pushing now.

PR #114, opened 2017-06-14
Some rendering tests

kusma:
Our rendering-code does not have awesome test-coverage, is pretty fragile, and have a lot of requirements encoded in the same logic. So it's a good idea to have *some* testing going on here. Let's add a couple of tests.

While we're at it, do a few cleanups, like fixing the coordinate-system used in the GL.ReadPixels helpers.

This PR contains:
- [x] Tests


Source/Fuse.Common/Tests/FuseTest/TestFramebuffer.uno
```diff
@@ -0,0 +1,67 @@
...
+			Framebuffer = FramebufferPool.Lock(size, Uno.Graphics.Format.RGBA8888, true);
+		}
+
+		public void Dispose()
+		{
+			FramebufferPool.Release(Framebuffer);
+		}
+
+		/**
+			Reads a pixel form the last CaptureDraw.
```
mortoray: This comment should be updated.

Source/Fuse.Common/Tests/FuseTest/TestFramebuffer.uno
```diff
@@ -0,0 +1,67 @@
...
+					temp[2] / 255.0f,
+					temp[3] / 255.0f);
+			}
+			finally
+			{
+				GL.BindFramebuffer(GLFramebufferTarget.Framebuffer, prevFramebuffer);
+			}
+		}
+
+		public void AssertPixel(int2 pos, float4 expectedColor, float tolerace = float.ZeroTolerance, [CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, [CallerMemberName] string memberName = "")
```
mortoray: Would it make sense to put the `expectedColor` first as that is how the other `Assert.*` methods are ordered?

Source/Fuse.Common/Tests/FuseTest/TestFramebuffer.uno
mortoray: Perhaps mention the top-left coordinate origin.

Source/Fuse.Common/Tests/FuseTest/TestFramebuffer.uno
kusma: Yeah, good call. Updating!

Source/Fuse.Common/Tests/FuseTest/TestFramebuffer.uno
kusma: Good catch, will fix. I'll add a mention about the coordinate system.

kusma:
@mortoray: Fixed!

PR #116, opened 2017-06-15
Unbork MapView docs

sebbert:
Our markdown renderer seems to not handle code blocks inside blockquotes.
This works around the issue by removing the blockquote.

This PR contains:
- [ ] Changelog N/A
- [X] Documentation
- [ ] Tests N/A


kusma:
This seems to be a duplicate of #104...

Issue #126, opened 2017-06-22
Fuse.Launcher.Maps problem with query strings


eksperts:
Initially reported on community: https://www.fusetools.com/community/forums/howto_discussions/fusejsmaps_module_js

The root cause of the problem is that the InterApp call opens a link to `"http://maps.apple.com/maps?q=" + query ...` here: https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Launcher.Maps/Maps/MapsLauncher.uno#L56

There are two potential solutions:
1. URLencode the query on Uno side before calling `LaunchUriiOS`.
2. Change the docs and explain that using `encodeURI()` on the query string in JavaScript when calling the method is required.

The problem manifests itself on iOS, but the Android counterpart uses a `"geo:0,0?q=" + query` scheme, so it might be vulnerable too.

kusma:
When people pass URLs, it should be reasonable to expect them to understand that the URLs needs to actually be valid URLs, which includes correctly encoding them.

Encoding the URI on the Uno-side is both incorrect (correctly encoded URIs are going to get double encoded) and would be a regression.

Simply telling users to pass the URI through `encodeURI()` is not going to help much, as it creates the illusion that this is actually the correct solution, which it is not.

So either we would have to teach users about URI encoding (which I think is a bad solution, there's already a bunch of resources out there), or we just expect them to know what a valid URI is. I (obviously) think the latter is the better option.

In either case, this is a docs-issue, if an issue at all.

kusma:
(strike that previous comment, that was based on a misunderstanding of the issue)

Yeah, looks like we should do the first, it's unreasonable to expect people to know that the query goes into a URI.

kusma:
There's similar issues with latitude and longitude, by the way. They can both easily contain illegal characters.

kusma:
Also the Android version has problems with not encoding URIs properly.

kusma:
Fuse.Launcher.* seems to be riddled with these issues...

The most depressing thing is seeing [this](https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Launcher.Email/Email/EmailLauncher.uno#L58)... Someone *knew* about encoding-issues, but decided to only solve them for spaces.

PR #127, opened 2017-06-23
Launcher encoding issues

kusma:
The launchers doesn't always encode URIs correctly, which cause issues on iOS. See #126 for more details.

Unfortunately, I don't have an iOS-device at hand now, but I've verified that things still works as it should on Android. So if someone else could test this on iOS, that would be great.

This also fixes fusetools/fuse-docs#790

This PR contains:
- [x] Changelog
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## 1.1.0
 
+### Fuse.Launchers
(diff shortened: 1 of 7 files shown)


kusma:
@eksperts: Yeah, good call. I'll update again.

PR #133, opened 2017-06-26
Optimization: Don't check for GL errors all the time

Duckers:
These GL errors should never trigger (indicates bugs in fuselibs, or custom user GL code messing up things). Introduced `-DDEBUG_GL` if the user wants the old behavior or to enable debugging in the future.

Achieves a 0.5-1.0ms speedup on my android phone per frame in the test case (a big app), and ~5ms speedup on my desktop (!!)

This PR contains:
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,8 @@
 - Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
 - Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).
 - Backwards incompatible optimization change: The `protected virtual void Visual.OnInvalidateWorldTransform()` method was removed. The contract of this method was very expensive to implement as it had to be called on all nodes, just in case it was overridden somewhere. If you have custom Uno code relying on this method (unlikely), then please rewrite to explicitly subscribe to the `Visual.WorldTransformInvalidated` event instead, like so: Override `OnRooted` and do `WorldTransformInvalidated += OnInvalidateWorldTransform;`, Override `OnUnrooted` and to `WorldTransformInvalidated -= OnInvalidateWorldTransform;`, then rewrite `protected override void OnInvalidateWorldTransform()` to `void OnInvalidateWorldTransform(object sender, EventArgs args)`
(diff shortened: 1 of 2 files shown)


kusma:
I think that this should be `if defined(DEBUG)`, not `DEBUG_GL`. My concern is the same as in fusetools/uno#1194, that people won't be running using these flags. For instance, we don't run our tests, nor manual testing using this new flag.

Besides, adding a flag multiplies the number of configuration-combinations by two, which is scary for testing; either we leave some combinations untested, or we need to double our testing effort. Neither of those options are very appealing to me.

Duckers:
>I would approve this if we used if defined(DEBUG) instead, because that would benefit uno build --configuration=Release-builds. As is, I think it takes away valuable verification.

The performance impact is too severe - crippling CPU usage on my desktop machine making any profiling meaningless, and pretty bad on device too. I will argue strongly that checking for GL errors cannot be default debug behavior. Note that preview builds (which is the 99% user experience case) are also debug builds.

Regular user apps should not need to check for this. Only developers writing custom GL code, and people debugging fuselibs GL code need to turn this on.

A solution to increase testing value is to make sure we run all tests and manualtesting apps with `-DDEBUG_GL` to keep at least some of the verification value.

kusma:
I think the debate here is pretty much the same as on fusetools/uno#1194, and the same way I think the problem here is that if you expect a performant preview, then preview should not use debug builds.

Duckers:
Well, now we end up with double-error-checking everything. https://github.com/fusetools/uno/pull/1194 instruments every single GL call, while this one redundantly extra-checks on some random places in DrawContext.

Would you be happier if this PR simply deleted the calls to `CheckGLError`, delegating to the lower level GL API to perform due error checking in debug mode?

kusma:
@Duckers: The reason why there's *some* double-error-checking is that the support in Uno only works in .NET-targets. So we need *something* on mobile as well. This is not ideal, but it's better than nothing. I would prefer if we could make the Uno support work in C++ targets as well, but it seems a bit costly due to extra indirection. Removing these calls here removes safety-nets that we don't have elsewhere in some cases.

The fact that this kind of functionality has been added twice should be a good indication that it's useful.

PR #137, opened 2017-06-27
Added AppConfig class to allow configuring App.Background dynamically.

Duckers:
This can be used as an optimization.

This PR contains:
- [x] Changelog
- [X] Documentation

### DIFF
--- a/Source/Fuse.Nodes/AppConfig.uno
+++ b/Source/Fuse.Nodes/AppConfig.uno
@@ -1,0 +1,23 @@
+namespace Fuse
+{
+    /** Allows configuration of App.Background through data bindings .
+
+        As the `App` class is not a `PropertyObject` and doesn't have a containing data context,
+        properties like `Background` cannot be databound regularly. This behavior class
+        allows you to do that from a specific data context.
+
+        ## Example
+
+            <App>
+                ...
+                <AppConfig Background="{foo}" />
+    */
+    public class AppConfig: Behavior
+    {
+        public float4 Background 
+        {
+            get { return AppBase.Current.Background; }
+            set { AppBase.Current.Background = value; }
+        }
+    }
+}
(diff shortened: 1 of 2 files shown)


Source/Fuse.Nodes/AppConfig.uno
```diff
@@ -0,0 +1,23 @@
...
+            <App>
+                ...
+                <AppConfig Background="{foo}" />
+    */
+    public class AppConfig: Behavior
+    {
+        public float4 Background 
+        {
+            get { return AppBase.Current.Background; }
+            set { AppBase.Current.Background = value; }
```
kusma: This is not really awesome... If there's two `AppConfig`-instances, they start overwriting each other's settings...

nekronos: Mhm 🤔 Could we make a special case for the Background property on App?

yupferris: That was my reaction as well, but I have no idea what a better alternative might be.

kusma: My thinking is that we should make change `App` so it is a normal Node. This would be a big change, though.

yupferris: Yeah that sounds like the right approach long-term, but appears to be a big change indeed.

yupferris: Alternatively, is it absolutely necessary that App maintains the clear color for the app, or is there another construct that could be better suited for this? (ignoring backwards-compatibility)

kusma: @yupferris: Well, this is kind of a hole we've dug ourselves into. The color really belongs to an implicit `GraphicsView` directly inside the App-tag. However, that is not exposed, so the user can't easily change that. That `App` has a `Background`-property is meant to be a way of exposing this to the API-user.

mortoray:
Closing as abandoned.

PR #139, opened 2017-06-28
Several template mini-features

mortoray:
Adds a few improvements to using templates.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


mortoray:
I just removed the `ITemplateSource.Templates {get}` interface method since I couldn't find a workaround to  the Uno problem (even giving it a new name). Nothing currently used it (this is a new interface), so whoever first needs it will have to deal with this.

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Unreleased
 
+## Templates
+- Triggers may now use templates which will be instantiated and added to the parent when active (like a node child).
+	<WhileActive>
+		<Circle ux:Generate="Template" Color="#AFA" Width="50" Height="50" Alignment="BottomRight"/>
+	</WhileActive>
+- Added templates to `NodeGroup`, which can now be used in `Each.TemplateSource` and `Instance.TemplateSource`
+- `Each`, using `TemplateSource`, will no longer respond to template changes after rooting. This was done to simplify the code, and to support alternate sources, and is a minor perf improvement. It's not likely to affect any code since it didn't work correctly, and there's no way in UX to modify templates after rooting.
```
kusma: I worry a bit about how this will affect preview... Does this mean preview will need to know, and keep track of all instantiations? (I know there's already some preview-specific template handling, but I'm not sure if that covers this)

Source/Fuse.Common/Internal/MiniList.uno
```diff
@@ -257,10 +258,10 @@ namespace Fuse.Internal
...
 			public void Reset()
-			{
-				//it's easier not to support this since it would involve tracking the initial iterator state (added overhead)
-				//or it would incorrectly iterate the current source, not the one at the time of creation
-				throw new Exception( "Reset not supported" );
+			{	
+				_first = true;
+				if (_mode == MiniListMode.List)
+					_iter.Reset();
 			}
```
kusma: Is the warning above about this change?

Source/Fuse.Nodes/NodeGroup.uno
```diff
@@ -2,16 +2,21 @@ using Uno;
...
 {
 	/**
 		A common base class that adds nodes and resources to the parent node while active.
 		
 		[subclass Fuse.NodeGroupBase]
+		
+		Be aware there is no ordering between the Nodes, Resources, and Templates. These are each independent lists which have their own order.
 	*/
-	public abstract class NodeGroupBase : Behavior
+	public abstract class NodeGroupBase : Behavior
```
kusma: This added space seems pointless

Source/Fuse.Nodes/NodeGroup.uno
```diff
@@ -30,9 +35,17 @@ namespace Fuse
...
+
+		TemplateSourceImpl _templates;
+		
+		[UXContent]
+		public IList<Template> Templates { get { return _templates.Templates; } }
+		public Template FindTemplate(string key) { return _templates.FindTemplate(key); }
 		
 		internal NodeGroupBase() { }
 		
+		protected bool UseTemplates = true;
```
kusma: Maybe you should mark this `readonly` and pass the value through the constructor above instead?

Something like:

```uno
readonly bool _useTemplates;
internal NodeGroupBase(bool useTemplates)
{
	_useTemplates = useTemplates;
}
```

That way it's clear to the reader that this can't change during the lifetime of the instance. It should also allow us to keep the actual field private.

Source/Fuse.Nodes/NodeGroup.uno
```diff
@@ -124,6 +137,7 @@ namespace Fuse
 		//used to prevent double-adding and double-removing if a derived class changes the UseContent
 		//state during rooting
 		bool _contentAdded;
+		MiniList<Node> _addedNodes;
```
kusma: Did you consider simply using an array of nodes here instead? It seems trivial to compute the size known size here, and we can null it out when there's nothing added. Should be both small and simple...

```uno
var nodeCount = _nodes.Count + (UseTemplates ? _templates.Count : 0);
_addedNodes = new Node[nodeCount];
```

We can probably survive null-references in the list in the failure-case for non-node-templates.

Alternatively, we can just append, and at the end, if the actual size did not meet the computed size, we can shrink the list. That way we get ideal memory-allocation in the non-error-code, and don't have to deal with null-references in the rest of the code. But we pay a small price in the error-case, which I don't think is too bad.

Source/Fuse.Reactive.Bindings/Instance.uno
```diff
@@ -145,10 +145,15 @@ namespace Fuse.Reactive
...
-		public Visual TemplateSource
+		public ITemplateSource TemplateSource
 		{
-			get; set;
+			get { return _weakTemplateSource; }
+			set { _weakTemplateSource = value; }
 		}
+		//https://github.com/fusetools/fuselibs-public/issues/135
+		[WeakReference]
+		ITemplateSource _weakTemplateSource;
```
kusma: Is the comment about the `[WeakReference]` avoiding the memory-reference-loop in that issue? If so, maybe you could spell out what you mean?

Something like:
```uno
/*
  To avoid memory-loops between the template-source and the
  instantiatior, we keep a weak reference here until rooting
  is performed, and only keep a strong reference until we're
  unrooted again.
*/
```

I'm not a huge fan of pointing the code to the issue-tracker (generally I think information should flow in the other direction), as it makes the code harder to read if you're offline. There's obviously cases where we can't avoid it (tests disabled due to issues for instance), but I don't think this one of them.

This is just my personal preference, though. Take it or leave it ;)

CHANGELOG.md
mortoray: This refers to the list of templates itself, nothing ever responded to changes inside a template before. 

Unless preview had special support for `Each` I doubt this would make an impact. Template changes require a reify of the parent node at a minimum -- this would be throughout fuselibs, not just in `Each`.

The mechanism was also somewhat broken before, only working in some cases (not the most common ones either). Thus I felt it a safe change to make.

Source/Fuse.Common/Internal/MiniList.uno
mortoray: No, this is just a somewhat unrelated change to fix a few test cases that somehow managed to use Reset. It wasn't too hard to give Reset a reasonable behaviour, so I did.

Source/Fuse.Nodes/NodeGroup.uno
mortoray: I'll remove the lone space.

Source/Fuse.Nodes/NodeGroup.uno
mortoray: Agreed.

Source/Fuse.Nodes/NodeGroup.uno
mortoray: I'll try it as a separate commit after fixing the other stuff.

Source/Fuse.Reactive.Bindings/Instance.uno
mortoray: Well, since you wrote a nice comment I'll use it. :)

I guess if it's short it makes sense, but sometimes I prefer links since the issue can be long and hard to understand (I don't think pasting stories into the code makes sense).

Source/Fuse.Nodes/NodeGroup.uno
```diff
@@ -30,8 +35,25 @@ namespace Fuse
...
+		
+		[UXContent]
+		public IList<Template> Templates { get { return _templates.Templates; } }
+		public Template FindTemplate(string key) { return _templates.FindTemplate(key); }
+		
+		[Flags]
+		internal enum ConstructFlags
+		{
+			None = 0,
+			DontUseTemplates = 1 << 0,
```
kusma: Nice, even cleaner than what I suggested :)

Source/Fuse.Nodes/NodeGroup.uno
```diff
@@ -194,14 +200,17 @@ namespace Fuse
 					if (n == null)
 					{
 						Fuse.Diagnostics.InternalError( "Template contains a non-Node", this );
-						continue;
+						n = new EmptyNode();
```
kusma: Neat trick :)

CHANGELOG.md
```diff
@@ -1,5 +1,13 @@
...
+- A memory leak was fixed by changing `Instantiator.TemplateSource` to a WeakReference. Only if you assigned this property a tempoary value in Uno would this change impact your code.
```
kusma: There should be a space between this and the next heading.

Source/Fuse.Common/Internal/MiniList.uno
```diff
@@ -257,10 +258,10 @@ namespace Fuse.Internal
...
 
 			public void Reset()
-			{
-				//it's easier not to support this since it would involve tracking the initial iterator state (added overhead)
-				//or it would incorrectly iterate the current source, not the one at the time of creation
-				throw new Exception( "Reset not supported" );
+			{
```
kusma: Could you drop the trailing tab here?

PR #144, opened 2017-06-29
ElementAtlas: Don't set rendertarget redundantly

Duckers:
PinAndValidate sets rendertarget and calls FillFramebuffer, but often nothing is actually drawn. This postpones setting the rendertarget until actually needed.

Causes significant speedups on some platofrms.

This PR contains:
- [x] Changelog

### DIFF
--- a/Source/Fuse.Elements/Caching/ElementAtlas.uno
+++ b/Source/Fuse.Elements/Caching/ElementAtlas.uno
@@ -147,18 +147,10 @@ namespace Fuse.Elements
 			if (_invalidElements > 0)
 			{
 				Rect scissorRectInClipSpace = GetScissorRectInClipSpace(dc);
-				dc.PushRenderTarget(fb);
-
 				bool drawAll = _invalidElements == _elements.Count;
-				if (drawAll)
-				{
-					dc.Clear(float4(0), 1);
-					FillFramebuffer(dc, true, scissorRectInClipSpace);
-				} else
-					FillFramebuffer(dc, false, scissorRectInClipSpace);
-
-				dc.PopRenderTarget();
+				FillFramebuffer(dc, fb, drawAll, scissorRectInClipSpace);
 			}
+
 			return fb;
 		}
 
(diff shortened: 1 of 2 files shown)


Source/Fuse.Elements/Caching/ElementAtlas.uno
```diff
@@ -167,8 +159,12 @@ namespace Fuse.Elements
 			_framebuffer.Unpin();
 		}
 
-		void FillFramebuffer(DrawContext dc, bool drawAll, Rect scissorRectInClipSpace)
+		void FillFramebuffer(DrawContext dc, framebuffer fb, bool drawAll, Rect scissorRectInClipSpace)
 		{
+			// Changing the framebuffer is expensive. This loop actually doesn't draw anything in
+			// common case. Don't push the render target unless we actually have to render something.
```
kusma: If this gets here without drawing anything, I think something higher level failed. So I think this is fixing the symptom, rather than the actual issue. The `_invalidElements > 0`-check around the calls to FillFramebuffer should have avoided this case.

Duckers: I thought so too initially, but no. The logic here is actually that we don't paint elements until they are *visible in the viewport*, yet we reserve space in the atlas. So the atlas can be used even though it has several invalid elements.

kusma: Ah, I think I see why. There's elements that are invalid, but are off-screen. So the cache tries to avoid updating them until it has to... That makes the early-out mechanism not work as it should.

Duckers: Indeed, so I think the commit is good as it stands?

Duckers: I'll add a comment to make it clear

kusma: @Duckers: Yeah, I think so. At least for now. Maybe it would be an idea to fix the underlying issue, but I don't have a great strategy for this that doesn't involve linear scanning yet. Perhaps some annotated KD-tree or something, but that's a bigger project.

PR #146, opened 2017-06-29
Bugfix iOS multitouch

nekronos:
For handling input in native on iOS we used [addTarget:action:forEvent](https://developer.apple.com/documentation/uikit/uicontrol/1618259-addtarget?language=objc) to listen for all touch events. During multitouch it looks like there is unexpected semantics or a bug here since all touch events stops when a finger is lifted.

This PR introduces a new way to handle touch events where the case above is not present. It also add the option to opt-out of automatic handling of touch events. Allowing the user to deal with the native-fuselibs interop if needed.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Optimizaitons
+- Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
+
 ## Attract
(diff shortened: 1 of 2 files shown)


mortoray:
I think a page in the ManualTestApp would be good here. Something that demonstrates the multi-touch is working, perhaps something which shows a circle around each touched finger, and moves with it?

nekronos:
Sounds like a good idea. I'll look into it

nekronos:
github broke my pr

PR #147, opened 2017-06-29
Added WhileScrolling trigger

igorls:
Fuse was missing a trigger to detect when the user is interacting with a ScrollView
Uno class was added to the Fuse.Triggers namespace. 
the trigger is active while a parent ScrollView element is being scrolled

I added a small example in the uno file comments

This PR contains:
- [ ] Changelog
- [x] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
+++ b/Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
@@ -26,6 +26,7 @@ namespace Fuse.Triggers
 
 		This example will change the @ScrollView background color while the user scrolls.
 		The trigger will revert when the ScrollView stops, even if the user is still touching the screen.
+		If the user lift of his finger and the ScrollView still got some momentum, the trigger will remain active.
 	*/
 	public class WhileScrolling : WhileTrigger
 	{
--- a/Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
+++ b/Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
@@ -1,0 +1,95 @@
+using Uno;
+using Uno.Collections;
+using Fuse;
+using Fuse.Controls;
+using Fuse.Triggers;
+using Fuse.Elements;
+using Fuse.Gestures;
+using Uno.UX;
+
+namespace Fuse.Triggers
+{
+	/**
+		Active while the user is scrolling the parent @ScrollView.
+
+			<ScrollView>
+				<SolidColor ux:Name="scrollBg" Color="#2A2A2A"/>
+				<StackPanel  ItemSpacing="15">
+					<Each Count="30">
+						<Rectangle  Width="200" Height="40" Color="#F0F1E7"/>
+					</Each>
+				</StackPanel>
+				<WhileScrolling>
+					<Change scrollBg.Color="#838383" Duration="0.5"/>
+				</WhileScrolling>	
+			</ScrollView>
+
+		This example will change the @ScrollView background color while the user scrolls.
+		The trigger will revert when the ScrollView stops, even if the user is still touching the screen.
+	*/
+	public class WhileScrolling : WhileTrigger
+	{
+		ScrollViewBase _scrollView;
+		Element _parent;
+
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+
+			if (Parent is Element) {
+				_parent = Parent as Element;
+			}
+			else {
+				Fuse.Diagnostics.UserError( "Parent must be an Element", this );
+				return;
+			}
+
+			_scrollView = Parent.FindByType<ScrollViewBase>();
+			if (_scrollView == null) {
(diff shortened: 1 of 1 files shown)


Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
```diff
@@ -0,0 +1,95 @@
...
+	public class WhileScrolling : WhileTrigger
+	{
+		ScrollViewBase _scrollView;
+		Element _parent;
+
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+
+			if (Parent is Element) {
```
kusma: We [don't use egyptian braces](https://github.com/fusetools/fuselibs-public/blob/master/Documentation/CodingStyle.md#uno--c) in Uno code.

Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
```diff
@@ -0,0 +1,95 @@
...
+				_parent = Parent as Element;
+			}
+			else {
```
kusma: Same as above, we don't use egyptian braces.

Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
```diff
@@ -0,0 +1,95 @@
...
+		protected override void OnUnrooted()
+		{
+
+			if (_scrollView != null)
+			{
+				_scrollView.ScrollPositionChanged -= OnScrollPositionChanged;
+
+			}
+
+			_scrollView = null;
```
kusma: You should also null out `_parent` here, otherwise our ARC will keep that object around until this one is deleted. For instance if an object gets rooted correctly, unrooted, then triggering the "parent is not Element" error-case in `OnRooted`, we'll be left with the old `_parent`.

Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
```diff
@@ -0,0 +1,95 @@
...
+	{
+		ScrollViewBase _scrollView;
+		Element _parent;
+
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+
+			if (Parent is Element) {
+				_parent = Parent as Element;
```
kusma: You're effectively casting twice here, which is a bit costy in terms of performance. I generally prefer this:
```uno
var foo = bar as Foo;
if (foo != null) 
{
	// do something with foo
}
```

However, in this case I think the following is better:
```uno
_parent = Parent as Element;
if (_parent == null)
{
	Fuse.Diagnostics.UserError("Parent must be an Element", this);
	return;
}
```
...because overwriting `_parent` with `null` here seems harmless.

Source/Fuse.Controls.ScrollView/Triggers/WhileScrolling.uno
```diff
@@ -0,0 +1,95 @@
...
+			_scrollView.ScrollPositionChanged += OnScrollPositionChanged;
+		}
+
+		protected override void OnUnrooted()
+		{
+
+			if (_scrollView != null)
+			{
+				_scrollView.ScrollPositionChanged -= OnScrollPositionChanged;
+
```
kusma: This empty line seems a bit pointless. I would even drop the braces in this case, but that's a matter of taste that differs a bit between fuselibs developers, so don't take that too seriously.

igorls:
Thanks for the review, just fixed the formatting stuff and nulled out the _parent element as requested.

@kusma if the user scrolls and let his finger off the screen and the ScrollView is still moving the trigger is still active. I don't know if I got your comment right? Maybe you can suggest another name?

igorls:
@mortoray do you mean to access the CheckNeedUpdated function from the WhileScrolling class? I don't believe that is possible. Or to implement the animation tracker as on the function you mentioned? in that case are you refering to this logic? (Scroller.uno(153))
```
bool needUpdated = _region != null && !_region.IsStatic && IsRootingStarted;
```

mortoray:
@igorls You will need to modify the `Scroller` class to get this working. The approach I'd take is to add a `IsAnimating`  public proeprty to `Scroller`, along with a `IsAnimatingChanged` event. This is most likely modified/triggered inside the `CheckNeedUpdated` function.

The high-level `WhileScrolling` trigger can then monitor both `IsAnimating` on `Scroller` and `IsInteracting` on the `ScrollView`. Look at `WhileInteracting` to see the logic for `IsInteracting`.

`IsAnimating` should be an internal property, and it's okay if `WhileScrolling` accesses it.

kusma:
@igorls: No, I was talking about the opposite case (which seems to be the same as @mortoray was talking about, AFAIK), when the user scrolls, then stops moving their finger (but still keeps it pressed). At least from one point of view, the user is still "Scrolling", they're just not changing the scroll-position right now.

mortoray:
No updates have come.

PR #148, opened 2017-06-29
Optimize WindowToLocal by using FastMatrix to transform vectors

Duckers:
The WindowToLocal function is called frequently while processing pointer events for hit testing. The old implementation did a full float4x4 vector transform for each node in the view heirarchy. This optimization delegates to FastMatrix to transform the vector, taking a fast path in the common case of translation-only.

This PR contains:
- [x] Changelog

### DIFF
--- a/Source/Fuse.Nodes/Visual.Transform.uno
+++ b/Source/Fuse.Nodes/Visual.Transform.uno
@@ -256,8 +256,7 @@ namespace Fuse
 			if (HitTestTransform == HitTestTransformMode.LocalPoint)
 			{
 				var parentCoord = (Parent == null) ? windowCoord : Parent.WindowToLocal(windowCoord);
-
-				return Vector.TransformCoordinate(parentCoord, LocalTransformInverse);
+				return LocalTransformInverseInternal.TransformVector(float3(parentCoord,0)).XY;
 			}
 			else
 			{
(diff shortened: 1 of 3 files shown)


Source/Fuse.Nodes/Visual.Transform.uno
```diff
@@ -256,8 +256,7 @@ namespace Fuse
 			if (HitTestTransform == HitTestTransformMode.LocalPoint)
 			{
 				var parentCoord = (Parent == null) ? windowCoord : Parent.WindowToLocal(windowCoord);
-
-				return Vector.TransformCoordinate(parentCoord, LocalTransformInverse);
+				return LocalTransformInverseInternal.TransformVector(float3(parentCoord,0)).XY;
```
mortoray: Might it not be faster to offer a `TransformVectorXY`? It's just a minor improvement, but simple enough to add.

Duckers: Good point - it can also be significantly faster in the float2 x float4x4 case, as we can throw away a lot of the calculations.

kusma:
I have no idea what happened here. It seems both appveyor and travis thinks the branch this PR came from was in the public repo (and not in @Duckers' fork). Perhaps Github deployed a bad update?

Source/Fuse.Nodes/Visual.Transform.uno
kusma: @Duckers: Are you planning on doing anything about this part, or are we saving this for a follow-up later? Just to be clear, I'm fine with either.

PR #149, opened 2017-06-29
Optimized how WorldTransform is invalidated

Duckers:
Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).

This required a new implementation strategy for emitting the WorldTransformInvalidated event, as we are no longer traversing the subtree eagerly to generate these events.

This PR contains:
- [x] Changelog

@mortoray - please evaluate carefully whether we have enough test coverage to safely make this optimization. If not, any help to extend the test coverage is highly appreciated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,8 +5,6 @@
 - Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).
 - Backwards incompatible optimization change: The `protected virtual void Visual.OnInvalidateWorldTransform()` method was removed. The contract of this method was very expensive to implement as it had to be called on all nodes, just in case it was overridden somewhere. If you have custom Uno code relying on this method (unlikely), then please rewrite to explicitly subscribe to the `Visual.WorldTransformInvalidated` event instead, like so: Override `OnRooted` and do `WorldTransformInvalidated += OnInvalidateWorldTransform;`, Override `OnUnrooted` and to `WorldTransformInvalidated -= OnInvalidateWorldTransform;`, then rewrite `protected override void OnInvalidateWorldTransform()` to `void OnInvalidateWorldTransform(object sender, EventArgs args)`
 
-
-
 ## Attract
 - Added the `attract` feature, which was previously only in premiumlibs. This provides a much simpler syntax for animation than the `Attractor` behavior.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Optimizaitons
 - Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
+- Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).
 
 ## Attract
 - Added the `attract` feature, which was previously only in premiumlibs. This provides a much simpler syntax for animation than the `Attractor` behavior.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## Optimizaitons
 - Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
 - Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).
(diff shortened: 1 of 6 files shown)


Duckers:
I've found at least one regression; text inputs no longer show up on android. Still work though, the native view is just not visible/where its supposed to be.

Need to write test coverage for the `WorldTransformInvalidated` event, as the native view integrations consume this.

Source/Fuse.Nodes/Visual.WorldTransformInvalidated.uno
```diff
@@ -0,0 +1,68 @@
...
+		The events are rarely used, and the optimizations are very important.
+	*/
+
+	public partial class Visual
+	{
+		static List<Visual> _wtiSubjects = new List<Visual>();
+
+		static void BeginWatchWTI()
+		{
+			UpdateManager.AddAction(ProcessWTIHandlers);
```
mortoray: I'd rather see this done in the App class, or in the Rootviewport rooting, than as a static. Statics and testing don't play well together.

Duckers:
I'd prefer if we can fix the existing event and look into whether it can be eliminated later.

Its not only used by native views - also by animators that are transform-relative.

nekronos:
```xml
<GraphcisView>
    ...
    <ScrollView>
        <NativeViewHost />
    </ScrollView>
    ...
</GraphcisView>
```
In the above case the native view backing `NativeViewHost` is rooted on the `GraphicsView`, meaning its transform is `GraphicsView.WorldTransformInverse * NativeViewHost.WorldTransform`

kusma:
> Failed:  Fuse.Test.TransformModesTest.OffsetModeChange

Duckers:
I've reimplemented it now - should be passing the tests.

@mortoray please approve or specify better what changes you want.

Source/Fuse.Nodes/Visual.WorldTransformInvalidated.uno
```diff
@@ -0,0 +1,85 @@
...
+		void WTIRooted()
+		{
+			if (_worldTransformInvalidated != null)
+				IncrementWTIListener();
+		}
+
+		void WTIUnrooted()
+		{
+			if (_worldTransformInvalidated != null)
+				DecrementWTIListener();
```
mortoray: I think we should have a sanity test here that our own count is now 0 (emit InternalError if something fishy). We may also wish to force the count to 0, here, or during the rooting.  We wouldn't one one errant root/unroot combination from messing this all up (and we know they happen at times).

Source/Fuse.Nodes/Visual.Transform.uno
```diff
@@ -66,31 +66,21 @@ namespace Fuse
...
-			add { AddEventHandler(_worldTransformInvalidatedHandle, VisualBits.WorldTransformInvalidated, value); }
-			remove { RemoveEventHandler(_worldTransformInvalidatedHandle, VisualBits.WorldTransformInvalidated, value); }
-		}
-		
-		protected virtual void OnInvalidateWorldTransform() 
-		{ 
-			RaiseEvent(_worldTransformInvalidatedHandle, VisualBits.WorldTransformInvalidated);
+			_worldTransformVersion++;
+
+			// Raises WorldTransformInvalidated for subtree
```
mortoray: I think the comment is wrong, you raise the event for `this` and the subtree.

Duckers:
Still a regression with native views, investigating

Duckers:
I've fixed the implementation and it now works as expected on Android with native views.

Shuold be ready to approve and merge now @mortoray @kusma

Source/Fuse.Nodes/Visual.WorldTransformInvalidated.uno
Duckers: Fixed, see latest changes

Duckers:
There we go, I think it's good now!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Optimizaitons
 - Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
+- Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).
+- Backwards incompatible optimization change: The `protected virtual void Visual.OnInvalidateWorldTransform()` method was removed. The contract of this method was very expensive to implement as it had to be called on all nodes, just in case it was overridden somewhere. If you have custom Uno code relying on this method (unlikely), then please rewrite to explicitly subscribe to the `Visual.WorldTransformInvalidated` event instead, like so: Override `OnRooted` and do `WorldTransformInvalidated += OnInvalidateWorldTransform;`, Override `OnUnrooted` and to `WorldTransformInvalidated -= OnInvalidateWorldTransform;`, then rewrite `protected override void OnInvalidateWorldTransform()` to `void OnInvalidateWorldTransform(object sender, EventArgs args)`
+
+
```
kusma: Don't add thse extra newlines

PR #151, opened 2017-06-30
Misc cleanups

kusma:
These are just a few random minor changes, intended to make things a bit easier to maintain. Hopefully nothing controversial.


kusma:
@mortoray: Good point, removed that PNG.

We already have targeted tests for shape and brush, and the SdfImage-implementation was actually somewhat buggy (it interpreted the stroke-width wrong, and drew it twice as wide, hence the corresponding change to make it look the same).

PR #152, opened 2017-06-30
Bugfix iOS multitouch

nekronos:
For handling input in native on iOS we used [addTarget:action:forEvent](https://developer.apple.com/documentation/uikit/uicontrol/1618259-addtarget?language=objc) to listen for all touch events. During multitouch it looks like there is unexpected semantics or a bug here since all touch events stops when a finger is lifted.

This PR introduces a new way to handle touch events where the case above is not present. It also add the option to opt-out of automatic handling of touch events. Allowing the user to deal with the native-fuselibs interop if needed.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [x] Tests


nekronos:
Added a test page @mortoray

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
```diff
@@ -8,131 +8,197 @@ namespace Fuse.Controls.Native.iOS
...
 			var rootViewport = rootVisual as NativeRootViewport;
 			if (rootViewport != null)
 				rootView = rootViewport.RootView.NativeHandle;
+			return new RootInfo(rootVisual, rootView);
+		}
 
-			for (var i = 0; i < touches.Length; i++)
-			{
-				var touch = touches[i];
+		const int PointerCount = 10;
```
kusma: `MaxPointerCount`?

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
```diff
@@ -8,131 +8,197 @@ namespace Fuse.Controls.Native.iOS
...
+			for (var i = PointerCount; i-- > 0;)
+			{
+				var pointer = _activePointers[i];
+				if (CompareUITouch(pointer, null))
+					availableIndex = i;
+				else if (CompareUITouch(pointer, uiTouch))
+					return i;
 			}
+			_activePointers[availableIndex] = uiTouch;
+			return availableIndex;
```
kusma: Why does this scan backwards? If it scanned forwards, and saved the first available index instead of continiously overwriting, wouldn't the loop terminate earlier in the common-case of few pointers down? I also think the logic would be a bit more straight forward...

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
```diff
@@ -8,131 +8,197 @@ namespace Fuse.Controls.Native.iOS
...
+
+		[Foreign(Language.ObjC)]
+		static bool CompareUITouch(ObjC.Object x, ObjC.Object y)
+		@{
+			UITouch* touch1 = (UITouch*)x;
+			UITouch* touch2 = (UITouch*)y;
+			return x == y;
+		@}
+
+		static PointerEventData NewPointerEventData(ObjC.Object uiTouch, ObjC.Object rootView)
```
kusma: What's the point in renaming this method? The old name seems better to me...

Source/Fuse.Controls.Native/iOS/View.uno
```diff
@@ -11,11 +12,66 @@ namespace Fuse.Controls.Native.iOS
...
+		{
+			if (ErrorCheck(touches, filePath, lineNumber))
+				InputDispatch.OnTouchesEnded(origin, touches);
+		}
+
+		protected void DispatchTouchesCancelled(Visual origin, ObjC.Object touches, [CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0)
+		{
+			if (ErrorCheck(touches, filePath, lineNumber))
+				InputDispatch.OnTouchesCancelled(origin, touches);
+		}
```
kusma: I can't find any call-sites for any of these `DispatchTouches*`-methods... What's up, how can any of this stuff even work?

Source/Fuse.Controls.Native/iOS/UIViewInputDispatch.mm
```diff
@@ -0,0 +1,109 @@
...
+
+void removeInputHandler(UIView* view)
+{
+	objc_setAssociatedObject(view, InputHandlerKey, NULL, OBJC_ASSOCIATION_COPY_NONATOMIC);
+	objc_setAssociatedObject(view, VisualKey, NULL, OBJC_ASSOCIATION_COPY_NONATOMIC);
+}
+
+const char* SubclassPrefix = "__FUSE_INPUTHANDLER__";
+
+void installMethods(UIView* view)
```
yupferris: Indentation throughout this method seems a bit dubious

Source/Fuse.Controls.Native/iOS/UIViewInputDispatch.mm
```diff
@@ -0,0 +1,109 @@
...
+{
+	Class baseClass = [view class];
+
+	NSString* baseClassName = NSStringFromClass(baseClass);
+	NSString* subclassPrefix = [NSString stringWithUTF8String: SubclassPrefix];
+
+	if ([baseClassName hasPrefix:subclassPrefix])
+        return;
```
kusma: Yeah, as @yupferris points out, this is indented with spaces.

Source/Fuse.Controls.Native/iOS/UIViewInputDispatch.mm
```diff
@@ -0,0 +1,109 @@
...
+		{
+			NSLog(@"Failed to allocate subclass - %@", subClassName);
+			return;
+		}
+
+		class_addMethod(subClass, @selector(touchesBegan:withEvent:), (IMP)touchesBegan, "v@:@@");
+        class_addMethod(subClass, @selector(touchesMoved:withEvent:), (IMP)touchesMoved, "v@:@@");
+        class_addMethod(subClass, @selector(touchesEnded:withEvent:), (IMP)touchesEnded, "v@:@@");
+        class_addMethod(subClass, @selector(touchesCancelled:withEvent:), (IMP)touchesCancelled, "v@:@@");
+        objc_registerClassPair(subClass);
```
kusma: ...and these are also indented with spaces.

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
nekronos: To track the PointerIndex I need to keep a list of references to the UITouch event. The same UITouch object is reused by iOS as long as a touch is active. So consider this case:
1. Two touches occupy index 0 and 1 in the list.
2. The touch at index 0 ends, leaving only index 1 occupied.
3. The active touch raises a moved event.
    - Searching forwards will hit a `null` at index 0, but the touch is already active at index 1
    - Searching backwards and keeping track of the last available index fixes this case

Source/Fuse.Controls.Native/iOS/View.uno
nekronos: Thats the methods you need to call if you want to handle input yourself

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
```diff
@@ -8,131 +8,197 @@ namespace Fuse.Controls.Native.iOS
...
+			var handler = _handlers[(int)type];
+			var deactivate = ((int)type) > ((int)EventType.Moved);
+			for (var i = 0; i < count; i++)
+			{
+				var uiTouch = NSArrayObjectAtIndex(touches, i);
+				handler(rootInfo.RootVisual, NewPointerEventData(uiTouch, rootInfo.RootView));
+				if (deactivate)
+					DeactivateTouch(uiTouch);
+			}
 		}
```
nekronos: its to avoid lots of duplicated code in both C# and Objective-C. Also to avoid allocating 4 delegates for every native view.

nekronos: I can clean up the Uno code though

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
mortoray: I think this logic and the max `PointerCount` might fail. Cosnider this sequence of events:

- place fingers 1-5 in sequence
- lift fingers 1-4
- place fingers 1-4
- life fingers 1-3
- place fingers 1-3

This will require availableIndex of 12 yet only 5 fingers are involved.

What's wrong with just using the first availbleIndex, even if it's before another touch?

kusma: @nekronos: Sure, but that problem isn't solved by scanning backwards, it's solved by dealing with exact hits and inexact hits separately.

Consider this approach:
```uno
var firstUnused = -1;
for (var i = 0; i < PointerCount; ++i)
{
	var pointer = _activePointers[i];
	if (CompareUITouch(pointer, null) && firstUnused < 0)
		firstUnused = i;
	else if (CompareUITouch(pointer, uiTouch))
		return i;
}

_activePointers[firstUnused] = uiTouch;
return firstUnused;
```

AFAICT, this should produce the same result, without the backwards-scanning weirdness. And because it scans for exact hits forward, it should early out earlier in the case where uiTouch were already active, and there's only a few active pointers.

Source/Fuse.Controls.Native/iOS/View.uno
kusma: Ah, I see. I somehow expected that to appear in a7dae23eacb5d62cd24d3fa0628939a7c6509824, but this is analogous to that. Good stuff.

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
kusma: @mortoray: Were you looking at a different function than I was? I think the old code *did* use the first available index regardless...?

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
```diff
@@ -8,131 +8,206 @@ namespace Fuse.Controls.Native.iOS
...
+			{
+				var pointer = _activePointers[i];
+				if (CompareUITouch(pointer, null) && firstUnused < 0)
+					firstUnused = i;
+				else if (CompareUITouch(pointer, uiTouch))
+					return i;
 			}
 
-			if (touchEnded)
+			_activePointers[firstUnused] = uiTouch;
```
mortoray: Minor point, it's highly unlikely, but maybe we should throw an exception if `firstUnused` is still -1. I can't get stats, but it's not impossible to imagine an iPad supporting more than 10 for multiple hands.

Source/Fuse.Controls.Native/iOS/InputDispatch.uno
mortoray: I checked again and it does now, so I don't know which version I looked at.

nekronos:
Fixes https://github.com/fusetools/fuselibs/issues/4220

Issue #154, opened 2017-07-03
New WorldTransfrom invalidation broke NativeViewHost


nekronos:
```xml
<App>
    <NativeViewHost />
</App>
```
This app crashes at https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Nodes/Visual.WorldTransformInvalidated.uno#L77

https://github.com/fusetools/fuselibs-public/pull/149

yupferris:
cc @mortoray , who I believe suggested that particular change in the PR

kusma:
Those exceptions are super-unhelpful. It should at least be `Fuse.Diagnostics.InternalError()` instead, or at least have some text in it. Empty exceptions isn't going to help *anyone*.

mortoray:
Yeah, not great exceptions, but they did their job in picking up a problem in the new code.

this probably needs a "Regression" flag (we haven't migrated over issue tags yet it seems)

kusma:
Yeah, nobody moved the issue-labels from the old repo. I can do it a bit later if I remember ;)

PR #155, opened 2017-07-03
NativeViewHost: listen to WorldTransform after base.OnRooted

nekronos:
WorldTransformInvalidated can only be listened to after base.OnRooted is done

Fixes https://github.com/fusetools/fuselibs-public/issues/154

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Panels/NativeViewHost.uno
+++ b/Source/Fuse.Controls.Panels/NativeViewHost.uno
@@ -310,8 +310,6 @@ namespace Fuse.Controls
 		extern(Android || iOS)
 		protected override void OnRooted()
 		{
-			WorldTransformInvalidated += OnInvalidateWorldTransform;
-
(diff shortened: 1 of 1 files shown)


kusma:
@nekronos: What stops us from having a test triggering this? Is it because `NativeViewHost` is platform specific?

nekronos:
We should have seen this in ManualTestApp. The NativeViewHost hosting WebView/MapView should have crashed

kusma:
@nekronos: Sure, that prevents us from shipping breakages like this as a part of a new release-cycle. But it doesn't guard us against mid-cycle changes. I'm not saying it needs to be ideal, but it would be nice if we could somehow get *some* code up and running during normal CI...

Duckers:
This fix is better: https://github.com/fusetools/fuselibs-public/pull/158

PR #157, opened 2017-07-03
Android image rotation crash avoidance

Sunjammer:
This PR is intended to fix issues when attempting to rotate images according to EXIF. This primarily affects Samsung devices from experience, but results in hard crashes due to OOM. Since OOM exceptions can't be effectively caught on Android(!) this method attempts to predict the OOM by checking the image dimensions times 2 times 4 to get the total bytes that must be allocated for two copies, and then compares that to the available memory. If it cannot carry out the op, it passes the un-rotated images through and emits a warning. 

Setting `android:largeHeap` in the app manifest should mitigate this issue but until https://github.com/fusetools/uno/pull/1198 is merged or similar that can't be expected.

This PR contains:
- [ ] Changelog

### DIFF
--- a/Source/Fuse.Camera/Android/AndroidCamera.uno
+++ b/Source/Fuse.Camera/Android/AndroidCamera.uno
@@ -96,7 +96,11 @@ namespace Fuse.Camera
 			{
 				case android.app.Activity.RESULT_OK:
 					Image p = (Image)photo;
-					p.correctOrientationFromExif();
+					try{
+						p.correctOrientationFromExif();
+					}catch(Exception e){
+						// Still allow the picture through: It's been created but possibly with wrong orientation
+					}
 					onComplete.run(p.getFilePath());
 					return;
 				case android.app.Activity.RESULT_CANCELED:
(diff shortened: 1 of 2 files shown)


Source/Fuse.Camera/Android/AndroidCamera.uno
```diff
@@ -96,7 +96,11 @@ namespace Fuse.Camera
 			{
 				case android.app.Activity.RESULT_OK:
 					Image p = (Image)photo;
-					p.correctOrientationFromExif();
+					try{
+						p.correctOrientationFromExif();
+					}catch(Exception e){
+						// Still allow the picture through: It's been created but possibly with wrong orientation
```
kusma: maybe debug-log a bit to let people know *something* is up, at least?

Sunjammer: I log a warning from the java end

mortoray:
Closing as abandoned. I'd assume we still have an open issue for rotation, or we've fixed it elsewhere.

PR #160, opened 2017-07-04
Fixed text documentation

petterroea:
The text documentation had some weird repetition and weirdly placed content. This fixes the issue.

For the record, `ux` and `remarks` was dropped as it made the content appear below the "Interface of Text" list.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


Source/Fuse.Controls.Primitives/Docs/Text/Examples.md
```diff
@@ -1,4 +1,4 @@
-# Examples
+# Examples
```
yupferris: Nit: extra whitespace here; less of a problem in docs than code due to git blame info being clobbered, but try to look out for that :)

petterroea: Will do

PR #161, opened 2017-07-04
(restored) Optimized invalidation strategy for transforms

Duckers:
Avoids subtree traversion on world transform invalidation. This improves performance generally when animating large subtrees (e.g. scrollviews).

This is a restoration of what was reverted in https://github.com/fusetools/fuselibs-public/commit/da640d44aa6633c51ea8578a09ae92a1404dc4f5

Fixes https://github.com/fusetools/fuselibs-public/issues/154

This PR contains:
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## Optimizations
 - Optimized hit testing calculations. Improves scrolling in large scroll views with deep trees inside, among other things.
 - Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
+- Optimized invalidation strategy for transforms, to avoid subtree traversion. This improves performance generally when animating large subtrees (e.g. scrollviews).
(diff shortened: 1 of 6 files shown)


Duckers:
Verified that ManualTestApp builds and runs now with this branch.

Duckers:
whops, still a crash in one of the tests.

Duckers:
There we go. Had to resign and give it a backing field to cover all the cases.

Should be good to go now. Verified that ManualTestApp works

kusma:
@Duckers: It kind of seems like `_wtiRooted` and `_wtiListeners` are strongly related; while `_wtiRooted` is `false`, it seems no listeners should be added (so `_wtiListeners` should be zero). How about dropping `_wtiRooted`, and just setting `_wtiListeners` negative? Wherever gets added / decremented, we should throw an exception if `_wtiListeners` is negative.

That should save some bytes on each Visual, and keep the semantics pretty clear IMO. We can use encapsulation to make this logic read well if needed.

Duckers:
@kusma While your point is valid I would call premature optimization on that. As this is green now and a very valuable optimization I would suggest we merge and file a ticket for the optimization for later.

kusma:
@Duckers: It was just a suggestion since you seemed a bit disappointed about having to add a field. I have no problem with it as-is.

PR #163, opened 2017-07-04
Fuse.Share: Fixed iPad crash bug

nekronos:
iOS design guidelines requires a point of reference for use as the popover spawn origin. Added an optional argument to the share api for passing a UX element for use as the spawn origin. If the user does not provide this, emit a UserError and the popover will spawn in a default location

Not to happy about this fix. Makes the share API clunky. I think the Share API should be UX based instead of JS

This PR contains:
- [x] Changelog
- [x] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - `Each`, using `TemplateSource`, will no longer respond to template changes after rooting. This was done to simplify the code, and to support alternate sources, and is a minor perf improvement. It's not likely to affect any code since it didn't work correctly, and there's no way in UX to modify templates after rooting.
 - A memory leak was fixed by changing `Instantiator.TemplateSource` to a WeakReference. Only if you assigned this property a tempoary value in Uno would this change impact your code.
 
+## Fuse.Share
+- Fixed issue where using Fuse.Share would crash on iPad. Users must provide a position for spawn origin for the share popover. Check the Fuse.Share docs for more details.
+- Made iOS implementation internal, this was never ment to be public in the first place
+
 ## Optimizations
 - Optimized hit testing calculations. Improves scrolling in large scroll views with deep trees inside, among other things.
 - Optimized redundant OpenGL rendertarget operations. Gives speedups on some platforms.
--- a/Source/Fuse.Share/ShareModule.uno
+++ b/Source/Fuse.Share/ShareModule.uno
@@ -2,6 +2,7 @@ using Uno;
 using Uno.UX;
 using Uno.Threading;
 using Fuse.Scripting;
+using Fuse.Elements;
 
 namespace Fuse.Share
 {
@@ -12,6 +13,8 @@ namespace Fuse.Share
 	Supports sharing of raw text, and files with associated [mimetype](http://www.iana.org/assignments/media-types/media-types.xhtml).
 
 	Uses Action Sheet on iOS and ACTION_SEND Intents on Android.
+
+	NB: on iPad, iOS design guidelines requires a point on screen as the origin for the share popover. You can do this by passing a reference to a UX element.
 
 	You need to add a reference to "Fuse.Share" in your project file to use this feature.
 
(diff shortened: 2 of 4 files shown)


Sunjammer:
We should have a broader discussion about which APIs should be UX or JS at some point, especially now that expressions are coming to the forefront

Source/Fuse.Share/ios/iOSShareImpl.uno
```diff
@@ -2,24 +2,63 @@ using Uno.Compiler.ExportTargetInterop;
 using Uno.Threading;
 namespace Fuse.Share
 {
-	public extern(iOS) class iOSShareImpl
+	internal extern(iOS) class iOSShareImpl
```
kusma: This should be mentioned in the changelog..

Source/Fuse.Share/ShareModule.uno
```diff
@@ -80,6 +104,22 @@ namespace Fuse.Share
...
+		bool TryGetPosition(object arg, out float2 position)
+		{
+			position = float2(0.0f);
+			var obj = arg as Fuse.Scripting.Object;
+			if (obj != null && obj.ContainsKey("external_object"))
+			{
+				var element = ((Fuse.Scripting.External)obj["external_object"]).Object as Element;
+				if (element != null)
+				{
+					position = element.ActualPosition;
```
mortoray: Doesn't this need to be translated into screen coordinates, or relative to the NativeView?

`ActualPosition` is not likely what we want here, either the center of the element or `ActualAnchor` would make more sense (as a default at least).

Source/Fuse.Share/ios/iOSShareImpl.uno
```diff
@@ -2,24 +2,63 @@ using Uno.Compiler.ExportTargetInterop;
...
 	{
-		[Foreign(Language.ObjC)]
 		public static void ShareText(string text, string description)
+		{
+			if (IsIPad())
+				Fuse.Diagnostics.UserWarning("iPad requires a position as the spawn origion of the Share popover", text + " - " + description);
```
mortoray: I don't think we need a warning. Just pick a reasonable default position. It's not a warning since the default is intended then.

nekronos: There is no reasonable default position. iOS design guidelines requires the popover to have a sensible origin. On iPad the popover looks like this: ipad

Source/Fuse.Share/ShareModule.uno
nekronos: Its easiest to just use absolute coordinates. I'll change to use the center of the element as the origin.

mortoray: I meant that `ActualPosition` is local coordinates, you need to somehow translate to the screen corodinates still (there are functions in fuselibs for this).

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@
 - `Each`, using `TemplateSource`, will no longer respond to template changes after rooting. This was done to simplify the code, and to support alternate sources, and is a minor perf improvement. It's not likely to affect any code since it didn't work correctly, and there's no way in UX to modify templates after rooting.
 - A memory leak was fixed by changing `Instantiator.TemplateSource` to a WeakReference. Only if you assigned this property a tempoary value in Uno would this change impact your code.
 
+## Fuse.Share
+- Fixed issue where using Fuse.Share would crash on iPad. Users must provide a position for spawn origin for the share popover. Check the Fuse.Share docs for more details.
+- Made iOS implementation internal
```
kusma: Maybe explain that this was never meant to be public API in the first place?

Source/Fuse.Share/ios/iOSShareImpl.uno
nekronos: https://github.com/fusetools/fuselibs-public/pull/163/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR14

PR #165, opened 2017-07-04
Faster & better Fuse.Profiling

Duckers:
Optimizations and improvements to make Fuse.Profiling more precise and usfeul

### DIFF
--- a/Source/Fuse.Common/Profiling.uno
+++ b/Source/Fuse.Common/Profiling.uno
@@ -245,7 +245,7 @@ namespace Fuse
 			Write(id);
 		}
 
-		static string[] _cachedStrings = new string[0xFF];
+		static string[] _cachedStrings = new string[0x100];
 
 		private static byte CacheString(string str)
 		{
--- a/Source/Fuse.Common/Profiling.uno
+++ b/Source/Fuse.Common/Profiling.uno
@@ -102,93 +102,6 @@ namespace Fuse
 
 	}
 
-	extern(FUSELIBS_PROFILING) class StringCache
-	{
-		class Cache
-		{
-			public string Value { get; private set; }
-			public double LastTimeUsed { get; private set; }
-			public int StringHash { get; private set; }
-			public double IdleTime
-			{
-				get { return Uno.Diagnostics.Clock.GetSeconds() - LastTimeUsed; }
-			}
-
-			public Cache(string s)
-			{
-				Value = s;
-				StringHash = s.GetHashCode();
-				UpdateLastTimeUsed();
-			}
-			public override int GetHashCode()
-			{
-				return Value.GetHashCode();
-			}
-			public void UpdateLastTimeUsed()
-			{
-				LastTimeUsed = Uno.Diagnostics.Clock.GetSeconds();
-			}
-		}
-
-		readonly Cache[] _cache = new Cache[0xff];
-
-		public bool IsStringCached(string s)
-		{
-			var hash = s.GetHashCode();
-			for (byte b = (byte)(_cache.Length - 1); b != 0x00; b--)
-			{
-				var c = _cache[b];
-				if (c != null && c.StringHash == hash)
-					return true;
-			}
-			return false;
-		}
-
-		public byte CacheString(string s)
-		{
-			var stringHash = s.GetHashCode();
-			short leastUsedIndex = -1;
-			short availableIndex = -1;
-			var maxIdle = 0.0;
-			for (byte b = 0; b < _cache.Length; b++)
-			{
-				var c = _cache[b];
-				if (c == null)
-				{
-					availableIndex = b;
-				}
-				else if (c.StringHash == stringHash)
-				{
-					c.UpdateLastTimeUsed();
-					return b;
-				}
-				else
-				{
-					var idleTime = c.IdleTime;
-					if (maxIdle < idleTime)
-					{
-						maxIdle = idleTime;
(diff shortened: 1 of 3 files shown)


Source/Fuse.Common/Profiling.uno
```diff
@@ -334,21 +245,20 @@ namespace Fuse
 			Write(id);
 		}
 
+		static string[] _cachedStrings = new string[0xFF];
```
kusma: This should probably be `0x100` instead.

Source/Fuse.Common/Profiling.uno
```diff
@@ -334,21 +245,20 @@ namespace Fuse
...
-			Write(Command.CacheString);
-			Write(id);
-			Write(str);
+			var hash = (byte)(str.GetHashCode() & 0xFF);
+			if (_cachedStrings[hash] != str)
+			{
+				Write(Command.CacheString);
+				Write(hash);
+				Write(str);
+				_cachedStrings[hash] = str;
```
kusma: This will fail in the case where `str.GetHashCode() & 0xFF == 0xFF`, since _cachedStrings is 255 elements big, as opposed to 256.

Duckers: Nice catch :) fixed now

Source/Fuse.Elements/Element.Drawing.uno
```diff
@@ -66,21 +66,17 @@ namespace Fuse.Elements
...
 			}
 
-			var visibleRect = GetVisibleViewportInvertPixelRect(dc, RenderBoundsWithEffects);
-			if (visibleRect.Size.X == 0 || visibleRect.Size.Y == 0)
-			{
-				if defined(FUSELIBS_PROFILING)
-					Fuse.Profiling.EndRegion(Uno.Diagnostics.Clock.GetSeconds() - t);
-				return;
-			}
-
```
kusma: This chance seems a bit dubious to me...  reducing profiling overhead have the flip-side of us missing the time spent inside `GetVisibleViewportInvertPixelRect` from our profiling...

Duckers: Not missing it, it will be added to the parent's time block.

Duckers: Whether viewport culling is something the parent or child does is a philosophical question.

kusma: Yeah, exactly. It's being accounted wrong.

Duckers: Having thousands of nodes in the profiling report that take insignificant amount of time is less useful than just compounding them to the parent's time block, imho.

kusma: Reporting thousands of nodes culling-time on their parent seems like a much worse option to me.

kusma: And more importantly, optimizing the profiler at the cost of it's accuracy seems like a bad trade-off.

Duckers: This is improving its accuracy by reducing the overhead of the profiler.

kusma: It's decreasing the accuracy by accounting information on the wrong node. The data you get our is garbage at this point. It doesn't matter if collecting it was fast.

Duckers: I disagree. This improves the usefulness of the profiler. As I said, whether viewport culling is an activity of the parent or child is a philosophical question.

Asking for some pragmatism here. I'm not going to revert this.

kusma: If you consider doing something like reporting the cull-time as an event if it took a lot of time, you might get the best of both worlds, by the way. As is, this is IMO an move to the worse. Which is sad, because it's holding back otherwise very nice changes. How about you split this part out into a separate PR?

Duckers: The old implementation makes the profiler borderline useless in the very common use case of having a long list of nodes where most of the nodes are outside the viewport. The profiler has a not insignificant overhead per node of data it needs to report back home.

kusma: I think you have no idea how expensive reading `RenderBoundsWithEffects` can be (worst case is full subtree traversal, ). This is very useful information, and you don't want to get that reported on the wrong element. That makes your data tell you the wrong story.

I already suggested an alternative approach that avoids the bulk of the profiling-cost and allows you to catch the cases where this ends up taking a significant amount of time.

Another alternative is to add profiling inside `CalcRenderBoundsWithEffects()`, which will only trigger when the cache invalid.

Duckers:
@kusma : added reporting of time spent in culling as separate event on parent

kusma:
@Duckers: That's a great compromise, thanks!

Source/Fuse.Elements/Element.Drawing.uno
```diff
@@ -58,6 +58,9 @@ namespace Fuse.Elements
 			return fb;
 		}
 
+		extern (FUSELIBS_PROFILING) double _childCullTime;
+		extern (FUSELIBS_PROFILING) double cullTime; // to avoid local variable unused warning
```
kusma: You don't need this. You can do `extern double cullTime;` instead.

PR #166, opened 2017-07-04
Optimization: Repairs the broken FastTrackDrawWithOpacity feature

Duckers:
When LayoutControl and Panel were split back in the days, it seems the FastTrackDrawWithOpacity was forgotten. The result has been that e.g. images don't use the fast track when they have nontrivial opacity, resulting in a redundant framebuffer allocation and render to texture.

This fixes it.

This PR contains:
- [x] Changelog


Duckers:
PR from wrong repo

PR #167, opened 2017-07-04
Feature repair fast track draw with opacity

Duckers:
When LayoutControl and Panel were split back in the days, it seems the FastTrackDrawWithOpacity was forgotten. The result has been that e.g. images don't use the fast track when they have nontrivial opacity, resulting in a redundant framebuffer allocation and render to texture.

This fixes it.

This PR contains:
- [x] Changelog

### DIFF
--- a/Source/Fuse.Controls.Panels/LayoutControl.uno
+++ b/Source/Fuse.Controls.Panels/LayoutControl.uno
@@ -197,10 +197,12 @@ namespace Fuse.Controls
 		protected override bool FastTrackDrawWithOpacity(DrawContext dc)
 		{
 			if (HasChildren) return false;
-			if (Background == null) return true;
 			
-			DrawBackground(dc, Opacity);
-			return base.FastTrackDrawWithOpacity(dc);
+			if (Background != null)
+				DrawBackground(dc, Opacity);
+			
+			// Asserting base class doesn't need to draw anything!
+			return true;
 		}
 	}
 }
--- a/Source/Fuse.Controls.Panels/LayoutControl.uno
+++ b/Source/Fuse.Controls.Panels/LayoutControl.uno
@@ -193,5 +193,14 @@ namespace Fuse.Controls
 				return LayoutDependent.Maybe;
 			}
 		}
+
+		protected override bool FastTrackDrawWithOpacity(DrawContext dc)
+		{
+			if (HasChildren) return false;
+			if (Background == null) return true;
+			
+			DrawBackground(dc, Opacity);
+			return base.FastTrackDrawWithOpacity(dc);
+		}
 	}
 }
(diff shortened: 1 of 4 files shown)


Source/Fuse.Controls.Panels/LayoutControl.uno
```diff
@@ -193,5 +193,14 @@ namespace Fuse.Controls
...
 			}
 		}
+
+		protected override bool FastTrackDrawWithOpacity(DrawContext dc)
+		{
+			if (HasChildren) return false;
+			if (Background == null) return true;
+			
+			DrawBackground(dc, Opacity);
+			return base.FastTrackDrawWithOpacity(dc);
```
mortoray: It seems odd to return the call to `base` here yet the `(BAckground == null)` check returns `true` without checking the call to base.

Though I'm not sure this can be done correctly without a query function:

```
if (HasChildren) return false;
var bRes = base.CanFastTrackDrawWithOpacity(dc);
if (!bRes) return false;

if (Background != null) DrawBackground(dc,Opacity);
return base.FastTrackDrawWithOpacity(dc);
```

Source/Fuse.Controls.Panels/LayoutControl.uno
```diff
@@ -193,5 +193,16 @@ namespace Fuse.Controls
...
 		}
+
+		protected override bool FastTrackDrawWithOpacity(DrawContext dc)
+		{
+			if (HasChildren) return false;
+			
+			if (Background != null)
+				DrawBackground(dc, Opacity);
+			
+			// Asserting base class doesn't need to draw anything!
```
kusma: "Asserting"?

mortoray: The wording's a bit weird, but it means it's intentionally not calling the base class since shouldn't have anything to do.

kusma: Isn't it just a typo for "Assuming"?

But, wouldn't it be better to do something like:
```uno
protected override bool FastTrackDrawWithOpacity(DrawContext dc)
{
	if (HasChildren)
		return false;

	return base.FastTrackDrawWithOpacity(dc);
}
```

I mean, LayoutControl inherits from Control, which already does the above...

Duckers:
I don't see that Control does what you say.

Source/Fuse.Controls.Panels/LayoutControl.uno
nekronos: I think `base` would return `false`

https://github.com/Duckers/fuselibs-public/blob/79441786ac9f73bf5b82960b10b0706409390d25/Source/Fuse.Elements/Element.Effects.uno#L131

kusma:
I could have sworn I looked at a version where Control did this, but @Duckers is right. I must have been hallucinating ;)

PR #168, opened 2017-07-05
fixes https://github.com/fusetools/fuselibs/issues/4246

mortoray:
fixes https://github.com/fusetools/fuselibs/issues/4246

This PR contains:
- [x] Changelog

### DIFF
--- a/Source/Fuse.Navigation/RoutePageProxy.uno
+++ b/Source/Fuse.Navigation/RoutePageProxy.uno
@@ -213,7 +213,13 @@ namespace Fuse.Navigation
 			float p = 1; //defensive value for odd state
 			for (int i=0; i < _levels.Count; ++i)
 			{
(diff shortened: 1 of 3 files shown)


Source/Fuse.Navigation/RoutePageProxy.uno
```diff
@@ -213,7 +213,13 @@ namespace Fuse.Navigation
 			float p = 1; //defensive value for odd state
 			for (int i=0; i < _levels.Count; ++i)
 			{
-				var lp = _levels[i].PageProxy.Navigation.GetPageState(_levels[i].PageProxy.Page).Progress;
+				var level = _levels[i];
```
kusma: Since you only ever touch `level.PageProxy`, maybe store that in a local variable instead?

PR #177, opened 2017-07-06
Remove unnecessary render bounds inflation in FuseTextRenderer

### LINKED ISSUE #176: Bloated RenderBounds with Harfbuzz
It appears that `RenderBounds` for text is much larger than needed, resulting in wasted fillrate when caching:

image

Quote @yupferris :
It appears when using harfbuzz our text rendering measures very conservatively. I've tinted the cache rects to show this (to make sure it wasn't my system borking up):

image
image

on Android the rects are much tighter except when using -Duse_harfbuzz, where it looks the same as .net.

yupferris:
Currently, `FuseTextRenderer` is [inflating its renderbounds significantly](https://github.com/fusetools/fuselibs-public/blob/fddc657cde5c877b0add26283885358a57623e68/Source/Fuse.Controls.Primitives/TextControls/FuseTextRenderer/FuseTextRenderer.uno#L216). From what I can tell there's no good reason for this; the text measuring is already conservative wrt glyph sizes and appears to be producing very sensible bounds already. My guess is this was left over from some testing when the renderer was first introduced just to be safe, but now it just seems excessive.

Fixes #176.

This PR contains:
- [x] Changelog


yupferris:
before:

image

after:

image

yupferris:
Scratch that, this actually breaks some really crazy font cases (Zapfino in particular), so need to look further into it:

image

PR #178, opened 2017-07-06
Mode="Full|Partial" support for WhileVisibleInScrollView

eksperts:
As 
 [requested on community](https://www.fusetools.com/community/forums/howto_discussions/trigger_for_when_whole_object_is_visible), it would be useful if `WhileVisibleInScrollView` had an option to activate when the whole element is inside the visible area, as opposed to the default behaviour when the trigger is active when at least a part of the element is inside visible area.

This is a stab at implementing exactly that, by adding a `How="Partial|Full"` property on the trigger, where `Partial` is the default and behaves exactly as the trigger did before. Changing `How` to `Full` is an opt-in choice.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


Source/Fuse.Controls.ScrollView/ScrollView.Layout.uno
```diff
@@ -230,5 +250,19 @@ namespace Fuse.Controls
 			return 0;
 		}
 		
+		float DistanceFromViewStartLinear(float min, float max, float sp, float size)
```
mortoray: I'd just fold these functions into the `DistanceFromViewStart/End` functions.

Source/Fuse.Controls.ScrollView/ScrollView.Layout.uno
```diff
@@ -221,6 +221,26 @@ namespace Fuse.Controls
...
+		{
+			return float2(
+				DistanceFromViewStartLinear(min.X, max.X, ScrollPosition.X, ActualSize.X),
+				DistanceFromViewStartLinear(min.Y, max.Y, ScrollPosition.Y, ActualSize.Y));
+		}
+		
+		/**
+			The distance from the visible view area end for the provided rectangle. It will be zero if any part of it resides after the view area end.
+		*/
+		internal float2 DistanceFromViewEnd(float2 min, float2 max)
```
mortoray: This pattern of functions seems to be proliferating. Since only WhileVisible... uses them it might make more sense to have just one generic measuring function and let WhileVisible... impement the logic parts.

Source/Fuse.Controls.ScrollView/Tests/UX/WhileVisibleInScrollView.ux
```diff
@@ -9,6 +9,12 @@
 		<WhileVisibleInScrollView ux:Name="V3" Distance="1" RelativeTo="ScrollViewSize">
 			<Nothing/>
 		</WhileVisibleInScrollView>
+		<WhileVisibleInScrollView ux:Name="V4" How="Full">
```
mortoray: I'd prefer if this were split off into a separate test. Different features are good to separate when possible.

Source/Fuse.Controls.ScrollView/Triggers/WhileVisibleInScrollView.uno
```diff
@@ -120,16 +146,29 @@ namespace Fuse.Triggers
...
-			SetActive( dist < (maxDist + float.ZeroTolerance) );
+
+			bool isInView = false;
+
+			switch (_how)
+			{
+				case WhileVisibleInScrollViewHow.Full:
+					var distStart = _scrollable.ToScalarPosition(_scrollable.DistanceFromViewStart(min, max));
+					var distEnd = _scrollable.ToScalarPosition(_scrollable.DistanceFromViewEnd(min, max));
+					isInView = (distStart > (maxDist + float.ZeroTolerance)) && (distEnd > (maxDist + float.ZeroTolerance));
```
mortoray: This should probably be `- float.ZeroTolerance`. The comparison is the oppposite as the one below, and the tolerance is supposed to be in the diretion of inclusion, not exclusion.

Source/Fuse.Controls.ScrollView/Triggers/WhileVisibleInScrollView.uno
```diff
@@ -6,6 +6,15 @@ using Fuse.Controls;
 namespace Fuse.Triggers
 {
 	/**
+		How the bounds of an element are treated.
+	*/
+	public enum WhileVisibleInScrollViewHow
+	{
+		Partial,
```
mortoray: Describe each enum value here with /** docs */. It need not be long but it's helpful for the API docs.

Source/Fuse.Controls.ScrollView/ScrollView.Layout.uno
```diff
@@ -230,5 +230,46 @@ namespace Fuse.Controls
...
+			Start,
+			/**
+				When used in `DistanceFromView` method, measures the distance from the view end.
+			*/
+			End
+		}
+		/**
+			The distance from the visible view area start or end for the provided rectangle. It will be zero if any part of it resides before or after the view area start, respectively.
+		*/
+		internal float2 DistanceFromView(float2 min, float2 max, DistanceFromViewTarget target)
```
mortoray: Hmm, something still doesn't feel right here. You only use one of `min` or `max`. It should probably be just one parameter called `position`. You'll find you cold probably simplify the code if you do this.

Can the `DistanceToViewLinear` also be merged into this function

Source/Fuse.Controls.ScrollView/Triggers/WhileVisibleInScrollView.uno
```diff
@@ -6,6 +6,21 @@ using Fuse.Controls;
...
+		How the bounds of an element are treated.
+	*/
+	public enum WhileVisibleInScrollViewHow
+	{
+		/**
+			Activates the `WhileVisibleInScrollView` trigger when at least some of the element bounds are in view.
+		*/
+		Partial,
+		/**
+			Activates the `WhileVisibleInScrollView` trigger when all of the element bounds are in view.
```
mortoray: Rephrase, "Activates while the element is completely visible within the ScrollView"

Source/Fuse.Controls.ScrollView/Triggers/WhileVisibleInScrollView.uno
```diff
@@ -6,6 +6,21 @@ using Fuse.Controls;
 namespace Fuse.Triggers
 {
 	/**
+		How the bounds of an element are treated.
+	*/
+	public enum WhileVisibleInScrollViewHow
+	{
+		/**
+			Activates the `WhileVisibleInScrollView` trigger when at least some of the element bounds are in view.
```
mortoray: Rephrase, "Activates while the element is at least partially within the ScrollView"

Source/Fuse.Controls.ScrollView/Triggers/WhileVisibleInScrollView.uno
```diff
@@ -111,7 +126,24 @@ namespace Fuse.Triggers
...
+
+		WhileVisibleInScrollViewHow _how = WhileVisibleInScrollViewHow.Partial;
+		/**
+			How the bounds of an element are treated.
+
+			Default is `Partial`. When set to `Full`, the whole element bounds need to be inside view for the `WhileVisibleInScrollView` trigger to activate.
+
+			Both options can be combined with `Distance` to adjust when the trigger activates.
+		*/
+		public WhileVisibleInScrollViewHow How
```
mortoray: I guess this should be `Mode` now?

Issue #103, opened 2017-06-12
StatusBarBackground and BottomBarBackground issues on Android Nougat


eksperts:
Reported on community: https://www.fusetools.com/community/forums/general/android_nougat_samsung_s8

Tested with Android 7 running on Samsung Galaxy S8. Use the [pull-to-reload example](https://www.fusetools.com/examples/pull-to-reload) for testing.

First issue: there is a double `StatusBarBackground`; the expected one, and another black one right below it:
![ib5csmpvxdwr-image-1497272087865](https://user-images.githubusercontent.com/7188878/27035638-bf681550-4f8a-11e7-86ed-fb6e0bf0b62a.jpeg)

Second issue: the `BottomBarBackgound` is weirdly transparent; if you switch to another application and then reopen the Fuse preview, the other app shines through the `BottomBarBackground`:
![sbry4oayfkmj-image-1497272112971](https://user-images.githubusercontent.com/7188878/27035645-c3b5d4d0-4f8a-11e7-9f88-2b62c363b51a.jpeg)

WanderAraujoFernandes:
In Galaxy S8 Samsung switched the screen 16: 9 to 18.5: 9. When the app does not officially support this screen aspect, it has two black bars, one at the top and one at the bottom of the screen.

There is a configuration in the device that makes it possible to make an automatic correction: Settings -> Display -> Full Screen Apps

In app code fixe have some alternatives, in link have some explanations:
 Https://android-developers.googleblog.com/2017/03/update-your-app-to-take-advantage-of.html?m=1

In Fuse this issue can be fixed with meta-data configuration in androidmanifest.xml, related in previous link.

To resolve I created a UXL file and put code:

`<Extensions Backend="CPlusPlus">
  <Require Condition="Android" AndroidManifest.ApplicationElement>
    <![CDATA[
        <meta-data android:name="android.max_aspect" android:value="2.1" />
    ]]>
  </Require>
</Extensions>`

stand1715:
Thanks @WanderAraujoFernandes. It works perfectly.

I simply created a uxl file with following codes:
```
<Package>
	<Extensions Backend="CPlusPlus"> 
		<Require Condition="Android" AndroidManifest.ApplicationElement> <![CDATA[ <meta-data android:name="android.max_aspect" android:value="2.1" /> ]]> 
		</Require> 
	</Extensions>
</Package>
```
and included it in my unoproj file. 
```
"Includes": [
    "*",
[...]

PR #180, opened 2017-07-07
Android system ui configuration

petterroea:
Introduces a new UX behavior called `SystemScreenConfig`, which allows you to control the visibility of the status bar and navigation bar. Also listens to changes made by outside factors, and automatically reverts

This PR contains:
- [ ] Changelog
- [x] Documentation
- [x] Tests


Source/Fuse.Android/SystemScreenConfig.uno
```diff
@@ -0,0 +1,170 @@
...
+							Status bar and navigation bar example
+						</Text>
+						<Button Clicked="{status}" Text="Status" Margin="20"/>
+						<Button Clicked="{all}" Text="All" Margin="20"/>
+						<Button Clicked="{none}" Text="None" Margin="20"/>
+					</StackPanel>
+				</ClientPanel>
+			</App>
+	*/
+	public class SystemScreenConfig : Behavior
```
mortoray: You should add `OnRooted` and `OnUnrooted` handlers. The actual changes should happen unless the node is rooted. The OnUnrooted must cleanup the Timer as well if it's still sticking around.

petterroea:
@mortoray how would show be independent when there isn't another platform with this behavior?

petterroea:
Changes and refactor is made, requesting review

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,132 @@
...
+		TODO write docs please
+	*/
+	public class SystemScreenConfigBase : Behavior
+	{
+		public enum Visibility
+	    {
+	        None,
+	        Minimal,
+	        Full,
+	    }
```
kusma: This is all white-space damanged. Please install an [EditorConfig](http://editorconfig.org/)-plugin for your text-editor, this will make white-space management much easier for you. This seems to be the case all over the place here, so please go through and clean up all of it.

petterroea:
@sebbert is currently checking if the docs are building properly, it's a bit experimental.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,132 @@
+using Uno;
+using Uno.UX;
+using Uno.Collections;
+using Fuse;
+
+namespace Fuse.Controls
+{
+	/**
+		TODO write docs please
```
mortoray: Yes, please do. :)

Source/Fuse.Common/Timer.uno
```diff
@@ -20,17 +20,23 @@ namespace Fuse
...
+		public void Stop()
 		{
-			_running = false;
-			UpdateManager.RemoveAction(Update);
+			if(_running) 
+			{
+				_running = false;
+				UpdateManager.RemoveAction(Update);
+			}
 		}
```
kusma: Rather than doing this, I suggest you make this IDisposable, and make the dispose-method call `Stop()`. That generalizes the interface a bit, down to only the things you need to know about.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,132 @@
...
+	public class SystemScreenConfigBase : Behavior
+	{
+		public enum Visibility
+	    {
+	        None,
+	        Minimal,
+	        Full,
+	    }
+
+	    private static List<SystemScreenConfigBase> rootedConfigs = new List<SystemScreenConfigBase>();
```
mortoray: You can use a `MiniList` here for efficiency. I assume 0 and 1 are the most common number of configs used.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,132 @@
...
+
+	    protected Timer _timer;
+		private bool _rooted;
+
+	    protected override void OnRooted()
+		{
+			base.OnRooted();
+
+			_rooted = true;
```
mortoray: This variable shoudl not be needed. Node's expose a `IsRootingStarted` and `IsRootingComplete`.

Source/Fuse.Common/Timer.uno
```diff
@@ -54,10 +60,11 @@ namespace Fuse
 			method can not be relied upon for very short durations. For high-performance timing,
 			consider spawning a new thread. 
 		*/
-		public static void Wait(double duration, Action callback)
+		public static Timer Wait(double duration, Action callback)
 		{
 			var t = new Timer(duration, callback);
 			t.Start();
+			return t;
```
kusma: If you make this interface return `IDisposable` instead, this should IMO be much better.

This change reveal that there's a design-problem in this class as-is, though: there isn't anything explicitly keeping the `Timer`-instance alive. Yeah, there will probably be an implicit reference held because of the UpdateManager-subscription, but that feels pretty fragile.

Source/Fuse.Common/Timer.uno
```diff
@@ -20,17 +20,23 @@ namespace Fuse
...
-			_startTime = Uno.Diagnostics.Clock.GetSeconds();
-			UpdateManager.AddAction(Update);
-			_running = true;
+			if(!_running)
+			{
+				_startTime = Uno.Diagnostics.Clock.GetSeconds();
+				UpdateManager.AddAction(Update);
+				_running = true;
+			}
 		}
```
kusma: Drop this, you're not calling `Start()`.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+### Fuse.Android.SystemScreenConfig
+ - Added configuration behavior, lets you control the visibility of the navigation bar and status bar on Android.
```
kusma: Public API-changes to `Fuse.Timer` must be mentioned here.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,132 @@
...
+
+			rootedConfigs.Remove(this);
+			_rooted = false;
+		}
+
+		protected double calculateResetTime() 
+		{
+			double lowestReset = _resetDelay;
+
+			foreach(SystemScreenConfigBase theBase in rootedConfigs)
```
mortoray: Don't use `foreach`. Our memory manager doesn't deal with it well. Though it's not called much here I'd like to just keep it out of fuselibs due to the memory problem.

Use a classic C loop.

mortoray:
@kusma We came to the conclusion that we have no way in UX now to make this some kind of singleton to prevent multiple instances.  One motivating use-class greatly limits what can be done:

	<Android>
		<SystemScreenConfig IsDim="true" Theme="Dark"/>
	</Android>
	<iOS>
		<SystemScreenConfig Theme="Light"/>
	</iOS>


But it should report an error if a second instance is rooted.

kusma:
@mortoray: I did not come to that conclusion, just that the approach you suggested wouldn't work. That's no excuse for making a broken API that we can't easily get rid of in the future.

A few options:
- Document that only one instance is allowed, and give run-time error if that requirement is broken. (Basically what you suggest above, @mortoray, but the code doesn't do it ATM)
- Inspect the rooted ones, and give an error if they are in an inconsistent state.
- Give the states a documented priority, and use the the currently higherst priority rooted state.
- Basically anything *except* from silently overwriting state.

mortoray:
@kusma Yes, by issue I warning I mean to imply that using more than one is strictly undefined behaviour. No need for priority or any kind of resolution. No need to check for consistency since there can only be one node.

kusma:
@mortoray: I kind of want this to be an error rather than a warning, so that *when* we can make this a singleton, we can expect apps to not try this kind of shenanigans ;)

mortoray:
@kusma I want it to be an error as well, that's why I said: "But it should report an error if a second instance is rooted." ;P

petterroea:
Whitespaces are removed, Timer is refactored, and small changes to SystemScreenConfig are applied.
Docs are verified to work by sebbert, i've seen it myself.

Should be all good. @mortoray @kusma (the reviewers)

mortoray:
@petterroea What about the changes I requested to the overall structure?

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,129 @@
...
+	public class SystemScreenConfigBase : Behavior
+	{
+		public enum Visibility
+		{
+			None,
+			Minimal,
+			Full,
+		}
+
+		private static MiniList<SystemScreenConfigBase> rootedConfigs = new MiniList<SystemScreenConfigBase>();
```
mortoray: `MiniList` is a `struct`, you don't need to `new `it.

mortoray: Actually, since you're only supposed to have one rooted Config you don't even need this list.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,129 @@
...
+
+			resetTimer();
+
+			rootedConfigs.Remove(this);
+		}
+
+		protected double calculateResetTime() 
+		{
+			double lowestReset = _resetDelay;
+			for(int i = 0; i < rootedConfigs.Count; i++)
```
mortoray: Allow only one rooted config and you don't need this.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,129 @@
...
+			set
+			{
+				_isDim = value;
+			} 
+		}
+
+		//private Theme _theme = Theme.Dark;
+		//public Theme Theme { get; set; }
+
+		private double _resetDelay = 5.0;
```
mortoray: Add a comment indicating why you chose 5 as the default.

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,129 @@
...
+			{
+				return _isDim;
+			} 
+			set
+			{
+				_isDim = value;
+			} 
+		}
+
+		//private Theme _theme = Theme.Dark;
```
mortoray: remove at least this line. Maybe comment on the next that Android doesn't support Theme.

petterroea: Long-term i can imagine setting this would set a sane status bar color instead. Will do that for now.

petterroea:
More changes by review of @mortoray

Source/Fuse.Common/Timer.uno
kusma: You still don't need this change.

Source/Fuse.Common/Timer.uno
kusma: ... Nor this.

Source/Fuse.Controls.Native/Android/SystemScreenConfig.uno
```diff
@@ -0,0 +1,175 @@
+using Uno;
+using Uno.UX;
+using Uno.Collections;
+using Fuse;
+using Fuse.Controls.Native;
+
+namespace Fuse.Controls
+{
+	/**
+		TODO write docs please
```
kusma: Fix the TODO

petterroea: TODO fixed

Source/Fuse.Controls.Native/SystemScreenConfigBase.uno
```diff
@@ -0,0 +1,123 @@
+using Uno;
+using Uno.UX;
+using Uno.Collections;
+using Fuse;
+using Fuse.Internal;
+
+namespace Fuse.Controls
+{
+	public class SystemScreenConfigBase : Behavior
```
kusma: Please drop this pointless base-class (inline everything into `SystemScreenConfig`). Unneeded base-classes can't be removed from the API without breaking the API, and we generally don't use base-classes to share implementation details.

Source/Fuse.Controls.Native/Android/SystemScreenConfig.uno
```diff
@@ -0,0 +1,175 @@
...
+	*/
+	public extern(Android) class SystemScreenConfig : SystemScreenConfigBase
```
kusma: Don't mark `SystemScreenConfig` as `extern(Android)`, instead create an internal helper-class that has `extern(Android)` instead.

This is because otherwise, the API actually changes based on the platform. We would rather just have the implementation change.

petterroea:
Re-refactored the code after discussion on slack

petterroea:
Here is a video showing the test in action:

https://petterroea.s-ul.eu/A5irVDL8

As you can see, there is some sizing glitches on the app, but this is apparently a fuselibs issue. @cbaggers mentioned this might require a rewrite of some fundamental Android stuff.

cbaggers:
it's just the old super&root view stuff again. This will probably work just fine after that.

related: this https://github.com/fusetools/fuselibs-public/issues/103 & 1 internal ticket

Source/Fuse.Common/Timer.uno
```diff
@@ -22,15 +22,21 @@ namespace Fuse
...
 		{
-			_startTime = Uno.Diagnostics.Clock.GetSeconds();
-			UpdateManager.AddAction(Update);
-			_running = true;
+			if(!_running)
+			{
+				_startTime = Uno.Diagnostics.Clock.GetSeconds();
+				UpdateManager.AddAction(Update);
+				_running = true;
+			}
```
kusma: You still shouldn't be needing this, please drop it,

Source/Fuse.Common/Timer.uno
```diff
@@ -22,15 +22,21 @@ namespace Fuse
...
 
 		void Stop()
 		{
-			_running = false;
-			UpdateManager.RemoveAction(Update);
+			if(_running) 
+			{
+				_running = false;
+				UpdateManager.RemoveAction(Update);
+			}
```
kusma: Nor this.

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
```diff
@@ -0,0 +1,104 @@
+using Fuse.Controls.Native;
+
+namespace Fuse.Controls.Native
+{
+	extern(Android) internal class SystemScreenConfigAndroidImpl
+	{
+		private static SystemUiVisibility.Flag _wantedFlags = SystemUiVisibility.Flag.None;
+		private static SystemScreenConfig config;
+
+		public static void init(SystemScreenConfig config)
```
kusma: methods have names that start with upper-case letters in fuselibs, so this should be "Init", not "init".

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
```diff
@@ -0,0 +1,104 @@
...
+		{
+			SystemScreenConfigAndroidImpl.config = config;
+			SystemUiVisibility.VisibilityChanged += visibilityChanged;
+		}
+
+		public static void deInit()
```
kusma: `deInit` -> `DeInit()`... And so on, please fix all methods here.

Source/Fuse.Controls.Native/Android/SystemUiVisibility.uno
```diff
@@ -0,0 +1,91 @@
...
+				VisibilityChanged((Flag)flags);
+		}
+
+		[Foreign(Language.Java)]
+		private static Flag getVisibilityFlags()
+		@{
+			View decorView = com.fuse.Activity.getRootActivity().getWindow().getDecorView();
+			return decorView.getSystemUiVisibility();
+		@}
+		[Foreign(Language.Java)]
```
kusma: please have a blank newline between these methods

Source/Fuse.Controls.Native/SystemScreenConfig.uno
```diff
@@ -0,0 +1,215 @@
...
+				_isDim = value;
+				if defined(Android)
+				{
+					SystemScreenConfigAndroidImpl.setDimState(value);
+				}
+			} 
+		}
+
+		//Sane default of 5 seconds
+		private double _resetDelay = 5.0;
```
kusma: Did you pick this default-value from somewhere?

Source/Fuse.Controls.Native/SystemScreenConfig.uno
```diff
@@ -0,0 +1,215 @@
...
+		/**
+			Sets the time before outside changes to visibility states is reset.
+			Setting the value to 0 disables the reset behavior.
```
kusma: I think you should document the default value here.

Source/Fuse.Common/Timer.uno
petterroea: Some kind of state machine is needed in order to stop it from calling Stop() twice(f.ex once by the callback calling dispose, once by itself calling stop). 
My suggested fix is to change Update() to call Dispose, and having a check in Dispose, returning if the timer isn't running

petterroea: I pushed a suggested solution to this

Source/Fuse.Common/Timer.uno
```diff
@@ -29,8 +29,8 @@ namespace Fuse
 
 		void Stop()
 		{
-			_running = false;
 			UpdateManager.RemoveAction(Update);
+			_running = false;
```
kusma: I don't think there's any point in swapping these around...

petterroea:
@kusma changes made. Highlight is a new suggestion to the timer API

Source/Fuse.Common/Timer.uno
```diff
@@ -41,23 +41,30 @@ namespace Fuse
...
+				if (_once) Dispose();
 				else _startTime = now;
 			}
 		}
 
+		public void Dispose()
+		{
+			if(!_running)
+				return;
+			Stop();
```
kusma: Could you do:
```uno
if (_running) 
	Stop();
```
instead? "If-not-running-return-else-stop" reads kind of worse, and is more awkward to extend in the future...

Source/Fuse.Controls.Native/Android/SystemUiVisibility.uno
```diff
@@ -0,0 +1,92 @@
...
+namespace Fuse.Controls.Native
+{
+	[ForeignInclude(Language.Java,
+		"android.graphics.Color",
+		"android.os.Build",
+		"android.view.View",
+		"android.view.WindowManager",
+		"android.view.Window",
+		"android.view.WindowManager.LayoutParams")]
+	//Requires android SDK version 16 or newer(Jelly Bean)
```
kusma: We [already require API level 16](https://www.fusetools.com/docs/basics/supported-platforms), so this comment isn't really necessary.

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
```diff
@@ -0,0 +1,104 @@
...
+				break;
+			}
+		}
+
+		public static void SetStatusState(bool val)
+		{
+			if(val)
+			{
+				_wantedFlags = _wantedFlags ^ SystemUiVisibility.Flag.Fullscreen;
+				SystemUiVisibility.Flags = SystemUiVisibility.Flags ^ SystemUiVisibility.Flag.Fullscreen;
```
kusma: xoring flags like that seems pretty fragile; it just flips the SystemUiVisibility.Flag.Fullscreen-bit regardless of what it used to be.

Wouldn't it be better to explicitly clear the flag instead, like so?
```uno
_wantedFlags &= ~SystemUiVisibility.Flag.Fullscreen;
SystemUiVisibility.Flags &= ~SystemUiVisibility.Flag.Fullscreen;
```

Also, it seems strange to track _wantedFlags separately from SystemUiVisibility.Flags... So maybe even this?
```uno
_wantedFlags &= ~SystemUiVisibility.Flag.Fullscreen;
SystemUiVisibility.Flags = _wantedFlags;
```

This seems like it could also get us rid of `getVisibilityFlags`, no?

Source/Fuse.Controls.Native/Android/SystemUiVisibility.uno
```diff
@@ -0,0 +1,92 @@
...
+		public delegate void VisibilityChangedHandler(Flag newFlag);
+
+		public static event VisibilityChangedHandler VisibilityChanged;
+
+		static SystemUiVisibility()
+		{
+			JavaInit();
+		}
+
+		public static Flag Flags {
```
kusma: newline before the curly brace, please

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
petterroea: nice xor catch, fixed it.

`_wantedFlags` represents the flag state the app wants. The actual value of `Flags` may change due to outside factors, such as the user swiping down from the top while the status bar is hidden.

This is corrected using the timer mechanic, as it is according to @nekronos what most apps do.

kusma: @petterroea: Sure, but this code is run when a property is changed, and I believe that should yield an immediate update to the wanted flags, no? In other words, I think the timer should only do it's work unless some other update happened in the meantime...

petterroea: Right. Committed.

petterroea: Not neccesarily, as the Flags might contain other flags this code doesn't deal with yet. I have written the code with the fact that not all the flags are being controlled by the code in mind.
Setting Flags to _wantedFlags directly would therefore step on potential other code controlling other flags that this code does not know of, so i feel this is safest as long as not all the flags are controlled from here.

Source/Fuse.Common/Timer.uno
petterroea: If you mean that there is no point in setting `_running` to false, i'd argue there is at it makes `_running` accurately represent the current state of the Timer. This might avoid unexpected behavior for the next person making changes relying on `_running` to Timer

kusma: No, `_running` was already being set to `false`, you just moved it below the `UpdateManager.RemoveAction(Update);`-line. That's just code-churn, and doesn't improve anything.

petterroea:
@kusma Allright, changes are made

Source/Fuse.Controls.Native/SystemScreenConfig.uno
kusma: You still haven't explained where this default-value comes from... What make 5 seconds "saner" than 1 or 10 seconds... or even no reset by default?

Source/Fuse.Controls.Native/SystemScreenConfig.uno
```diff
@@ -0,0 +1,241 @@
...
+					SystemScreenConfigAndroidImpl.SetDimState(value);
+				}
+			} 
+		}
+
+		//Sane default of 5 seconds
+		private double _resetDelay = 5.0;
+		/**
+			Sets the time before outside changes to visibility states is reset, defaults to 5.
+			Setting the value to 0 disables the reset behavior.
```
kusma: Why does 0 disable as opposed to immediately reset? Wouldn't -1.0 (e.g any negative value) be a better choice for disabling resetting, so people who want immediately reset don't have to do ugly things like 0.000001?

petterroea: Its mostly to protect against weird behavior. The status bar some times flickers if you try hard enough to force it visible, which is a bit ugly.

petterroea: The battery icon visibly flashes as well, instantly returning the status bar to hidden is ugly

kusma: Won't people get that strange behavior still if they use a tiny value?

petterroea: Correct, but i am assuming the general usecase will be entering seconds. If someone tries to use a tiny value, they will quickly find out for themselves that it is a bad idea.

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
```diff
@@ -0,0 +1,123 @@
...
+											SystemUiVisibility.Flag.LightStatusBar;
+			
+			SystemUiVisibility.Flag otherFlags = newFlag & (~mask);
+
+			SystemUiVisibility.Flag actualFlags = newFlag & mask;
+
+			//Reset the timer anyways, in case the state changed to what we want
+			config.resetTimer();
+			//Was things changed due to an outside influence?
+			if(actualFlags != _wantedFlags && config.ResetDelay > 0.0001f) //Cheeky
```
kusma: This will treat tiny values (like 0.00001) as zero, which is unexpected (and undocumented) behavior.

petterroea: Good catch, and i can change this, but wouldn't that make the disable mechanic vulnerable to floating point precision errors in animators(if someone was to, for some weird reason, want to animate the time it takes it to change).

It's not a case we should be interrested in supporting anyways, but it's a side effect nevertheless

kusma: If we tell the users to use negative values to disable instead, that problem goes away. And then we can have zero as a "glitchy, but otherwise working case". I think that'd be fine, people would understand that not every command is instant.

petterroea: Fair. Making change.

Source/Fuse.Controls.Native/SystemScreenConfig.uno
petterroea: Nothing. It is just a suggestion.

petterroea: As far as i know there is no standard as to how long this is supposed to take, i set 5 as it seemed, to me, as a sane default.

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
```diff
@@ -22,6 +22,7 @@ namespace Fuse.Controls.Native
 		public static void TimerDone()
 		{
 			SystemUiVisibility.Flags = _wantedFlags;
+			debug_log "Settingh state";
```
kusma: I don't think you meant to re-introduce this

Source/Fuse.Controls.Native/Android/SystemScreenConfigAndroidImpl.uno
```diff
@@ -114,7 +115,7 @@ namespace Fuse.Controls.Native
 			//Reset the timer anyways, in case the state changed to what we want
 			config.resetTimer();
 			//Was things changed due to an outside influence?
-			if(actualFlags != _wantedFlags && config.ResetDelay > 0.0001f) //Cheeky 
+			if(actualFlags != _wantedFlags && config.ResetDelay >= 0) //Cheeky
```
kusma: Is this "Cheeky" now? I don't see why...

Source/Fuse.Controls.Native/SystemScreenConfig.uno
petterroea: As to why no reset isn't the default, i'l argue that if the developer wants to explicitly set the state of the System UI, that probably means the developer wishes for the system ui to stay at the set state.

Tests/ManualTests/ManualTestingApp/Singles/SystemScreenConfigPage.ux
```diff
@@ -66,7 +66,7 @@
 			<Text>Press this button:</Text>
 			<StdButton Clicked="{none}" Text="Press me!" Margin="20,50,20,20"/>
 
-			<Text TextWrapping="Wrap">Swipe down from the top of the device to bring back the status bar manually. It should dissapear again in 5 seconds.</Text>
+			<Text TextWrapping="Wrap">Swipe down from the top of the device to bring back the status bar manually. It should dissapear again in 2 seconds.</Text>
```
kusma: Typo: "dissapear" should be "disappear".

Source/Fuse.Controls.Native/SystemScreenConfig.uno
```diff
@@ -0,0 +1,241 @@
...
+			Full,
+		}
+
+		public enum StatusBarTheme
+		{
+			Light,
+			Dark
+		}
+
+		private static SystemScreenConfig rootedConfig = null;
```
kusma: This should have an underscore-prefix.

Source/Fuse.Controls.Native/SystemScreenConfig.uno
```diff
@@ -0,0 +1,241 @@
...
+				SystemScreenConfigAndroidImpl.Init(this);
+			}
+		}
+		protected override void OnUnrooted()
+		{
+			base.OnUnrooted();
+
+			resetTimer();
+
+			rootedConfig = null;
```
kusma: We should probably only null out this if `rootedConfix == this`.

Source/Fuse.Controls.Native/SystemScreenConfig.uno
kusma: Yeah, totally fair. I just wondered if there were any particular reasoning behind this value. It seems fine to me.

kusma:
Yeah, so just two minor nits, and I think this is good to go!

petterroea:
@kusma 🎊 However, i want to hear @cbaggers ' opinion on this being merged now, given the underlying fuse Android bootstrap has some fundamental issues that causes some weird layout glitches on the app when the status bar is toggled

kusma:
@petterroea: Fair enough! Thanks for bearing with me, and I think the result looks really good!

petterroea:
@kusma I had a chat with @cbaggers , he is against merging as this would introduce a visual bug caused by errors in our Android root view code. 

This bug is a part of a bigger issue that, according to @cbaggers , needs to be fixed some time anyways

kusma:
@petterroea: OK. That sucks a bit, but I guess it is what it is...

petterroea:
If anything this means we should make an issue on, and start thinking of fixing this, but i have no idea if this would happen any time soon

kusma:
I have no idea what the underlying issue is. @cbaggers, is this something you'll follow up on? If not, care to elaborate?

cbaggers:
@kusma merging this would introduce a graphical bug as a statusbar size gap is left below the app root view.

The issue is due to our current root/super view setup in Fuse.Platform.SystemUI

I can't think of a way to fix this other than to change that bootstrapping to do 'the usual android thing' but this means we loose callback on softkeyboard resize etc. This is the correct way to go but I believe it needs to be a separate task.

kusma:
@cbaggers: Right, so if I understand you correctly, you're saying that `Fuse.Platform.SystemUI` expects to be the only one meddling with the system UI, or it'll get in a bad state? If so, I totally get that, and yeah, we'll probably have to fix this first. I guess an alternative could be to let `Fuse.Platform.SystemUI` in on the action in this PR instead of going around its back?

mortoray:
Closing as abandoned. REcorded desired feature in https://github.com/fusetools/fuselibs-public/issues/825

Issue #182, opened 2017-07-07
Lingering cursor bug is back


yupferris:
When focusing a `TextBox` and then unfocusing it, its cursor remains visible, although it stops blinking. Likely a bug with caching and the `TextBox` possibly not invalidating its visual properly when losing focus.

![image](https://user-images.githubusercontent.com/3166056/27957919-e94ca7e2-6320-11e7-9b7a-4ddb9815228e.png)

yupferris:
Minimal repro app:

```
<App>
	<StackPanel>
		<TextInput>
			<!-- This rect isn't necessary for repro but helps to make the TextInput visible -->
			<Rectangle Fill="#eee" Layer="Background" />
		</TextInput>
	</StackPanel>
</App>
```

kusma:
Which platform do you observe this on?

yupferris:
Windows in .net

yupferris:
Just tested with release-1.1, problem is there too.

yupferris:
Ok, it only affects `FuseTextEdit`, not `FallbackTextEdit`. Tested by commenting out the harfbuzz lines in the factories. Looking into the difference there.

yupferris:
It seems `InvalidateVisual` from `FuseTextEdit`'s `INotifyFocus.OnFocusLost` impl isn't causing redraw (its `DrawVisual` isn't called after that), which seems a bit odd..

yupferris:
Broken in 81eae3e4, reverting that should fix this.

yupferris:
Fixed in https://github.com/fusetools/fuselibs-public/pull/184

PR #183, opened 2017-07-07
Revert "Move IncreaseFrameIndex to prior to Update to make it consist…

yupferris:
…ent across platforms"

This reverts commit 81eae3e47960ee0624de073d5d702005b0de9e0e.

This commit causes inconsistent frame index for input events, causing
issue #182 and likely some other problems we haven't caught yet.


yupferris:
We need to figure out a more proper fix for this, but for now this will mitigate the regression from the bad commit.

yupferris:
Superceded by an upcoming @kusma PR :)

PR #184, opened 2017-07-07
make sure IncreaseFrameIndex happens right after draw

kusma:
Doing things like InvalidateVisual in response to input requires that
the UpdateManager.FrameIndex was increased right after drawing, because
otherwise the frame index will be the same when the input happens,
causing elements that were validated the same frame to not be
invalidated.

This problem was introduced in 81eae3e, where another similar issue
was worked around. This should fix both cases.

Fixes issue #182

Unfortunately, this doesn't seem to be possible to write a test-case for,
because the logic happens in AppBase, and we don't have an instance
of that during `uno test`.

### DIFF
--- a/Source/Fuse.Desktop/DesktopApp.uno
+++ b/Source/Fuse.Desktop/DesktopApp.uno
@@ -66,6 +66,8 @@ namespace Fuse
 
 			if defined(FUSELIBS_PROFILING)
 				Profiling.EndDraw();
+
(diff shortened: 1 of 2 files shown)


Duckers:
If this has been manually verified to fix #182, then this LGTM

yupferris:
Yeah #182 is def fixed; that was verified.

PR #185, opened 2017-07-07
Prevents JavaScript from crashing when dependencies have reserved ES5.1 names

Duckers:
This also fixes a regression introduced by https://github.com/fusetools/uno/pull/1215 where the root node `this` is now included in NameTable, which causes a crash in JavaScript when this is introduced as a variable to the script.

This PR contains:
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## JavaScript
+- Fixed a bug where if any injected dependencies to a JavaScript (e.g. `ux:Name`, `ux:Dependency`, `dep:`) have names that correspond to EcmaScript 5.1 reserved words, these dependencies will now be available with a dollar sign `$` prefix (e.g. `$undefined`) instead of causing a script compile error.
+
 ## Templates
 - Triggers may now use templates which will be instantiated and added to the parent when active (like a node child).
 	<WhileActive>
--- a/Source/Fuse.Reactive.JavaScript/JavaScript.ReservedWords.uno
+++ b/Source/Fuse.Reactive.JavaScript/JavaScript.ReservedWords.uno
@@ -1,0 +1,31 @@
+using Uno.Collections;
+
+namespace Fuse.Reactive
+{
+	partial class JavaScript
+	{
+		// Taken from the ECMAScript 5.1 Standard
+		// https://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205th%20edition%20December%202009.pdf
+		static HashSet<string> _reservedKeywords = new HashSet<string>()
+		{
+			"break", "do", "instanceof", "typeof",
+			"case", "else", "new", "var",
+			"catch", "finally", "return", "void",
+			"continue", "for", "switch", "while",
+			"debugger", "function", "this", "with",
+			"default", "if", "throw",
+			"delete", "in", "try", 
+			"undefined" // not actually a reserved word in the spec, but most linters etc treat it as such
+		};
+
+		internal static bool IsReservedKeyword(string s)
+		{
+			return _reservedKeywords.Contains(s);
+		}
+
+		internal static string ToValidName(string name)
+		{
+			return IsReservedKeyword(name) ? "$" + name : name;
+		}
+	}
+}
--- a/Source/Fuse.Reactive.JavaScript/ObservableProperty.uno
+++ b/Source/Fuse.Reactive.JavaScript/ObservableProperty.uno
@@ -8,7 +8,7 @@ namespace Fuse.Reactive
 	{
 		public LazyObservableProperty(ThreadWorker w, Scripting.Object obj, Property p): base(w, obj, p)
 		{
-			w.Context.ObjectDefineProperty(obj, p.Name.ToString(), Get);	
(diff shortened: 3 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## JavaScript
+- Fixed a bug where if any injected dependencies to a JavaScript (e.g. `ux:Name`, `ux:Dependency`, `dep:`) have names that correspond to EcmaScript 5.1 reserved words, these dependencies will now be available with a dollar sign `$` prefix (e.g. `$undefined`) instead of causing a script compile error.
```
kusma: Would it be possible to *always* make a dependency named `foo` available as `$foo` as well, so people don't need to manually keep track of the current list of reserved keywords?

Source/Fuse.Reactive.JavaScript/ObservableProperty.uno
```diff
@@ -8,7 +8,7 @@ namespace Fuse.Reactive
 	{
 		public LazyObservableProperty(ThreadWorker w, Scripting.Object obj, Property p): base(w, obj, p)
 		{
-			w.Context.ObjectDefineProperty(obj, p.Name.ToString(), Get);	
+			w.Context.ObjectDefineProperty(obj, JSName, Get);
```
kusma: For instance if we did this:
```uno
var name = p.Name.ToString();
if (!IsReservedKeyword(name))
	w.Context.ObjectDefineProperty(obj, name, Get);	
w.Context.ObjectDefineProperty(obj, "$" + name, Get);
```
?

CHANGELOG.md
Duckers: This would be a backwards incompatible change

Source/Fuse.Reactive.JavaScript/ObservableProperty.uno
Duckers: This would be backwards incompatible

CHANGELOG.md
kusma: In what way? Because we would inject more symbols from the data-context? If so, fusetools/Uno#1193 would also be a backwards incompatible change, no? In fact, adding *anything* would be a problem.

If that's really the case, maybe we should add an API to be able to resolve things like this manually rather than $-prefixing ourselves? Something like:

```JavaScript
var dep = require("FuseJS/Dependency");
var $class = dep("class");
```

...and only implicitly import non-keyword dependencies?

Duckers: If you are suggesting injecting `$foo` *in addition* to injecting `foo`, then it would be backwards compatible. However, I question the usefulness of this. JS programmers must be expected to know the language they are programming in.

Source/Fuse.Reactive.JavaScript/ObservableProperty.uno
Duckers: This particular case is incompatible because you now have two observables representing the same value (double source of truth)

Duckers: Or, hm, no, they just point to the same getter. I suppose this is fine. However, I am generally against it because this adds a lot of bloat, makes the syntax uglier and additional members only to allow programmers to not know their own programming language.

Duckers: There is no point adding these unless we expect people to use it and make it the best practice. This means rewriting a lot of docs and ending up with uglier-looking code.

CHANGELOG.md
kusma: Read my comment again. The motivation of this change is *just* as backwards incompatible as injecting $foo in addition to injecting foo.

The second part of your comment has nothing to do with what I'm talking about.

Source/Fuse.Reactive.JavaScript/JavaScript.ReservedWords.uno
```diff
@@ -0,0 +1,31 @@
+using Uno.Collections;
+
+namespace Fuse.Reactive
+{
+	partial class JavaScript
+	{
+		// Taken from the ECMAScript 5.1 Standard
+		// https://www.ecma-international.org/publications/files/ECMA-ST-ARCH/ECMA-262%205th%20edition%20December%202009.pdf
+		static HashSet<string> _reservedKeywords = new HashSet<string>()
```
nekronos: Missing some reserved keywords. Look at `7.6.1.2` in the PDF linked above

nekronos:
This feature needs to be reflected in our docs as well.

Source/Fuse.Reactive.JavaScript/Tests/UX/JavaScript.ReservedWord.ux
```diff
@@ -0,0 +1,14 @@
+<Panel ux:Class="UX.JavaScriptReservedWord">
+	<Panel ux:Class="UX.JSReserved">
+		<Router ux:Name="undefined" />
```
kusma: IsMasterRouter="false"

mortoray:
Closing as abandoned. Opened issue https://github.com/fusetools/fuselibs-public/issues/823 since this still seems relevant.

PR #186, opened 2017-07-07
WrapPanel ContentAlignment

bolav:
Trying to make https://upload.wikimedia.org/wikipedia/en/f/f2/Tastes_in_Foursquare_8.jpeg layout is impossible with todays `WrapPanel`. Changing `RowAlignment` to `ContentAlignment` and ordering the content in the major direction as well as in the minor direction.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


Source/Fuse.Controls.Panels/WrapPanel.uno
```diff
@@ -82,12 +82,21 @@ namespace Fuse.Controls
...
 		}
 
-		/** The alignment of the content within the wrap panel rows. */
+		/**
+			@deprecated 2017-07-07
+		*/
 		public Alignment RowAlignment
 		{
 			get { return _wrapLayout.RowAlignment; }
 			set { _wrapLayout.RowAlignment = value; }
```
kusma: Please use `Fuse.Diagnostics.Deprecated` here, to warn people using this.

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
```diff
@@ -66,18 +66,36 @@ namespace Fuse.Layouts
 		}
 
 		Alignment _rowAlignment = Alignment.Default;
+		/**
+			@deprecated 2017-07-07
+		*/
 		public Alignment RowAlignment
```
kusma: Please add a `[Obsolete("User ContentAlignment instead")]` here.

Source/Fuse.Controls.Panels/WrapPanel.uno
```diff
@@ -82,12 +82,21 @@ namespace Fuse.Controls
 			set { _wrapLayout.FlowDirection = value; }
 		}
 
-		/** The alignment of the content within the wrap panel rows. */
+		/**
+			@deprecated 2017-07-07
+		*/
 		public Alignment RowAlignment
```
kusma: Please add a `[Obsolete("User ContentAlignment instead")]` here.

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
```diff
@@ -135,6 +153,8 @@ namespace Fuse.Layouts
 			var placements = new float4[elements.Count];
 			//minorMaxSize in each major row, assinged per element
 			var minorSizes = new float[elements.Count];
+			// save the available space for each row
+			var majorRest = new Dictionary<float, float>();
```
mortoray: This should be an array like the other tracking avraible `placements` adn `minorSizes`... or wiat, I'm not clear what this does. But a `float` should not be used as  dictionary key. Precision problems will prevent this from working.

Source/Fuse.Controls.Panels/WrapPanel.uno
bolav: Is this not enough? 

https://github.com/bolav/fuselibs-public/blob/WrapPanel_ContentAlignement/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno#L77

kusma: Hmm, yeah. My bad. That should cover this case. Sorry for missing it!

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
bolav: I don't understand how this could be a problem with precision? I'm taking the values as is. No arithmetic on it, so it should literally be the same value. 

So the idea is that `majorRest` will store the remaining value of each row. This could be done like `placements`, but it would be more involved, because several elements share the same row. So if we would like to do this we would save the index of the last time we changed row, and loop over the elements on this row, and update the available space for each of these. It's possible, but I like the current solution better.

bolav:
@mortoray Just to be clear:

The existing `RowAlignment` will align the content on the row in the minor direction of the `WrapPanel`. In addition we need a way to align the content on the row in the major direction. That is we need to be able to begin, center and end justify it on the individual rows, in addition to doing that for each element within the row. 

If I'm understanding what you are saying correctly you would like `ContentAlignment` to also grow the element (`WrapLayout`) in the opposite direction of the alignment, so that if the parent of the panel is 400 points high, and then 2 rows of 100 points, they should be located at y 200-300 and y 300-400? If that is correct I think we need an additional attribute. Maybe we could use `RowAlignment` for both major and minor alignment, and then implement `ContentAligment` later, for alignment of the whole content inside the `WrapPanel`?

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
mortoray: `minorUsed` is a calculated value.

You should only need one `majorRest` per row, shouldn't you?  It's just the left-over major space after sizing each element.

mortoray:
@bolav You marked `RowAlignment` deprecated though, as though it won't be supported!

What I want is that the default `ContentAlignment` is the `Alignment` from the panel where it is used. This alignment then decides the overall alignment of the entire block of the wrap panel content. The `RowAlignment` then should also default to this value, but still allow an override.

Let me create a picture.

mortoray:
screenshot_20170710_160936

The `ContentAlignment`  decides the overall alignment of the "entire" content in the panel. This is generally how it is interpreted elsewhere.  It also sets the default for the `RowAlignment` (which makes sense visually).

The `RowAlignment` decides how the items are layed out in a row based on the available space. In the minor direction it applies per element, in the major direction it applies to the whole row.

bolav:
Yes. I deprecated the `RowAlignment` because my implementation of `ContentAlignment` just extended the behaviour of `RowAlignment`. From what I see from your drawing you want `RowAlignment` to be extended and keep being named `RowAlignment`, and then we can add `ContentAlignment` later. Is this okay?

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
bolav: Yes. So the alternative to the solution I made is to have an int array of which row each element is on, and then a float array of majorRest per row. The current implementation is easier, but this one might be easier to understand. I'm still not convinced that is possible to miss on the `Dictionary` though, without a repro.

mortoray: That, or take the `Grid` approach which stores the Row/Column number on the properties of the elements. It's a bit more costly and not necessary here though I think. Your array idea is good.

mortoray:
@bolav Yes, it appears you're adding the other dimension to `RowAlignment`, not introducing `ContentAlignment`.  I wonder if this can be done in a backwards compatible fashion, we might have to give it a new name. Maybe `ItemAlignment`?

To make this future compatbile with `ContentAlignment`, make it default to the container's `Alignment`. Though it shouldn't be too hard `ContentAlignment` as well (it's about the same as what you did now, just on a different placement area).

bolav:
@mortoray I think it can be added backwards compatible. I will look into it.

bolav:
@mortoray I'm fine either way about the default value. Since this will break backwards compatibility I don't feel qualified to make this call.

kusma:
@mortoray: I'm guessing you're good with it in this shape, then? If so, I'll just pull it as-is.

mortoray:
@kusma Yes, I'm happy with this.

PR #187, opened 2017-07-07
Cleaned up interface semantics & added docs in Fuse.Reactive

Duckers:
This PR cleans up and documents more precisely the semantics around the interfaces in `Fuse.Reactive`. This only changes internal APIs, for now. The intention is to later make these APIs public to allow third parties to implement data providers to the data binding system.

This PR also introduces the `IObservableObject` interface and makes the data binding system respect it, which is needed to implement arbitrary data providers.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,11 +1,12 @@
 # Unreleased
 
 ## Fuse.Reactive cleanup (Uno-level)
-- These changes shouldn't affect your code unless you are specifically dealing with the `Fuse.Reactive` interfaces in your Uno code (unlikely): 
+- These changes shouldn't affect your code unless you are working against internal interfaces in `Fuse.Reactive` interfaces (unlikely): 
   * Introduced `IObservableArray` as an intermediary interface between `IArray` and `IObservable`, which holds the `Subscribe()` method.
   * Changed the return type of `IObservableArray.Subscribe` method from `ISubscription` to `IDisposable`. The method *may* still return
     an `ISubscription`, but this is no longer guaranteed. The subscriber must check this manually (with `value is ISubscription`) if an 
     `ISubscription` is anticipated.
+  * Introduced `IObservableObject` as a way for `IObject`s to propagate property changed events.
 - Added docs for many of the interfaces in the `Fuse.Reactive` namespace.
 
 ## Templates
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Unreleased
 
+## Fuse.Reactive cleanup (Uno-level)
+- These changes shouldn't affect your code unless you are specifically dealing with the `Fuse.Reactive` interfaces in your Uno code (unlikely): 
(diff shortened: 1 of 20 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
...
 
+## Fuse.Reactive cleanup (Uno-level)
+- These changes shouldn't affect your code unless you are working against internal interfaces in `Fuse.Reactive` interfaces (unlikely): 
+  * Introduced `IObservableArray` as an intermediary interface between `IArray` and `IObservable`, which holds the `Subscribe()` method.
+  * Changed the return type of `IObservableArray.Subscribe` method from `ISubscription` to `IDisposable`. The method *may* still return
+    an `ISubscription`, but this is no longer guaranteed. The subscriber must check this manually (with `value is ISubscription`) if an 
+    `ISubscription` is anticipated.
+  * Introduced `IObservableObject` as a way for `IObject`s to propagate property changed events.
+- Added docs for many of the interfaces in the `Fuse.Reactive` namespace.
+
```
kusma: I don't think we should start listing purely internal changes like this. Looks like this PR shouldn't have a changelog to me.

Source/Fuse.Common/Tests/FuseTest/ObservableCollector.uno
```diff
@@ -55,7 +55,7 @@ namespace FuseTest
 			CleanSubscription();
 			if (_items == null)
 				return;
-			_subscription = _items.Subscribe(this);
+			_subscription = (ISubscription)_items.Subscribe(this);
```
kusma: ...however, in the changelog above, you write that IObservableArray.Subscribe() isn't guaranteed to return an ISubscription. This code assumes it does, which seems like a bad assumption.

Source/Fuse.Reactive.Bindings/Instance.uno
```diff
@@ -694,10 +695,13 @@ namespace Fuse.Reactive
 		string _matchKey;
 
 		/** Name of the field on each data object which selects templates for the data objects.
-
+
```
kusma: Please don't add random indentations like this.

Source/Fuse.Common/Tests/FuseTest/ObservableCollector.uno
Duckers: This test has to assume this, because it is testing writebacks to observables. It will throw an exception otherwise.

kusma: @Duckers: So, are you saying that in this case, it's guaranteed? I mean, the test will obviously fail if it doesn't, it just sounds a bit fragile...

kusma: It doesn't seem like the class will actually try to do anything with that depends on ISubscription... Why is this cast useful for the test? Couldn't it just depend on the IDisposable-ism here?

Duckers:
Will make another PR when the entire feature is ready for review

PR #192, opened 2017-07-10
iOS: Make InputDispatch expand number of activePointers

nekronos:
There is no public API on iOS that will tell us how many simultanious touches are supported. When writing this code I believed that 10 fingers were the max, but when testing this on iPad I found that iPad supports 11 simultanious touches. Also according to the internet iPhone supports only 5 simultanious touches.

Refactoring to just use a list instead so that we can support an arbitrary number of active pointers

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Native/iOS/InputDispatch.uno
+++ b/Source/Fuse.Controls.Native/iOS/InputDispatch.uno
@@ -103,14 +103,12 @@ namespace Fuse.Controls.Native.iOS
 			return new RootInfo(rootVisual, rootView);
 		}
 
-		const int MaxPointerCount = 10;
-
-		static ObjC.Object[] _activePointers = new ObjC.Object[MaxPointerCount];
+		static List<ObjC.Object> _activePointers = new List<ObjC.Object>();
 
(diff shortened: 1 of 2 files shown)


Source/Fuse.Controls.Native/iOS/InputDispatch.uno
```diff
@@ -103,29 +103,31 @@ namespace Fuse.Controls.Native.iOS
...
 				if (CompareUITouch(pointer, null) && firstUnused < 0)
 					firstUnused = i;
 				else if (CompareUITouch(pointer, uiTouch))
 					return i;
 			}
-
+			if (firstUnused < 0)
+			{
+				_activePointers.Add(uiTouch);
+				firstUnused = _activePointers.IndexOf(uiTouch);
```
kusma: This IndexOf is a bit pointless; you can just take `_activePonters.Count - 1` ;)

nekronos: 👍

PR #194, opened 2017-07-10
UX expressions name-value pairs

Duckers:
Adds support for Name-value pairs.

Requires new Uno stuff with https://github.com/fusetools/uno/pull/1222

This feature enables things like `<Button Clicked="goto('settings', {enabled: true})" />`

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Common/NameValuePair.uno
+++ b/Source/Fuse.Common/NameValuePair.uno
@@ -1,0 +1,56 @@
+using Uno;
+using Uno.Collections;
+
+namespace Fuse
+{
+	/** Represents a name-value pair, as denoted by `name: value` in UX expressions.
+		
+		Implements `IObject`, which means the `NameValuePair` can be viewed as an object
+		with a single property.
+
+		An `IArray` containing some `NameValuePairs` can be converted to an `IObject`
+		implementation containing all those properties using the `ObjectFromArray` method.
+	*/
+	public sealed class NameValuePair : IObject
+	{
+		public string Name { get; private set; }
+		public object Value { get; private set; }
+		public NameValuePair(string name, object value)
+		{
+			Name = name;
+			Value = value;
+		}
+
+		public override string ToString()
+		{
+			return "(" + Name + ": " + Value + ")";
+		}
+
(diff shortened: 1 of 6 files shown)


Source/Fuse.Common/NameValuePair.uno
```diff
@@ -0,0 +1,56 @@
...
+				if (key != Name) throw new ArgumentException("Object (NameValuePair) does not contain the given key");
+				return Value;
+			}
+		}
+
+		/** Creates an IObject implementation from an `IArray` of `NameValuePair`.
+			The array doesn't need to contain exclusively `NameValuePair` instances. Objects of other
+			types are ignored.
+		*/
+		public static IObject ObjectFromArray(IArray list)
```
mortoray: Why is this here? Nothing uses it.

Source/Fuse.Reactive.Expressions/NameValuePair.uno
```diff
@@ -0,0 +1,21 @@
...
+
+		[UXConstructor]
+		public NameValuePair([UXParameter("Name")] string name, [UXParameter("Value")] Expression value) : base(value)
+		{
+			Name = name;
+		}
+
+		protected override object Compute(object value)
+		{
+			return new Fuse.NameValuePair(Name, value);
```
mortoray: Since you create a new object everytime we could make `name` an `Expression` as well, so long as it evaluates to a string.

Duckers: We could, but I figured it would be a rarely used feature and it adds quite a lot of cost. We could add this later if it turns out to be in high demand.

Source/Fuse.Common/NameValuePair.uno
Duckers: It would be needed pretty soon. Any feature that actualy relies on NameValuePairs will probably want to convert an IArray of those to an IObject.

Duckers:
Needs to be rebased on master after #195 is merged

PR #195, opened 2017-07-10
Support for variable arguments to UX functions.

Duckers:
Inheriting  from the `Fuse.Reactive.VarArgFunction` class gives variable argument support.

The classes `Vector2`, `Vector3` and `Vector4` in `Fuse.Reactive` are now removed and replaced with the general purpose, variable-argument version `Vector` instead. This ensures vectors of any length are treated the same way. These changes are covered by lots of existing tests.

Needs new Uno stuff with https://github.com/fusetools/uno/pull/1223

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Marshal/Marshal.Cast.uno
+++ b/Source/Fuse.Marshal/Marshal.Cast.uno
@@ -251,6 +251,7 @@ namespace Fuse
 			if (o is Size2) return (Size2) o;
 			else if (o is Size) return new Size2((Size)o, (Size)o);
 			else if (o is string) return StringToSize2((string)o);
+			else if (o is IArray) return ToSize2(ArrayToVector(o));
 			else return new Size2(ToFloat2(o).X, ToFloat2(o).Y);
 		}
 
--- a/Source/Fuse.Marshal/Marshal.Cast.uno
+++ b/Source/Fuse.Marshal/Marshal.Cast.uno
@@ -251,7 +251,7 @@ namespace Fuse
 			if (o is Size2) return (Size2) o;
 			else if (o is Size) return new Size2((Size)o, (Size)o);
 			else if (o is string) return StringToSize2((string)o);
-			else if (o is IArray) return ToSize2(ArrayToVector(o));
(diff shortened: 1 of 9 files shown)


Source/Fuse.Marshal/Marshal.Vector.uno
```diff
@@ -0,0 +1,37 @@
...
+				return float3(Marshal.ToFloat(arr[0]), Marshal.ToFloat(arr[1]), Marshal.ToFloat(arr[2]));
+			}
+			else if (arr.Length == 4)
+			{
+				return float4(Marshal.ToFloat(arr[0]), Marshal.ToFloat(arr[1]), Marshal.ToFloat(arr[2]), Marshal.ToFloat(arr[3]));
+			}
+			else throw new MarshalException(arr, typeof(float4));
+		}
+
+		static object ArrayToVector(object arg)
```
mortoray: Can the name add a `If` or `Maybe` to indicate it doesn't always return a vector, only if the input is an array.

Source/Fuse.Marshal/Marshal.Cast.uno
```diff
@@ -251,6 +251,7 @@ namespace Fuse
 			if (o is Size2) return (Size2) o;
 			else if (o is Size) return new Size2((Size)o, (Size)o);
 			else if (o is string) return StringToSize2((string)o);
+			else if (o is IArray) return ToSize2(ArrayToVector(o));
```
kusma: I think it would be better to do `return ToSize2(ToVector((IArray)o));`, because the cast can't fail.

kusma:
> Source\Fuse.Reactive.Bindings\Tests\UX\Expression.Anchor.ux(6): E8001: Type not found: Fuse.Reactive.Vector2
> Source\Fuse.Reactive.Bindings\Tests\Fuse.Reactive.Bindings.Test.unoproj: E0000: Object reference not set to an instance of an object.

Missing new uno-stuffs?

Issue #197, opened 2017-07-11
TextAlignment broken on TextInput in Fuse 1.1.0


eksperts:
[Reported on community](https://www.fusetools.com/community/forums/bug_reports/textinput_not_respecting_textalignment_properly).
Tested on Windows 10 and macOS.
Fuse 1.1.0 build 13808.

Affected: local preview and dotnet build, Android preview and build.
Not affected: iOS preview and build.

Repro:
```
<App>
    <TextInput Alignment="VerticalCenter" TextAlignment="Center" PlaceholderText="This Should be centered." Margin="8" />
</App>
```

On local preview, the caret is aligned to the very left edge of the `TextInput`, when it should be aligned to the left edge of the `PlaceholderText` value (it does that on iOS). When you start typing, caret moves to the expected place and the content you enter appears correctly centered. When you release focus, the content stays centered too which is as expected.

On Android, the `PlaceholderText` is aligned to the left (should be in center), then when you focus the `TextInput` it becomes centered and looks exactly like it would on iOS which is as expected, and it's fine while you're typing; no problems with caret position either. Then when you release focus, the content aligns to the left again.

fusebuild:
The forum thread [Textinput not respecting textalignment properly](https://www.fusetools.com/community/forums/bug_reports/textinput_not_respecting_textalignment_properly) posted by krister@dealpass.no was linked to this issue.

nekronos:
Not broken in `release-1.0`. git bisect incoming

PR #198, opened 2017-07-11
Android: Fix TextAlignment regression

nekronos:
Fixes https://github.com/fusetools/fuselibs-public/issues/197

Regressed in f791ebbc8a20b0d53b6d94a0ae392eab5cfa96ee

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests


nekronos:
@kusma maybe this should go on `release-1.1` if we are doing a patch release?

kusma:
@nekronos: There's still a long time until we release 1.2, so yeah, I think we should prepare it for a 1.1 patch-release.

PR #200, opened 2017-07-11
Update templates

Sunjammer:
This PR implements the following template changes:

* New app template with folder structure and model js
* New camera app template
* New master/detail template
* New Empty template (the current app template)
* New "package" template (empty with unoproj with IsTransitive)

The PR depends on https://github.com/fusetools/Fuse/pull/4023

### DIFF
--- a/ProjectTemplates/Projects/App-Camera/Pages/Lightbox/LightBoxPage.js
+++ b/ProjectTemplates/Projects/App-Camera/Pages/Lightbox/LightBoxPage.js
@@ -1,13 +1,14 @@
 var App = require("_$filename$_");
 var images = App.getPictures();
+
 module.exports = {
-	images : images,
-	currentIndex : this.Parameter.map(function(param){
-		for(var i=0;i<images.length;i++){
-			if(images.getAt(i).id == param.id){
+	images: images,
+	currentIndex: this.Parameter.map(function (param) {
+		for (var i = 0; i < images.length; i++) {
+			if (images.getAt(i).id === param.id) {
 				return i;
 			}
 		}
 		return 0;
 	})
-}
+};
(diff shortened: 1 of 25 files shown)


ProjectTemplates/Projects/App-Camera/Pages/Lightbox/LightBoxPage.js
```diff
@@ -0,0 +1,14 @@
+var App = require("_$filename$_");
+var images = App.getPictures();
+module.exports = {
+	images : images,
+	currentIndex : this.Parameter.map(function(param){
+		var idx = images.length;
```
nekronos: Looks like `idx` is never used?

Sunjammer: Yeah it was used for a reverse while over the images at one point, removed now :)

kristianhasselknippe:
@Sunjammer: I think we should use ClientPanel instead of DockPanel + TopBar-/BottomBarBackground

Sunjammer:
Can you set the background on the statusbar with clientpanel?

kristianhasselknippe:
No i don't think so

Sunjammer:
Then we have a problem. You pretty much always have to set the background color on android or you'll get invisible statusbar icons.

kristianhasselknippe:
Then i agree we should keep the templates as is :)

Sunjammer:
I should probably add in the background color though to make the reasoning clear

COCPORN:
@Sunjammer I added a pull request to your repo from my "ign97"-user which cleans up JS and UX formatting. Will be back with some general comments on the templates.

Sunjammer:
Ruh roh, looks like we got ourselves an _illegal_

danhollick:
Is there anything happening with these templates? We need to know whether or not to accommodate space for these in the dashboard.

kusma:
@danhollick: Dunno. This PR stalled due to lacking reviews and CLA-signing. The latter has been fixed, and if someone who feels they know what this is about would review it, I'd be happy to merge it.

Sunjammer:
These templates need to be redone to reflect new best practises. They're sadly outdated already.

kusma:
Then I suggest we close this PR, and reopen once that's fixed...

Issue #205, opened 2017-07-11
Go over UX functions and take advantage of VarArgs where appropriate


Duckers:
For example, `index(elm)`  should be possible to write as just `index()` when you want to reference the local node. It should also take an extra optional argument; `index(element, instantiator)` where `instantiator` can be different from the most local instantiator. For example when you have nested Each's, you might want an inner element to use the index from the outer Each.

Ref discussions like https://github.com/fusetools/fuselibs-public/issues/204

mortoray:
The forced parameter is a result of https://github.com/fusetools/fuselibs/issues/4199

PR #206, opened 2017-07-12
Index() with parameters

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/205

This PR contains:
- [-] Changelog: no change to released API yet
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Reactive.Bindings/InstantiatorFunction.uno
+++ b/Source/Fuse.Reactive.Bindings/InstantiatorFunction.uno
@@ -5,30 +5,31 @@ using Fuse.Reactive;
 
 namespace Fuse.Reactive
 {
-	public abstract class InstantiatorFunction : UnaryOperator
+	public abstract class InstantiatorFunction : VarArgFunction
 	{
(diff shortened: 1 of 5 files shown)


kusma:
> Source\Fuse.Reactive.Expressions\VarArgFunction.uno(45.32): E3104: Uno.Collections.IList<Fuse.Reactive.Expression> does not contain a member called 'Length'. Could you be missing a package reference?

Source/Fuse.Navigation/RouterModify.uno
```diff
@@ -106,8 +106,12 @@ namespace Fuse.Navigation
 		/** The operation style of the transition. */
 		public string Style { get; set; }
 		
+		public IArray Path { get; set; }
+		
 		protected override void Perform(Node n)
 		{
+			debug_log Path;
+
```
kusma: This seems unintentional

mortoray: Ugh, how did that get in there. :/

PR #217, opened 2017-07-12
fixed visibility error on JavaScript member

kristianhasselknippe:
Since RootableScriptModule is `internal` i made JavaScript._scriptModule `protected internal`.

The unodoc backend complains otherwise even though the uno compiler doesn't for some reason.

### DIFF
--- a/Source/Fuse.Reactive.JavaScript/JavaScript.uno
+++ b/Source/Fuse.Reactive.JavaScript/JavaScript.uno
@@ -25,7 +25,7 @@ namespace Fuse.Reactive
 		internal static ThreadWorker Worker { get { return _worker; } }
 
 		readonly NameTable _nameTable;
-		protected RootableScriptModule _scriptModule;
+		RootableScriptModule _scriptModule;
 		internal RootableScriptModule ScriptModule { get { return _scriptModule; } }
 
 		[UXConstructor]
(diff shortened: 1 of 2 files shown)


kusma:
I don't think this is correct. `protected internal` means "accessible to *either* derived classes or classes in the same project", not "accessible to derived classes in the same project". The same inconsistency still exists after your change; an internal class is potentially being exposed to an non-internal derived class.

I think the correct thing to do would be to mark it as `internal` only.

kristianhasselknippe:
Fixed @kusma. Thanks for pointing that out. I has misunderstood `protected internal` to mean `protected` _and_ `internal`.

kusma:
@kristianhasselknippe: Why is this only needed on master? This has been this way since 1.0, why didn't we have issues with this before?

kristianhasselknippe:
@kusma I built the docs with this fixed locally, but didn't get around to making a PR or issue for it. iirc..

kusma:
@kristianhasselknippe: OK, fair enough. Thanks! :)

kusma:
@kristianhasselknippe: I'm not sure what's up here. It seems this change somehow make the test fail. I've tried rebuilding two times, and it still fails in the same place after your change.

kusma:
Actually, seems this is a pure docs-generator bug, not a fuselibs bug. The code is legal in C#. So let's fix the docs-generator instead.

kusma:
Seems I misunderstood.

PR #220, opened 2017-07-13
Trigger Action Node

mortoray:
Corrects/changes what node a trigger provides to an action. Updates all actions to work appropriately, including extend their ability to search upwards.
This branch is preparation for another branch. I wanted to separate out this change.

This PR contains:
- [x] Changelog
- [x] Documentation -- minimal added, but mostly no new public API
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.ScrollView/ScrollView.uno
+++ b/Source/Fuse.Controls.ScrollView/ScrollView.uno
@@ -377,7 +377,12 @@ namespace Fuse.Controls
 		*/
 		public void GotoRelative( float2 position )
 		{
-			Goto( (MaxScroll - MinScroll) * position + MinScroll );
+			Goto( RelativeToAbsolutePosition(position) );
+		}
+		
+		public float2 RelativeToAbsolutePosition( float2 pos )
+		{
+			return MinScroll + (MaxScroll - MinScroll) * pos;
 		}
 		
 		float2 FromScalarPosition( float value )
--- a/Source/Fuse.Controls.ScrollView/ScrollView.uno
+++ b/Source/Fuse.Controls.ScrollView/ScrollView.uno
@@ -380,7 +380,7 @@ namespace Fuse.Controls
 			Goto( RelativeToAbsolutePosition(position) );
 		}
 		
-		public float2 RelativeToAbsolutePosition( float2 pos )
+		internal float2 RelativeToAbsolutePosition( float2 pos )
 		{
 			return MinScroll + (MaxScroll - MinScroll) * pos;
 		}
--- a/Source/Fuse.Navigation/TriggerActions.uno
+++ b/Source/Fuse.Navigation/TriggerActions.uno
@@ -12,10 +12,7 @@ namespace Fuse.Navigation
 
 		protected override void Perform(Node n)
 		{
-			var v = n as Visual;
-			if (v == null) return;
-
-			var ctx = NavigationContext ?? Navigation.TryFind(v);
+			var ctx = NavigationContext ?? Navigation.TryFind(n);
 			if (ctx == null)
 			{
 				Fuse.Diagnostics.UserError( "No navigation context was found", this );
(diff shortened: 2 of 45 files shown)


Source/Fuse.Nodes/Node.uno
```diff
@@ -156,5 +156,22 @@ namespace Fuse
 			}
 			return null;
 		}
+		
+		public T FindByType<T>() where T : class
```
mortoray: I loosened it because I need to look for things that aren't nodes. In particular I'm looking for an interface. Even though I know what I find will be a node, the type I'm looking for isn't.

kusma: Right, I guess that makes sense. So we're not interested in narrowing it down later either, sounds good.

Source/Fuse.Controls.ScrollView/ScrollView.uno
```diff
@@ -377,7 +377,12 @@ namespace Fuse.Controls
 		*/
 		public void GotoRelative( float2 position )
 		{
-			Goto( (MaxScroll - MinScroll) * position + MinScroll );
+			Goto( RelativeToAbsolutePosition(position) );
+		}
+		
+		public float2 RelativeToAbsolutePosition( float2 pos )
```
kusma: Is this really a good name? "Absolute position" can have multiple meanings, for instance "not relative to the position of the element", which isn't what this deals with... Maybe we should make this function internal to avoid making this a part of our public API?

kusma: To make it a bit clear, I think "absolute position" should refer to a position that isn't relative to *anything*, not just "not relative to scroll position"...

mortoray: We refer to this as "absolute" elsewhere in the ScrollView, so I'd have to change all doc references. It's just a contrast to "relative" position"...

...but I'll just make this function internal so we don't have to worry about it now.

Source/Fuse.Triggers/Actions/Visibility.uno
```diff
@@ -31,7 +31,11 @@ namespace Fuse.Triggers.Actions
 	{
 		protected override void Perform(Node target)
 		{
-			if (target is IShow) ((IShow)target).Show();
+			var t = target.FindByType<IShow>();
```
kusma: I really wish TriggerActions would be nodes so these could be found while rooting... scanning along the tree could be a tad expensive :/

Source/Fuse.Navigation/TriggerActions.uno
```diff
@@ -115,10 +112,7 @@ namespace Fuse.Navigation
 
 		protected override void Perform(INavigation ctx, Node n)
 		{
-			var v = n as Visual;
-			if (v == null) return;
-
-			var page = Navigation.TryFindPage(Target ?? v);
+			var page = Navigation.TryFindPage((Node)Target ?? n);
```
kusma: I don'tthink you need the cast here. `Target` is a `Visual`, and a `Visual` is a `Node`...

Source/Fuse.Triggers/Actions/Visibility.uno
mortoray: In most cases it'll find the node just 2 parent's away. For most actions we actually wish to defer the scanning for perofrmance reasons. It'd be bad if all actions resolved themselves during the rooting phase -- it'd take a lot longer.  Of course, it'd be nicer for detecting errors statically. :/

Source/Fuse.Navigation/TriggerActions.uno
mortoray: I think I only added it since it complained. The left/right sides need to be the same type I think.

kusma: Ah, you're right.

kusma: I actually think it doesn't in C#, but I think we have a bug in our C++ code generator that means the resulting code won't compile.

PR #221, opened 2017-07-13
Feature defer pushnotification permissions ios

cbaggers:
Some users were having issues with the fact that users were being asked for permission to user push notifications immediately on launch. This PR gives the option to opt out of auto-registration and start it manually using a JS call.

This PR contains:
- [x] Changelog
- [x] Documentation
- [ ] Tests (none as we dont have a mechanism for testing PN yet)


CHANGELOG.md
```diff
@@ -185,6 +198,8 @@
 ## 1.0.2
 
 This release only upgraded Uno.
+## Bugfixes
```
mortoray: This doesn't seem related to your fix?

Source/Fuse.PushNotifications/JS.uno
```diff
@@ -125,5 +126,17 @@ namespace Fuse.PushNotifications
...
+
+		/**
+			@scriptmethod register
+
+			Registers the app with APNS. Only neccesary if APNS.RegisterOnLaunch was
+			set to false in the unoproj file.
+		*/
+		public object Register(Context context, object[] args)
+		{
+			Fuse.PushNotifications.PushNotify.Register();
```
mortoray: for future-proofing it's best to check the number of arguments and issue UserError's if not 0.

Source/Fuse.PushNotifications/iOS/AppDelegatePushNotify.mm
```diff
@@ -9,17 +9,9 @@ - (void)application:(UIApplication *)application initializePushNotifications:(NS
...
-		[application registerUserNotificationSettings: [UIUserNotificationSettings settingsForTypes:(UIUserNotificationTypeSound  | UIUserNotificationTypeAlert | UIUserNotificationTypeBadge)  categories:nil]];
-		[application registerForRemoteNotifications];
-	} else {
-		// use registerForRemoteNotificationTypes:
-		[application registerForRemoteNotificationTypes:
-		 UIRemoteNotificationTypeBadge |
-		 UIRemoteNotificationTypeSound |
-		 UIRemoteNotificationTypeAlert];
-	}
+#if (!@(Project.iOS.PushNotifications.RegisterOnLaunch:IsSet)) || @(Project.iOS.PushNotifications.RegisterOnLaunch:Or(0))
```
mortoray: This seems really odd to have fuselibs code statically differ based on user project settings. Do we do this anywhere else?  Could this not be a run-time check instead somehow?

CHANGELOG.md
cbaggers: Oh odd. thanks I'll find out what happened there

Source/Fuse.PushNotifications/JS.uno
cbaggers: thanks. Won't change the others and previous reviews have preferred not touching code not related to the PR

Source/Fuse.PushNotifications/iOS/AppDelegatePushNotify.mm
cbaggers: In fuselibs the only other place we use it with `#if` is Fuse.Platform/Android/SystemUI.uno#L117

A similar approach has been used to [splice in the the android senderID](https://github.com/fusetools/fuselibs-public/Source/Fuse.PushNotifications/Android/Impl.uno#L45) in push (which has been then for a year or two). In uxl files for local & push notifications and in WebView for the `Mobile.UriScheme`

In uno it's used in some [java code](Core/UnoCore/Targets/Android/@(Java.SourceDirectory)/com/fuse/App.java#L214) but mainly in build related uxl files.

It's not an ideal approach, I'd rather have the project data was included a queryable at runtime.

mortoray: Okay. Just thought I'd point it out...

...and as long as you can live with be hunting down by the guy that eventually tries to distribute precompiled libraries. :)

cbaggers: yeah. I've tried raising this point before and got nowhere. I'm just waiting now until this issue manifests and we are forced to talk about it. :(

cbaggers: Agreed. Have raised internally previously, I guess it's in a queue :)

Source/Fuse.PushNotifications/JS.uno
```diff
@@ -125,5 +126,22 @@ namespace Fuse.PushNotifications
...
+			@scriptmethod register
+
+			Registers the app with APNS. Only neccesary if APNS.RegisterOnLaunch was
+			set to false in the unoproj file.
+		*/
+		public object Register(Context context, object[] args)
+		{
+			if (args.Length != 0)
+			{
+				Fuse.Diagnostics.UserError( "Push.register takes no arguments");
```
kusma: Fails to compile:
> Source/Fuse.PushNotifications/JS.uno(140.22): E3128: Call to 'Fuse.Diagnostics.UserError(string,object,[string],[int],[string],[Uno.Exception])' has some invalid arguments (string)

cbaggers: thanks

Issue #223, opened 2017-07-14
EdgeNavigator with Activated is not working


mortoray:
https://github.com/fusetools/support-idesse/issues/29

PR #224, opened 2017-07-14
Fixes Activation and EdgeNavigator

mortoray:
I'm not entirely happy with how I fixed EdgeNavigator but I saw no other simple short-term solution, especially not for a hotfix. A lot of the clarified stuff was undefined before, so it shouldn't be backwards incompatible. Some stuff is still undefined unfortunately.

fixes https://github.com/fusetools/fuselibs-public/issues/223

This PR contains:
- [x] Changelog
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 # 1.1
 
+### Navigation
+- Fixed an issue where `Activated` and `WhileActivated` within an `EdgeNavigator` did not correctly identify an active state
+- Changed `EdgeNavigation` to return a page in `Active` when no side-panels are active
+- Added `Init` to `RoutePageProxy` that needs to be called to complete initialization -- the previous setup was broken.
+
 ### Fuse.Share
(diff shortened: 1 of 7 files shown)


kusma:
@mortoray: This fails a test:

> FAILED:  Fuse.Gestures.Test.Fuse.Gestures.Test.GestureTest.Layer
>         Assertion failed in 'Layer', at C:\projects\fuselibs-public\Source\Fuse.Gestures\Tests\Gesture.Test.uno:31
>           Expected: null
>           But got : Fuse.Controls.PageControl, Name: edgeMain

mortoray:
@kusma I'd expect that test to fail, it needs to be fixed... but it's not on the 1.1 branch. Ah. I marked the PR incorrectly. I'll update the PR's merge target.

mortoray:
@kusma How do I trigger a rerun after changing the target branch?

And how can I deal with fixing the test on `master`? Will you just let me know once you merge back to master and I can fix it?

mortoray:
Travis has the base branch confused.

PR #225, opened 2017-07-17
Fixes Activation and EdgeNavigator

mortoray:
I'm not entirely happy with how I fixed EdgeNavigator but I saw no other simple short-term solution, especially not for a hotfix. A lot of the clarified stuff was undefined before, so it shouldn't be backwards incompatible. Some stuff is still undefined unfortunately.

fixes https://github.com/fusetools/fuselibs-public/issues/223

This PR contains:
- [x] Changelog
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 # 1.1
 
+### Navigation
+- Fixed an issue where `Activated` and `WhileActivated` within an `EdgeNavigator` did not correctly identify an active state
+- Changed `EdgeNavigation` to return a page in `Active` when no side-panels are active
+- Added `Init` to `RoutePageProxy` that needs to be called to complete initialization -- the previous setup was broken.
+
 ### Fuse.Share
 - Fixed a crash in the iOS implementation for Fuse.Share that could happen on iPad.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,6 @@
 ### Navigation
 - Fixed an issue where `Activated` and `WhileActivated` within an `EdgeNavigator` did not correctly identify an active state
 - Changed `EdgeNavigation` to return a page in `Active` when no side-panels are active
-- Added `Init` to `RoutePageProxy` that needs to be called to complete initialization -- the previous setup was broken.
 
 ### Fuse.Share
 - Fixed a crash in the iOS implementation for Fuse.Share that could happen on iPad.
--- a/Source/Fuse.Navigation/RoutePageProxy.uno
+++ b/Source/Fuse.Navigation/RoutePageProxy.uno
@@ -65,21 +65,23 @@ namespace Fuse.Navigation
 		}
 		List<Level> _levels = new List<Level>();
 		
+		/** `Init()` must be called to finish initialization. */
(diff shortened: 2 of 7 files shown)


Source/Fuse.Navigation/RoutePageProxy.uno
```diff
@@ -65,21 +65,23 @@ namespace Fuse.Navigation
...
 		{
 			_activeChanged = activeChanged;
-			Init(source);
+			_source = source;
 		}
-		
-		void Init( Visual source )
+
+		/** Call after configuration is completed. This could start the actual registration and messaging */
+		public void Init()
```
kusma: I'm a bit skeptical to this on a release-branch. It's technically an API-change, although not a very big one.

It looks to me like the setter for `RoutePageProxy.Path` could be used to skunk this in on the release-branch without touching the API, as all places we need this seems to call that setter right before where you add the Init()-call here...

That being said, the Init-approach seems much nicer. So that's probably what we want going forward.

mortoray: I thought about another approach and this one is more likely to not be fail. It's definitely an API failure in the way it was setup before. Making a live update to a `Path` change take effect would have been complicated, since it would already be in the middle of updating subscriptiosn. So this was the more obvious solution.

Since this API was first introduced in 1.1 though, and already broken, I think we're safe in changing it in the patch. Nobody could be using it effectively.

mortoray: I'm actually uncertain why a lot of this stuff ended up public at all. But that's even more of a change to make it all internal. :(

kusma: OK, the API being broken is a good argument for fixing things without being too worried. But, if all of this is new and broken, maybe we do a hard backpaddle and close down the API also?

mortoray: I can mark these interfaces as internal. I didn't since the old one was technically open, and it's kind of mean to make a new better one that is private and deprecate the old one without a replacement.  Though I'd much prefer for them both to be internal. It's a mechanims that will change.

mortoray: Actually RoutePageProxy is already internal and I marked NavigationPageProxy as deprecated. I think that's enough for now.

kusma: Doh. Then this API-change is no problem at all. In fact, it shouldn't even be mentioned in the changelog, as it isn't user-visible. Sorry for the noise ;)

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 # 1.1
 
+### Navigation
+- Fixed an issue where `Activated` and `WhileActivated` within an `EdgeNavigator` did not correctly identify an active state
+- Changed `EdgeNavigation` to return a page in `Active` when no side-panels are active
+- Added `Init` to `RoutePageProxy` that needs to be called to complete initialization -- the previous setup was broken.
```
kusma: Please remove this last entry; `RoutePageProxy` is internal.

PR #232, opened 2017-07-18
Fuse.Nodes: survive if there's no DataWatcher

kusma:
This seems to happen during shutdown in some tests, if enough
objects are being disposed. So let's try to survive in this case.

This was noticed while adding some disposing to a test, so let's
add that here also, so we have a chance to notice similar regressions
in this functionality in the future.

This PR contains:
- [x] Changelog
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # 1.0
 
+### FuseJS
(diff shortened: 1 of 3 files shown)


kusma:
I'm sending this against release-1.0, because it's present there as well. I doubt we'll bother releasing another 1.0 patch-release just for this, but if there's a better reason in the future, this should probably be included as well. So let's try to get it in there as well.

PR #234, opened 2017-07-18
Fuse.Reactive: survive Dispose after null-subscriptions

kusma:
Some subscriptions (like constant values that cannot change) simply
return a null-pointer, to avoid having to create needless objects.
This is considered legal by the expression-system.

However, the newly added array-lookup expression didn't take this
into account. Let's do this, by simply not disposing null-pointers.

This was discovered while adding test-disposal to the test for this
feature, which triggers the issue. So let's add that disposal around
this test now, so we'll detect if this regresses in the future. Some
minor clean-ups are also included.

This PR contains:
- [x] Tests


kusma:
The change to the test might look a bit intimidating; in reality it's just reindentation, because of a new using-statement.

PR #236, opened 2017-07-18
Allow NameValuePairs to have expression for names.

Duckers:
Needs new Uno stuff with https://github.com/fusetools/uno/pull/1235


kusma:
> Source/Fuse.Reactive.JavaScript/Tests/UX/NameValuePair.ux(9): E8001: Sequence contains no matching element

This sounds like it revealed a bug in the UX-compiler...

Duckers:
woops, on it!

kusma:
Also, why was the checklist removed from the PR-template? It's meant to make sure these things are considered, and tests are certainly appropriate in this case.

Duckers:
Superceded by the above PR

PR #240, opened 2017-07-18
Fuse.Nodes: add missing packages in InternalsVisibleTo

mortend:
This adds missing packages in InternalsVisibleTo which are using internal entities from Fuse.Nodes.
The current Uno version fails to give errors on this, but I'm working on a branch of Uno where this is fixed.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
@mortend: Could you do something like "git commit --amend --reset-author" and force-push? It looks like you sent this while AppVeyor were having issues, and I can't find another way of re-triggering the build.

mortend:
@kusma: Force-pushed now.

PR #241, opened 2017-07-18
Fix a visibility error

kristianhasselknippe:
The problem was that an internal enum was made visible through a protected
member which takes that type as an argument. Changing the visibility of
the member to internal instead.

Note: This only caused an issue for the docs generator.

### DIFF
--- a/Source/Fuse.Controls.ScrollView/ScrollView.Layout.uno
+++ b/Source/Fuse.Controls.ScrollView/ScrollView.Layout.uno
@@ -238,7 +238,7 @@ namespace Fuse.Controls
 		/**
 			The distance from the visible view area start or end for the provided position.
 		*/
-		protected float2 DistanceFromView(float2 position, DistanceFromViewTarget target)
+		internal float2 DistanceFromView(float2 position, DistanceFromViewTarget target)
 		{
 			float x = 0;
 			float y = 0;


nekronos:
it should not be `internal protected`?

kristianhasselknippe:
@nekronos : then it would still be exposed publicly since it would then be internal _or_ protected.

kristianhasselknippe:
or did you mean to make the enum protected internal?

mortoray:
`internal` is correct here. It's kind of annoying how C# interprets `internal protected`.

nekronos:
`internal protected` mean that it has `protected` visibility, but only internally. `internal` alone is the same as `internal public`

mortoray:
https://stackoverflow.com/questions/12661494/protected-internal
They say it means `internal` or `protected`.
I've had this come up before with docs and the code. There is no way to create the desired "internal and protected" modifiers? :(

kusma:
MSDN says
> The type or member can be accessed by any code in the assembly in which it is declared, or from within a derived class in another assembly. Access from another assembly must take place within a class declaration that derives from the class in which the protected internal element is declared, and it must take place through an instance of the derived class type.

The first sentence seems to suggest the *or*-interpretation. However, the second, more fleshed-out sentence, seems to suggest the *and*-interpretation.

This is very confusing. Maybe we should just test this in C#?

kristianhasselknippe:
> Access from another assembly must take place within a class declaration that derives from the class in which the protected internal element is declared

To me this says the exact same thing. If you create a subclass, it is available as a protected member.

kusma:
@kristianhasselknippe: Maybe I just read it wrong, then.

But how can such an interpretation even work? The internal type is unknown if the derived type is not in the same assembly, so it can't really do anything with it...

PR #244, opened 2017-07-19
Native view impl of the CoreGraphics surface API

nekronos:
This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortoray:
@nekronos Does the charting sample app work with this branch in native mode?

Source/Fuse.Drawing.Surface/CoreGraphics/NativeSurface.uno
```diff
@@ -0,0 +1,48 @@
...
+	extern(iOS||OSX)
+	class NativeSurface : CoreGraphicsSurface
+	{
+
+		IntPtr _cgContext = IntPtr.Zero;
+
+		public void Begin(IntPtr cgContext, float pixelsPerPoint)
+		{
+			// The CGContext is already scaled
+			_pixelsPerPoint = 1.0f;//pixelsPerPoint;
```
mortoray: Does this assume the pixelsPerPoint is the same as pixelsPerOSPoint?  I guess the entire native layer does that?

Source/Fuse.Controls.Panels/TreeRenderer.iOS.uno
```diff
@@ -86,7 +86,11 @@ namespace Fuse.Controls
 
 		void ITreeRenderer.Placed(Element e)
 		{
-			_elements[e].SetSize(e.ActualSize);
+			var viewHandle = _elements[e];
+			if (viewHandle.NeedsRenderBounds)
+				viewHandle.SetSizeAndVisualBounds(e.ActualSize, e.RenderBoundsWithoutEffects);
```
nekronos: I am using the renderbounds to position a `CALayer` inside a `UIView`, so as long as the Rect has a valid position it should be fine

nekronos: I recall observing renderbounds with both positive and negative values for position. And it looks correct in the Surface testapp

Source/Fuse.Drawing.Surface/CoreGraphics/NativeSurface.uno
nekronos: Fix in https://github.com/fusetools/fuselibs-public/pull/244/commits/db27316bd8ad9a759fe843f0ae8bb0cc42ff6227

PR #246, opened 2017-07-19
Fuse.Scripting.V8: do not release on the wrong context

### LINKED ISSUE #237: AccessViolationException in ScriptingTest.ExternalSameObject

kusma:
Sometimes when releasing JS-values after a context has been disposed,
we get an `AccessViolationException`, probably due to using different
contexts (or rather, null-contexts) than the actual context used.

the `_context == null`-case seems to be a similarly mis-guided attempt
at surviving, where we just end up releasing in a collected context.

Since disposing sets the _context-member to the same value as the
`_context == null`-case fudged it with, these two cases should actually
be the same, and just as potentially crashy.

Let's instead assume that disposing a context correctly destroy all of
its resources. There's no way we can do any better, because the handle
to the context is lost at this point.

This fixes #237

Unfortuantely, I was unable to come up with a way of testing this, as
it depends on a lot of hard-to-control things, like timing of finalizers,
and the memory mapping / unmapping pattern of the JS VM code
genrator.

### DIFF
--- a/Source/Fuse.Scripting.V8/Array.uno
+++ b/Source/Fuse.Scripting.V8/Array.uno
@@ -25,8 +25,8 @@ namespace Fuse.Scripting.V8
 
 		~Array()
 		{
-			var cxt = _context == null ? default(Simple.JSContext) : _context._context;
-			_array.AsValue().Release(cxt);
+			if (_context != null && !_context.IsDisposed)
+				_array.AsValue().Release(_context._context);
 		}
 
 		public override object this[int index]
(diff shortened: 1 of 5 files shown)


yupferris:
LGTM!

kusma:
I was initially thinking of including this in 1.1.1. However, it seems we only dispose `Fuse.Scripting.Context` in tests, so I don't think this matter unless the issue actually triggers.

yupferris:
I don't think I've ever heard of a user running into this, nor have I ever seen this myself, so I think it's safe to hold back.

kusma:
Whoops, seems Scripting.ExternalSameObject is disabled on master now, so I should also enable it again in this PR.

PR #250, opened 2017-07-19
Make sure tests dispose TestRootPanel

kusma:
This cleans up a bunch of tests, and most importantly, makes sure all `TestRootPanel`s
gets disposed before the next test runs. This should greatly help with earlier tests
affecting later tests.

This PR contains:
- [x] Tests
- [x] Tests
- [x] More tests


kusma:
If this is too hard to review, I could send the first 10 (misc cleanups) as a separate PR first. The following tests are all mechanical addings of using (and *some* opting to use helpers instead of explicitly newing TestRootPanel and calling Layout etc)... Just let me know.

Issue #247, opened 2017-07-19
GestureTest.Layer fails


kusma:
```
FAILED:  AllTests.Fuse.Gestures.Test.GestureTest.Layer
         Assertion failed in 'Layer', at C:\Users\erik\src\fuselibs\Source\Fuse.Gestures\Tests\Gesture.Test.uno:31
           Expected: null
           But got : Fuse.Controls.PageControl, Name: edgeMain
```

mortoray:
This is not a random failure. This is expected with the merge for the EdgeNAvigator fix that came in 1.1. I couldn't fix this test on that branch since it wasn't there.

kusma:
@mortoray: Oh, OK. Well, it's broken on master now. Would you be able to fix it now?

kusma:
Seems github has gotten more eager in how it closes issues. I just mentioned the issue in a commit message, yet it thought this meant it was closed. Silly github.

Anyway, I pushed a commit that ignored the test. Please un-ignore when it's fixed.

PR #251, opened 2017-07-19
fixed an issue with this test, a merge in 1.1 changed how EdgeNAvigat…

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/247

This PR contains:
- [x] Tests

### DIFF
--- a/Source/Fuse.Gestures/Tests/Gesture.Test.uno
+++ b/Source/Fuse.Gestures/Tests/Gesture.Test.uno
@@ -11,7 +11,6 @@ namespace Fuse.Gestures.Test
 	public class GestureTest : TestBase
 	{
 		[Test]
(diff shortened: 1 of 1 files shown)


kusma:
I think we also need this no release-1.2...

kusma:
@mortoray: Sorry if I'm being very forgetful here, but did we get this on release-1.2 as well? I can't find any traces of that... If not, could you please back-port this PR there?

mortoray:
@kusma There is another PR for 1.2 https://github.com/fusetools/fuselibs-public/pull/259 that has a commit with the same fix.

kusma:
Ah, never mind. This is a test, and it passes on release-1.2, so I guess everything is fine.

PR #258, opened 2017-07-20
Detects and fixes a regression in 1.2 with Each node removal

mortoray:
fixes a regression in 1.2 in Each for removing nodes

This PR contains:
- [-] Changelog: not yet released
- [-] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Nodes/Visual.BeginRemove.uno
+++ b/Source/Fuse.Nodes/Visual.BeginRemove.uno
@@ -84,7 +84,7 @@ namespace Fuse
 			}
 			else
 			{
-				args.Remove();				
+				UpdateManager.AddDeferredAction(args.Remove);
 			}
 		}
 
--- a/Source/Fuse.Reactive.Bindings/Tests/UX/Each.ReplaceWithLessData.ux
+++ b/Source/Fuse.Reactive.Bindings/Tests/UX/Each.ReplaceWithLessData.ux
@@ -1,20 +1,20 @@
 <Panel ux:Class="UX.Each.ReplaceWithLessData">
-    <JavaScript>
-        var Observable = require("FuseJS/Observable")
+	<JavaScript>
+		var Observable = require("FuseJS/Observable")
 
-        exports.monthsToMaturity = Observable(3);
-        exports.payments = Observable();
-    
-        exports.calculate = function() {
-            var payments = [];
-            for (var t = 0; t < exports.monthsToMaturity.value; t++) {
-                payments.push(t)
-            }
-            exports.payments.replaceAll(payments);
-        }
+		exports.monthsToMaturity = Observable(3);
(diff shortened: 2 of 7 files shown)


kusma:
Fails tests:
> Failed:  Fuse.Controls.Test.PanelTest.Issue1063
> Failed:  Fuse.Triggers.Test.TriggerTest.ChildOrder1
> Failed:  Fuse.Triggers.Test.TriggerTest.ChildOrder3
> Failed:  Fuse.Triggers.Test.TriggerTest.NoChild

kusma:
Also, the commit message and PR subject should be updated to match the new status.

Source/Fuse.Reactive.Bindings/Tests/UX/Each.ReplaceWithLessData.ux
```diff
@@ -0,0 +1,25 @@
...
+        exports.calculate = function() {
+            var payments = [];
+            for (var t = 0; t < exports.monthsToMaturity.value; t++) {
+                payments.push(t)
+            }
+            exports.payments.replaceAll(payments);
+        }
+
+        exports.calculate();
+    </JavaScript>
```
kusma: Mixed indentation for the JavaScript-tag.

Source/Fuse.Nodes/Visual.BeginRemove.uno
```diff
@@ -84,7 +84,7 @@ namespace Fuse
 			}
 			else
 			{
-				args.Remove();				
+				UpdateManager.AddDeferredAction(args.Remove);
```
kusma: This is probably going to require that a lot more tests do root.PumpDeferred(), which makes me skeptical to this change after testing has been performed.

mortoray: Yes, the tests have to call PumpDeferred. Virtuallly all code had to work with removal being async anyways, due to RemovingAnimation. So it's not as scary as it seems.  Most places probably already assumed it was async.

kusma: I still think this is the wrong time to change this timing. Swapping the order of two events is hard to reason about the consequences of, and we're already through testing.

kusma: We were able to do this correct on release-1.1 without having to change the tests. Why can't we get this working here?

mortoray: Because I fixed a memory leak on 1.1 which requires new removal code in Each.
I think there might be an option though... Let me investigate.

mortoray:
https://github.com/fusetools/fuselibs-public/pull/259#issuecomment-316744823 is probably safer for 1.2
I'll make this messenging change in master.

PR #259, opened 2017-07-20
Detects and fixes a regression in 1.2 with Each node removal (Alternate fix)

mortoray:
fixes a regression in 1.2 in Each for removing nodes

It's an alternate solution to https://github.com/fusetools/fuselibs-public/pull/258

This PR contains:
- [-] Changelog
- [-] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Gestures/Tests/Gesture.Test.uno
+++ b/Source/Fuse.Gestures/Tests/Gesture.Test.uno
@@ -28,7 +28,7 @@ namespace Fuse.Gestures.Test
 				//swipe away the right EgeNavigator panel
 				root.PointerSwipe(float2(810,50), float2(950,50), 100);
 				root.StepFrame(5);//stabilize
-				Assert.AreEqual(null, p.edge.Active);
+				Assert.AreEqual(p.edgeMain, p.edge.Active);
 
 				//swipe the RangeControl
 				root.PointerSwipe(float2(600,10), float2(400,10), 100);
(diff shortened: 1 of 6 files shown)


kusma:
Please update the commit messages of the first change to reflect what it actually does, not what it did back when the first version was written...

mortoray:
@kusma I'm not sure what is wrong in the commit message. It's still correct.

kusma:
@mortoray: The subject talks about the test, but that's not the main feature of the change.

mortoray:
@kusma I'm really not sure what the problem is, the commit message is: 

```
Added test for broken case. This test passes on release-1.1

fixes a regression in 1.2 in Each for removing nodes

alternate fix for the defect
```

I'm not sure how much more specific I should be?!

kusma:
@mortoray: In Git, the first line is considered the "subject", similar to the subject of an e-mail. My point is that this should summarize the *important* part of a commit. The addition of the test is no longer the imporant part here. For instance, when looking [here](https://github.com/fusetools/fuselibs-public/pull/259/commits), only the subject is shown, and the commit ends up being misrepresented. The same thing goes for `git log --oneline` etc.

Having a misleading subject can be worse than not having anything, because it makes people skip over commits they don't think can be relevant when searching for the culprit of something.

kusma:
There's already a bunch of not-ideal commit messages in there, so I guess we can take this as-is. But going forward, it'd be nice if people paid a bit more attention to the subject of the commit-messages, as it has some quote practical benefits.

PR #260, opened 2017-07-20
Implemented Visual.Children as an implicitly linked list (optimization)

Duckers:
This implementation avoids all memory allocation, copying and shifting in connection with manipulating the Children list, and all the list operations are now O(1), including Contains(), InsertAfter() and Remove()

Updated all relevant call sites to use this API over the IList<T> interface (which is slower).

This is a big optimization over using MiniList, shaving 2-3 seconds of a case where 5000 text elements where added to a list.

This PR contains:
- [x] Changelog
- [x] Documentation
- [-] Tests - These code paths are covered plenty by our existing tests

### DIFF
--- a/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
+++ b/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
@@ -134,9 +134,8 @@ namespace Fuse.Layouts
 			int currentRow = 0;
 			
 			int i = 0;
-			for (var cn = container.Children_first; cn != null; cn = cn.Children_next, i++)
+			for (var e = container.FirstChild<Visual>(); e != null; e = e.NextSibling<Visual>(), i++)
 			{
-				var e = cn as Visual;
 				if (!AffectsLayout(e))
 					continue;
 
@@ -184,10 +183,8 @@ namespace Fuse.Layouts
 				var saMin = IsVert ? AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment);
 				var saMaj = IsVert ? AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment);
 				var elp = lp.CloneAndDerive();
-				for (var cn = container.Children_first; cn != null; cn = cn.Children_next)
+				for (var element = container.FirstChild<Visual>(); element != null; element = element.NextSibling<Visual>())
 				{
-					var element = cn as Visual;
-					if (element == null) continue;
 					if (ArrangeMarginBoxSpecial(element, padding, lp ))
 						continue;
 
--- a/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
+++ b/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
@@ -82,21 +82,21 @@ namespace Fuse.Layouts
 		public string ID { get; set; }
 
 		internal override float2 GetContentSize(
-			Visual elements,
+			Visual container,
 			LayoutParams lp)
 		{
-			return Arrange(elements, lp, false);
+			return Arrange(container, lp, false);
 		}
 		
 		internal override void ArrangePaddingBox(
-			Visual elements,
+			Visual container,
 			float4 padding,
 			LayoutParams lp)
 		{
-			Arrange(elements, lp, true, padding);
+			Arrange(container, lp, true, padding);
 		}
 		
-		float2 Arrange(Visual elements, LayoutParams lp,	
+		float2 Arrange(Visual container, LayoutParams lp,	
 			bool doArrange, float4 padding = float4(0))
 		{
 			var nlp = lp.CloneAndDerive();
@@ -121,20 +121,20 @@ namespace Fuse.Layouts
 			if (_hasItemHeight)
 				clp.SetY(ItemHeight);
 
-			var placements = new float4[elements.Children.Count];
+			var placements = new float4[container.Children.Count];
 			//minorMaxSize in each major row, assinged per element
-			var minorSizes = new float[elements.Children.Count];
+			var minorSizes = new float[container.Children.Count];
 			// save the row each element is on
-			var elementOnRow = new int[elements.Children.Count];
+			var elementOnRow = new int[container.Children.Count];
 			// save the available space for each row
-			var majorRest = new float[elements.Children.Count];
+			var majorRest = new float[container.Children.Count];
 			//where this row starts
 			int majorStart = 0;
 			// current row
 			int currentRow = 0;
 			
 			int i = 0;
-			for (var cn = elements.Children_first; cn != null; cn = cn.Children_next, i++)
+			for (var cn = container.Children_first; cn != null; cn = cn.Children_next, i++)
 			{
 				var e = cn as Visual;
 				if (!AffectsLayout(e))
@@ -174,7 +174,7 @@ namespace Fuse.Layouts
 			}
 
 			//final bits
-			for (int j=majorStart; j < elements.Children.Count; ++j)
+			for (int j=majorStart; j < container.Children.Count; ++j)
 				minorSizes[j] = minorMaxSize;
 			majorMaxUsed = Math.Max(majorMaxUsed, majorUsed);
 			minorUsed += minorMaxSize;
@@ -184,7 +184,7 @@ namespace Fuse.Layouts
 				var saMin = IsVert ? AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment);
 				var saMaj = IsVert ? AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment);
 				var elp = lp.CloneAndDerive();
-				for (var cn = elements.Children_first; cn != null; cn = cn.Children_next)
+				for (var cn = container.Children_first; cn != null; cn = cn.Children_next)
 				{
 					var element = cn as Visual;
 					if (element == null) continue;
--- a/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
+++ b/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
@@ -183,7 +183,8 @@ namespace Fuse.Layouts
 				var saMin = IsVert ? AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment);
 				var saMaj = IsVert ? AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment);
 				var elp = lp.CloneAndDerive();
-				for (var element = container.FirstChild<Visual>(); element != null; element = element.NextSibling<Visual>())
+				i = 0;
+				for (var element = container.FirstChild<Visual>(); element != null; element = element.NextSibling<Visual>(), i++)
 				{
 					if (ArrangeMarginBoxSpecial(element, padding, lp ))
 						continue;
--- a/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
+++ b/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
@@ -188,6 +188,7 @@ namespace Fuse.Layouts
 				for (var n = container.FirstChild<Node>(); n != null; n = n.NextSibling<Node>(), i++)
 				{
 					var element = n as Visual;
+					if (element == null) continue;
 					if (ArrangeMarginBoxSpecial(element, padding, lp ))
 						continue;
 
--- a/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
+++ b/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
@@ -134,8 +134,9 @@ namespace Fuse.Layouts
 			int currentRow = 0;
 			
 			int i = 0;
-			for (var e = container.FirstChild<Visual>(); e != null; e = e.NextSibling<Visual>(), i++)
+			for (var n = container.FirstChild<Node>(); n != null; n = n.NextSibling<Node>(), i++)
 			{
+				var e = n as Visual;
 				if (!AffectsLayout(e))
 					continue;
 
@@ -184,8 +185,9 @@ namespace Fuse.Layouts
 				var saMaj = IsVert ? AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment);
 				var elp = lp.CloneAndDerive();
 				i = 0;
-				for (var element = container.FirstChild<Visual>(); element != null; element = element.NextSibling<Visual>(), i++)
+				for (var n = container.FirstChild<Node>(); n != null; n = n.NextSibling<Node>(), i++)
 				{
+					var element = n as Visual;
 					if (ArrangeMarginBoxSpecial(element, padding, lp ))
 						continue;
 
--- a/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
+++ b/Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
@@ -82,21 +82,21 @@ namespace Fuse.Layouts
 		public string ID { get; set; }
 
 		internal override float2 GetContentSize(
-			IList<Node> elements,
+			Visual elements,
 			LayoutParams lp)
 		{
 			return Arrange(elements, lp, false);
 		}
 		
 		internal override void ArrangePaddingBox(
-			IList<Node> elements,
+			Visual elements,
 			float4 padding,
 			LayoutParams lp)
 		{
 			Arrange(elements, lp, true, padding);
 		}
 		
-		float2 Arrange(IList<Node> elements, LayoutParams lp,	
+		float2 Arrange(Visual elements, LayoutParams lp,	
 			bool doArrange, float4 padding = float4(0))
 		{
 			var nlp = lp.CloneAndDerive();
@@ -121,21 +121,22 @@ namespace Fuse.Layouts
 			if (_hasItemHeight)
 				clp.SetY(ItemHeight);
 
-			var placements = new float4[elements.Count];
+			var placements = new float4[elements.Children.Count];
 			//minorMaxSize in each major row, assinged per element
-			var minorSizes = new float[elements.Count];
+			var minorSizes = new float[elements.Children.Count];
 			// save the row each element is on
-			var elementOnRow = new int[elements.Count];
+			var elementOnRow = new int[elements.Children.Count];
 			// save the available space for each row
-			var majorRest = new float[elements.Count];
+			var majorRest = new float[elements.Children.Count];
 			//where this row starts
 			int majorStart = 0;
 			// current row
 			int currentRow = 0;
 			
-			for (int i = 0; i < elements.Count;++i)
+			int i = 0;
+			for (var cn = elements.Children_first; cn != null; cn = cn.Children_next, i++)
 			{
-				var e = elements[i] as Visual;
+				var e = cn as Visual;
 				if (!AffectsLayout(e))
 					continue;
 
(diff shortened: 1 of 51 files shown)


Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
```diff
@@ -82,21 +82,21 @@ namespace Fuse.Layouts
...
-			IList<Node> elements,
+			Visual elements,
 			float4 padding,
 			LayoutParams lp)
 		{
 			Arrange(elements, lp, true, padding);
 		}
 		
-		float2 Arrange(IList<Node> elements, LayoutParams lp,	
+		float2 Arrange(Visual elements, LayoutParams lp,
```
mortoray: The name of the argument needs to change on these, `elements` is  misleading, especially when you see things like `elements.Children.Count` in the code.

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -29,24 +29,24 @@ namespace Fuse
...
-					return c;
+				var v = c as T;
+				if (v != null) return v;
+				c = c.Children_next;
 			}
 			return null;
 		}
 
-		public void RemoveAllChildren<T>()
+		public void RemoveAllChildren<T>() where T: Node
```
mortoray: This restriction really isn't needed, `where T : class` is enough. Somebody (not us though) might be calling this with an interface that doesn't derive from `Node`.

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -29,24 +29,24 @@ namespace Fuse
...
-			while (i >= 0)
+			var c = Children_first;
+			while (c != null)
 			{
-				if (Children[i] is T)
-					Children.RemoveAt(i);
-
-				i = Math.Min(i, Children.Count) - 1;
+				var v = c as T;
+				if (v != null) Children_Remove(v);
```
mortoray: Doesn't this imply the loop needs `Children_SafeForEach` or equivalent?  It's removing items from the list it is iterating.

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -167,19 +165,21 @@ namespace Fuse
 
 		void ICollection<Node>.Clear()
 		{
-			foreach (var child in _children)
+			foreach (var child in Children)
```
mortoray: Why isn't this using the linked list iteration?

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -258,18 +250,16 @@ namespace Fuse
 					var c = items.Current;
 					if (allowMove)
 					{
-						var where = IndexOf(c);
-						if (where != -1)
+						if (Children_Contains(c))
```
mortoray: (not about this line, about the HashSet)

I wonder if we can't assume that `InsertOrMove` is exclusive and add an intrinsic boolean flag to node to track the `moved` status done here. That would avoid the HashSet.

Source/Fuse.Nodes/Visual.Layout.uno
```diff
@@ -329,10 +329,13 @@ namespace Fuse
 			if (newValue != SnapToPixels)
 			{
 				SetBit(FastProperty1.ContextSnapToPixelsCache, newValue);
-				for (int i = 0; i < Children.Count; i++)
+
+				var c = Children_first;
+				while (c != null)
```
mortoray: Why did you switch to `while` here instead of the `for` loop used elsewhere?

Source/Fuse.Nodes/Visual.ZOrder.uno
```diff
@@ -14,9 +14,14 @@ namespace Fuse
 		{ 
 			get
 			{
-				if (!HasVisualChildren) return null;
-
-				return FirstChild<Visual>();
```
mortoray: I see no reason to change this bit since `FirstChild` uses the new iteration. You're just inlining a copy of it here.

Source/Fuse.Nodes/Visual.ZOrder.uno
```diff
@@ -142,15 +147,18 @@ namespace Fuse
 				0  // Layer.Overlay
 			};
 
-			for (int i = 0; i < nodes.Count; i++)
+			var c = Children_first;
+			while (c != null)
```
mortoray: Why not the `for` loop?

Source/Fuse.Reactive.Bindings/Instance.WindowItems.uno
```diff
@@ -175,7 +175,7 @@ namespace Fuse.Reactive
 			//find last node prior to where we want to introduce
 			var lastNode = GetLastNodeFromIndex(windowIndex-1);
 
-			Parent.InsertOrMoveNodes( Parent.Children.IndexOf(lastNode) + 1, wi.Nodes.GetEnumerator() );
+			Parent.InsertOrMoveNodes( lastNode, wi.Nodes.GetEnumerator() );
```
mortoray: The `InsertOrMoveNodes` name should probably be changed to `InsertOrMoveNodesAfter` to match the `InsertNodesAfter` function.

mortoray:
Your original idea for copy-on-modify for iteration is better. The `SafeForEach` would then look something more like this:

```
		class SafeIterator
		{
			public SafeIterator Previous;
			public Node[] List;
			public Node Current;
			public Action<Visual, Node> Action;
		}
		SafeIterator Children_CurrentIterator;

		void Children_SafeForEach(Action<Visual, Node> action)
		{
			Children_CompleteCurrentAction();
			
			var iter = new SafeIterator{ Current = Children_first,
				Action = action, 
				Previous = Children_CurrentIterator }; //we could use a simple pool if allocation is an issue
			Children_CurrentIterator = iter;
			while (iter.Current != null) 
			{
				iter.Action(iter.Current);
				//if interrupted use the copied list instead
				if (iter.List != null )
				{
					for (int i=0; i < iter.List.Length; ++i)
						iter.Action(iter.List[i]);
					break;
				}
				
				iter.Current = iter.Current.Children_next;
			}
			
			if (iterm.List == null)
				Children_CurrentIterator = iter.Previous;
		}

		void Children_CompleteCurrentAction()
		{
			var iter = Children_CurrentIterator;
			while (iter != null)
			{
				if (iter.List == null)
					iterList = Children_ToArrayAfter(iter.Current);
				
				iter = iter.Previous;
			}
			
			//no need to track current anymore
			Children_CurrentIterator = null;
		}
```

This avoids any deep stacks. It keeps the same processing order we had previously. It allows nesting of iteration.

You'd have to implement the `Children_ToArrayAfter` which is like `ToArray` but starting with the node after the provided one.

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
```diff
@@ -121,21 +121,22 @@ namespace Fuse.Layouts
 			if (_hasItemHeight)
 				clp.SetY(ItemHeight);
 
-			var placements = new float4[elements.Count];
+			var placements = new float4[elements.Children.Count];
```
Duckers: My idea is to refactor this so only the Visual.Children.LinkedList.uno file uses the weird `Children_*` symbols. All other files should use a "proper" interface

Source/Fuse.Nodes/Visual.Children.LinkedList.uno
```diff
@@ -0,0 +1,216 @@
...
+			if (child.Children_parentID != -1) throw new Exception();
+			child.Children_parentID = parent.Children_thisID;
+		}
+
+		void Children_MakeOrphan(Node child)
+		{
+			child.Children_parentID = -1;
+		}
+
+		internal Node Children_first;
```
Duckers: As this is a mixin-like pattern, I use this naming convention to show that something fishy is going on. I'm thinking of this as a field of the `Children` mixin.

These symbols are not meant to be used outside of the file implementing the "mixin", I will refactor this to use proper properties instead.

Source/Fuse.Nodes/Visual.Children.LinkedList.uno
```diff
@@ -0,0 +1,216 @@
...
+			return true;
+		}
+
+		int Children_IndexOf(Node n)
+		{
+			var k = Children_first;
+			int i = 0;
+			while (k != null)
+			{
+				if (k == n) return i;
```
Duckers: Why would a reference comparison be slower?

Comparing the ID would be faster, but only `Visual`s have those, not all nodes.

Source/Fuse.Nodes/Visual.Children.LinkedList.uno
```diff
@@ -0,0 +1,216 @@
...
+		// if the intention is to Insert(index, elm)
+		// This means it returns null if the index == 0, and throws exception
+		// if the index was outside bounds
+		Node Children_ItemBefore(int index)
+		{
+			if (index == 0) return null;
+			else return Children_ItemAt(index-1);
+		}
+
+		void Children_Insert(int index, Node newNode)
```
Duckers: It is still needed to implement `IList<T>`. Hopefully not used much in performance critical code paths, but must be there.

Source/Fuse.Nodes/Visual.Children.uno
Duckers: That would be a separate optimziation for later. I don't want the extra code churn and added field on this branch, and the HashSet doesn't show up as significant in any profiling.

Duckers:
@mortoray I've addressed all your review points and also implemented a `SafeIterator` similar to what you suggested. As this in either case creates a new object to iterate, I simply went for implementing `IEnumerator<Node>` as a safe iterator (which performs lazy copy only if it has to).

Duckers:
Please re-review/approve.

kusma:
> Source\Fuse.Controls.Navigation\PageIndicator.uno(106.21): W4139: 'Fuse.Visual.get_LastVisualChild()' is obsolete: 'Use LastChild<Visual>() instead'

...and:

> From 31 projects, ran 911 tests, 134 failed.

Source/Fuse.Reactive.JavaScript/Tests/UX/ClassConstructor.ux
```diff
@@ -1,6 +1,5 @@
 <UX.ClassConstructorBase ux:Class="UX.ClassConstructor">
 	<Panel ux:Dependency="A" ux:Binding="A" />
 	<Panel ux:Dependency="B" /> <!--  This one is here to force this and UX.ClassConstructorBase to get different constructor signatures -->
-	<Panel ux:Name="_a" Children="A" ux:AutoBind="false" />
```
kusma: You can't do this. This is a regression-test that verifies that we don't generate invalid Uno-code in this case, changing this completely undermines the test-case.

mortoray:
@Duckers How do this avoid creating a memory leak on the nodes? A list of children references prev/next, thus creating a memory loop. How is this loop cleaned up?

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
```diff
@@ -121,21 +121,21 @@ namespace Fuse.Layouts
...
-			var majorRest = new float[elements.Count];
+			var majorRest = new float[container.Children.Count];
 			//where this row starts
 			int majorStart = 0;
 			// current row
 			int currentRow = 0;
 			
-			for (int i = 0; i < elements.Count;++i)
+			int i = 0;
+			for (var e = container.FirstChild<Visual>(); e != null; e = e.NextSibling<Visual>(), i++)
```
mortoray: The variable `i` has a different value than previously. You're only iterating `Visual` now instead of `Node`, thus `i++` is not incremented as often. You probably need to change this to `FirstChild<Node>` and `NextSibling<Node>` just to be safe.

Source/Fuse.Controls.Panels/Layouts/WrapLayout.uno
```diff
@@ -183,10 +183,9 @@ namespace Fuse.Layouts
 				var saMin = IsVert ? AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment);
 				var saMaj = IsVert ? AlignmentHelpers.GetVerticalSimpleAlignOptional(RowAlignment) : AlignmentHelpers.GetHorizontalSimpleAlignOptional(RowAlignment);
 				var elp = lp.CloneAndDerive();
-				for (int i=0; i < elements.Count; ++i)
+				i = 0;
+				for (var element = container.FirstChild<Visual>(); element != null; element = element.NextSibling<Visual>(), i++)
```
mortoray: Same here, not the same `i` value  (if this change was consistent it might work, but it feels safer to iteratre over `Node` like before).

Source/Fuse.Controls/Tests/Fuse.Controls.Test.Helpers/PanelTester.uno
```diff
@@ -122,6 +125,8 @@ namespace Fuse.Controls.Test
 				LayoutTestHelper.TestElementLayout(root, panelToTest, int2(800,600),
 					root.SnapToPixelsSize(float2(100, 71)), float2(800 - 100 - root.SnapToPixelsPos(21.4f),
 					root.SnapToPixelsPos(5.3f)));
+
+				root.Children.Remove(parent);
```
mortoray: Why did you add this to test tests?

Source/Fuse.Navigation/Navigation.Locators.uno
```diff
@@ -26,11 +26,10 @@ namespace Fuse.Navigation
 			if (t != null)
 				return t;
 				
-			for (int i = 0; i < node.Children.Count; i++)
+			for (var x = node.FirstChild<Behavior>(); x != null; x = x.NextSibling<Behavior>())
 			{
-				var c = node.Children[i] as IBaseNavigation;
-				//mean to check only behaviours (as was a distinct list before), so exclude visuals
-				if (c != null && !(c is Visual)) return c;
+				var c = x as IBaseNavigation;
```
mortoray: Why not just do `node.FirstChild<IBaseNavigation>` in the for loop, or in place of the for-loop entirely?

Source/Fuse.Nodes/Node.DataContext.uno
```diff
@@ -87,13 +87,13 @@ namespace Fuse
 				var p = np as Visual;
 				if (p != null)
 				{
-					for (var i = p.Children.Count-1; i >= 0; i--)
+					for (var dp = p.LastChild<Node>(); dp != null; dp = dp.PreviousSibling<Node>())
 					{
-						var sibdp = p.Children[i] as ISiblingDataProvider;
-						if (sibdp != null)
+						var sdp = dp as ISiblingDataProvider;
```
mortoray: Is this not a `LastChild<ISiblingDataProvider>` loop instead of using a cast here?

Source/Fuse.Nodes/Visual.Children.SafeIterator.uno
```diff
@@ -0,0 +1,171 @@
...
+				{
+					_array = _v.Children_cachedArray; // If we have a cached array, go ahead and use that
+				}
+
+				_pos++;
+
+				if (_array != null)
+					return _pos < _array.Length;
+				
+				if (_current == null) _current = _v._firstChild;
```
mortoray: I think this is technically wrong. If you call `MoveNext` two time as the end of a list this will result in it returning to the start of the list. It should instead continue to return `false` if you keep iterating.

Source/Fuse.Nodes/Visual.Children.SafeIterator.uno
```diff
@@ -0,0 +1,171 @@
...
+				else _current = _current._nextSibling;
+
+				return _current != null;
+			}
+
+			public void Dispose()
```
mortoray: And now you hit the Uno problem: https://github.com/fusetools/uno/issues/1148
Basically you can't rely on `Dispose` being called. I hit that prolem when trying to optimize an iterator as well.

Source/Fuse.Nodes/Visual.Children.SafeIterator.uno
```diff
@@ -0,0 +1,171 @@
...
+						// the copied array is just a subset, so reset index
+						if (_pos != -1 && _array.Length > 0) _pos = 0; 
+
+						// tempting, but not correct. when _pos != -1, it uses _array[_pos]
+						// _current = _array[0]; 
+					}
+				}
+
+				if (_nextIterator != null) _nextIterator.SecureCopy();
+			}
```
mortoray: Once you've done a `SecureCopy` you can remove `this` from the `_safeIterator` list. Since it has it's own copy there's no longer a need to track it as being an iterator.

Source/Fuse.Nodes/Visual.Children.SafeIterator.uno
```diff
@@ -0,0 +1,171 @@
...
+				}
+			}
+
+			bool MultipleIterators { get { return _v._safeIterator._nextIterator != null; } }
+
+			internal void SecureCopy()
+			{
+				if (_array == null)
+				{
+					if (MultipleIterators)
```
mortoray: Perhaps check if there is a cachedArray instead, since there might be one even without multiple iterators.

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -18,35 +18,59 @@ namespace Fuse
...
-				if (Children[i] is T)
-					Children.RemoveAt(i);
+				var v = c as T;
+				if (v != null) return v;
+				c = c._previousSibling; 
+			}
+			return null;
+		}
 
-				i = Math.Min(i, Children.Count) - 1;
```
mortoray: The reason the loop iterating backwards before, and did this check, was because removing one node could potentially cause additional nodes to be unrooted -- such as unrooting a behaviour that added children...

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -18,35 +18,59 @@ namespace Fuse
...
+			
+			To remove all children (of all types), use `RemoveAllChildren<Node>()`.
+		*/
+		public void RemoveAllChildren<T>() where T: Node
+		{
+			var c = _firstChild;
+			while (c != null)
+			{
+				var v = c as T;
+				c = c._nextSibling; // important to do this before calling Remove!
```
mortoray: ...by changing it to this you might actually be capturing a node that is no longer part of the children list.

I don't know of an easy solution other than using a SafeIterator here.

Source/Fuse.Nodes/Visual.Children.uno
```diff
@@ -219,27 +248,24 @@ namespace Fuse
...
 		{
-			InsertNodesImpl(index, items, true);
+			InsertNodesAfterImpl(preceeder, items, true);
 		}
 		
 		
-		void InsertNodesImpl(int index, IEnumerator<Node> items, bool allowMove)
+		void InsertNodesAfterImpl(Node preceeder, IEnumerator<Node> items, bool allowMove)
 		{
-			if (index <0 || index > Children.Count)
```
mortoray: Perhaps replace this sanity check with one that checks that `preceeder` is a child of this node.  It's zero cost now I think, and it'd give a cleaner error than waiting until deep inside the loop.

Source/Fuse.Controls/Tests/Fuse.Controls.Test.Helpers/PanelTester.uno
Duckers: It is now illegal to be a child of more than more parent. This test violated that.

Source/Fuse.Navigation/Navigation.Locators.uno
Duckers: The template argument must be `T : Node`, otherwise you can't call `.NextSibling<T>()`

Source/Fuse.Nodes/Node.DataContext.uno
Duckers: The template argument must be T : Node, otherwise you can't call .NextSibling<T>()

Source/Fuse.Nodes/Visual.Children.SafeIterator.uno
Duckers: @mortend  will address this shortly

Source/Fuse.Nodes/Visual.Children.SafeIterator.uno
Duckers: Fixed in the code.

Source/Fuse.Nodes/Visual.Children.uno
Duckers: Fixed in the code.

Source/Fuse.Navigation/Navigation.Locators.uno
mortoray: The template argument doesn't need to be T to call `FirstChild<T>` though. It can be any type. It's only if you need iteration it needs to be a `Node`, but that doesn't affect `FirstChild`.  Actually, it doesn't even affect `NextSibling<T>` either. It's only when used together in a for-loop they need to be nodes, but that should just work since the for-loop knows the type.

I'm pointing this out since I know there's at least one case on master nor where `FindFirst` is expected to work with non-node types.

mortoray:
Once the tests pass and https://github.com/fusetools/uno/issues/1148 is fixed I can approve. I'd just like a chance to review any potential changes required to get the tests working.

Duckers:
@mortoray 1148 is merged to uno master now

Source/Fuse.Reactive.JavaScript/Tests/UX/ClassConstructor.ux
Duckers: Fixed now

PR #261, opened 2017-07-20
Optimizes VisualBounds.BoxTransform

Duckers:
Takes advantage of FastMatrix to avoid a ridiculous amount of float4x4 * float4 transforms of trivial matrixes.

Unrolls min/max of vectors to avoid lots of redundant compares.

Large performance gains in cases with many Children on a visual.

This PR contains:
- [x] Changelog
- [-] Documentation
- [-] Tests

### DIFF
--- a/Source/Fuse.Nodes/VisualBounds.uno
+++ b/Source/Fuse.Nodes/VisualBounds.uno
@@ -251,6 +251,32 @@ namespace Fuse
 			return BoxTransform(box, FastMatrix.FromFloat4x4(transform));
 		}
 
+		static float Min8(float a, float b, float c, float d, float e, float f, float g, float h)
+		{
+			float min = a;
+			if (b < min) min = b;
+			if (c < min) min = c;
+			if (d < min) min = d;
+			if (e < min) min = e;
+			if (f < min) min = f;
+			if (g < min) min = g;
+			if (h < min) min = h;
+			return min;
+		}
+
+		static float Max8(float a, float b, float c, float d, float e, float f, float g, float h)
+		{
+			float max = a;
+			if (b > max) max = b;
+			if (c > max) max = c;
+			if (d > max) max = d;
+			if (e > max) max = e;
+			if (f > max) max = f;
+			if (g > max) max = g;
+			if (h > max) max = h;
+			return max;
+		}
+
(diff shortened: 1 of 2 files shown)


Source/Fuse.Nodes/VisualBounds.uno
```diff
@@ -245,21 +245,79 @@ namespace Fuse
...
+			if (B.Z > maxZ) maxZ = B.Z;
+			if (C.Z > maxZ) maxZ = C.Z;
+			if (D.Z > maxZ) maxZ = D.Z;
+			if (E.Z > maxZ) maxZ = E.Z;
+			if (F.Z > maxZ) maxZ = F.Z;
+			if (G.Z > maxZ) maxZ = G.Z;
+			if (H.Z > maxZ) maxZ = H.Z;
 
-			return new Box(
-				Math.Min(Math.Min(Math.Min(Math.Min(Math.Min(Math.Min(Math.Min(A, B), C), D), E), F), G), H),
```
mortoray: It really bothers me that we can't rely on our compiler to optmize this to be the same as the above hand-coded comparisons. :(

Source/Fuse.Nodes/VisualBounds.uno
```diff
@@ -245,21 +245,79 @@ namespace Fuse
 			return "" + _box.Minimum + " " + _box.Maximum;
 		}
 		
-		//uses the W paramete runlike Box.Transform (which may be a defect there)
+		/** @deprecated Please use the other overload (for performance) */
 		public static Box BoxTransform(Box box, float4x4 transform)
```
mortoray: Add [Obsolete] tag to get compile-time warnings.

Duckers:
Fixed, please approve @mortoray

Source/Fuse.Nodes/VisualBounds.uno
kusma: Yes, indeed. Are we sure this part actually helps? If it does, please at least make Min8/Max8()-helpers to not have to repeat this so many times.

Duckers: Pretty sure. `Min` and `Max` showed up as around 10% CPU each in the profiling. After the rewrite there is no trace of the `BoxTransform` method anywhere anymore.

PR #264, opened 2017-07-21
Optmize call to InsertNodes or InsertOrMoveNodes

mortoray:
The move command is slower, so for Each users that don't require move this will avoid some of the overhead introduced by 1.2.

This PR contains:
- [-] Changelog - no public change
- [-] Documentation - no api change
- [-] Tests -- covered by existing tests


mortoray:
The failure has been fixed on multiple other branches 1.1, 1.2, and another PR. I'll wait for it to be merged then update this PR.

Issue #265, opened 2017-07-21
Animations/Actions are not being bypassed


mortoray:
The code in the issue https://github.com/fusetools/fuselibs-public/issues/242 shows the animations and Actions can be executed in the frame they are being rooted. This is unintended and a defect. `While...` triggers should skip any activity in the rooting frame (standard bypass mechanism).

I'm marking this High since it might be a regression. If not, it's a problem for future compatibility that people will suddenly find triggers not working the same when this is fixed.

PR #266, opened 2017-07-21
Fixes status update for placed elements in WhilevisibleInScrollView

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/242

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

PR #267, opened 2017-07-21
Fixed root grouping to consider deferred actions and placement

mortoray:
Fixes https://github.com/fusetools/fuselibs-public/issues/265

This PR includes the https://github.com/fusetools/fuselibs-public/pull/266 as it builds on that fix.

This PR contains:
- [x] Changelog
- [x] Documentation - minimal
- [x] Tests - incomplete, adding another one now

### DIFF
--- a/Source/Fuse.Triggers/Trigger.uno
+++ b/Source/Fuse.Triggers/Trigger.uno
@@ -16,6 +16,11 @@ namespace Fuse.Triggers
 		Never,
 		/** Only changes during the rooting frame are handled as bypass, without special exceptions. */
 		Rooting,
+		
+		/** Deprecated: 2017-07-21
+			For possible backwards compatibilty, like Standard but excludes the check for layout bypass.
+			This mode should not be used. */
+		ExceptLayout,
 	}
 
 	public enum TriggerPlayState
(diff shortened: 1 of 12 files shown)


Source/Fuse.Triggers/Trigger.uno
```diff
@@ -16,6 +16,11 @@ namespace Fuse.Triggers
 		Never,
 		/** Only changes during the rooting frame are handled as bypass, without special exceptions. */
 		Rooting,
+		
+		/** Deprecated: 2017-07-21
+			For possible backwards compatibilty, like Standard but excludes the check for layout bypass.
+			This mode should not be used. */
+		ExceptLayout,
```
kusma: I guess "[Obsolete]" won't work on single enum values?

mortend: I think `[Obsolete]` also works on single enum values.

mortoray: Hmm, it does, but then using it inside Trigger also creates warnings:

`Source/Fuse.Triggers/Trigger.uno(109.36): W4139: Fuse.Triggers.TriggerBypassMode.ExceptLayout is obsolete`

How could we avoid those?

mortoray:
I'm going to avoid merging/rebasing on master until https://github.com/fusetools/fuselibs-public/pull/266 is merged. Or I can just fold that PR into that one if we think it's okay.

Source/Fuse.Triggers/Trigger.uno
kusma: @mortoray: Ugh, yeah. How about assigning it a special value, and checking for that special value or something? Sounds a bit nasty, though...

It's not a big deal, though. We can mark this as "harder" deprecated in a follow-up if we figure out a way that doesn't suck...

Issue #268, opened 2017-07-21
Add more colors 


danhollick:
When making some of the teaching content, it's really nice to use `Color="Red"` for basic examples. It helps users see a relationship between the code and screenshots a lot better than a hex-code. But the downside is most of the examples end up being really hard to look at. Primary colors on Primary colors.
Can we add the default css colors to the default Uno colors? https://www.w3schools.com/cssref/css_colors.asp

I think they just need to be replaced here?
https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Drawing/Colors.uno

kusma:
@cbaggers: [You added](https://github.com/fusetools/Fuse/commit/2f3ec00b9e51a04abcbbab428988d911cc0a2b0d) the current color-palette back in early 2015. I know it's a lot to ask, and if you don't know, that's totally fine. But do you remember why you chose this particular seemingly arbitrary subset of the CSS color palette?

Sunjammer:
He just looked for the most unemployed, british colors he could find

jesusmartinoza:
I personally like these http://www.flatuicolorpicker.com/

yupferris:
@Sunjammer: hate to be _that guy_, but I'm not so sure some of our office in-jokes are really appropriate on our public repos.

mortoray:
If the purpose of this is just for teaching examples, then the list of names should be stripped to only those that would be understood by a majority of our readers.

In the current PR there are too many names that nobody would know, and nor I doubt any designer would ever refer to.  Things like t`Dark...` names are good, as are `Light...`, as it adding common names like `Crimson` or `NavyBlue`,  but names like `GoldenRod` and `MintCream` are just noise.

Duckers:
The main problem with colors is that everyone has their preference and its really up to the *theme* being used.

So in my opinion, colors are more appropriately added to *theme kits*, such as the Alive kit defining its own colors `Alive.Blue` etc, and Material design defining its own colors `Material.Red` : https://github.com/Duckers/Fuse.MaterialDesign/blob/master/Style/ColorPalette.ux

PR #269, opened 2017-07-21
New Colors added

LuisRodriguezLD:
Now all CSS default colors included as requested #268


mortoray:
I wonder if a large number of static instances might negatively affect build and/or load time? I think we need to answer that question before introducing so many names. I think we strip unused stuff, but that's later after build, and possibly not on debug builds.

@duckers Any idea on that?

Duckers:
Yeah I'm not too keen on adding a lot of bloat to a "plain fuse app". These colors (including their *string* names) can never be stripped away from any export, because `{Resource xxx}` must find these as a global fallback. 

Instead, let's add colors to packages that are *optionally* linked in to projects.

For example, create a new package called `Fuse.UXKits.CSSColors` and declare the colors there. They should also be declared in UX files as `ux:Global` to make them easier to maintain.

LuisRodriguezLD:
I'll give that a try. Thanks for the comments.

PR #280, opened 2017-07-24
Optimized how the ElementBatcher calculates renderbounds for larger lists

Duckers:
This PR contains:
- [x] Changelog
- [-] Documentation
- [-] Tests

### DIFF
--- a/Source/Fuse.Common/FastMatrix.uno
+++ b/Source/Fuse.Common/FastMatrix.uno
@@ -7,7 +7,7 @@ namespace Fuse
 		float4x4 _matrix;
 		public float4x4 Matrix { get { return _matrix; } }
 
-		float3 Translation { get { return _matrix.M41M42M43; } }
+		internal float3 Translation { get { return _matrix.M41M42M43; } }
 
 		bool _hasNonTranslation;
 		public bool HasNonTranslation { get { return _hasNonTranslation; } }
@@ -18,6 +18,19 @@ namespace Fuse
 			 such as a failed inversion. If the matrix is not valid then the `Matrix` values are undefined.
 		*/
 		public bool IsValid { get { return _isValid; } }
+
+		public bool IsIdentity
+		{
+			get
+			{
+				if (!_hasNonTranslation)
+					return _matrix.M41M42M43 == float3(0);
+				else
+				{
+					return _matrix == float4x4.Identity;
+				}
+			}
+		}
 
 		FastMatrix()
 		{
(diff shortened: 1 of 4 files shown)


Source/Fuse.Common/FastMatrix.uno
```diff
@@ -19,6 +19,19 @@ namespace Fuse
...
 
+		public bool IsIdentity
+		{
+			get
+			{
+				if (!_hasNonTranslation)
+					return _matrix.M41M42M43 == float3(0);
+				else
+				{
+					return _matrix == float4x4.Identity;
```
mortoray: I wonder, given how CLR isn't inlining stuff, if manually inlining this here might be helpful?

Source/Fuse.Elements/Caching/ElementBatch.uno
```diff
@@ -140,12 +140,28 @@ namespace Fuse.Elements
...
-			var rect = VisualBounds.Empty;
-			for (int i = 0; i < _elements.Count; i++)
-				rect = rect.Merge(_elements[i]._elm.CalcRenderBoundsInParentSpace());
-			return rect;
+			if (_elements.Count == 0) return VisualBounds.Empty;
+
+			var box = GetRenderBounds(_elements[0]._elm);
+			for (int i = 1; i < _elements.Count; i++)
+			{
+				var b = GetRenderBounds(_elements[i]._elm);
```
mortoray: This doesn't seem to account for a child returning Infinite or Empty for the render bounds.

Source/Fuse.Elements/Caching/ElementBatch.uno
```diff
@@ -140,12 +140,28 @@ namespace Fuse.Elements
 				elm.AbsoluteZoom)), CachingRectPadding);
 		}
 
+		Box GetRenderBounds(Element elm)
+		{
+			var t = elm.InternLocalTransformInternal;
+			return VisualBounds.BoxTransform((Box)elm.LocalRenderBounds, t);
```
mortoray: This seems to break the Infinite and Empty bounds, both of which are important at times.
A transformed 0,0,0 is not the same as empty -- it will result in bounds that are too big in many situations. Empty needs to be empty.

Source/Fuse.Elements/Caching/ElementBatch.uno
mortoray: Hmm, or has it already excluded the empty/infintie ones from the batching?

Source/Fuse.Common/FastMatrix.uno
Duckers: Premature optimization/orthogonal enhancement, doesn't show up in profiling.

Duckers:
Let's merge!

PR #282, opened 2017-07-24
Optimizes the deferred actions queue to deal better with high numbers of inserts in a single frame

Duckers:
This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Common/UpdateManager.uno
+++ b/Source/Fuse.Common/UpdateManager.uno
@@ -116,10 +116,6 @@ namespace Fuse
 			list.Insert(0,us);
 		}
 
-		/*
-			Keeps a sorted list of deferred actions. This list is processed in parallel, thus
-			the add only scans back to PhaseDeferredActionsAt (the one currently being executed).
-		*/
 		public void AddDeferredAction( Action pu, IUpdateListener ul, int priority = 0 )
 		{
 			Queue<UpdateAction> list;
--- a/Source/Fuse.Common/UpdateManager.uno
+++ b/Source/Fuse.Common/UpdateManager.uno
@@ -62,8 +62,7 @@ namespace Fuse
 	{
 		public Action action;
 		public IUpdateListener update;
-		public int priority;
-		
+
 		public void Invoke()
 		{
 			if (action != null)
@@ -73,7 +72,7 @@ namespace Fuse
 				update.Update();
 		}
 	}
-	
+
 	class Stage
 	{
 		public UpdateStage UpdateStage;
@@ -82,8 +81,18 @@ namespace Fuse
 		public List<UpdateListener> Onces = new List<UpdateListener>();
 		public List<UpdateListener> OncesPending = new List<UpdateListener>();
 		
-		public List<UpdateAction> PhaseDeferredActions = new List<UpdateAction>();
-		public int PhaseDeferredActionsAt = -1;
+		public Dictionary<int, Queue<UpdateAction>> PhaseDeferredActions = new Dictionary<int, Queue<UpdateAction>>();
+
+		public int HighestPriority
+		{
+			get
+			{
+				int max = -1;
+				foreach (var p in PhaseDeferredActions)
+					if (p.Value.Count > 0 && p.Key > max) max = p.Key;
+				return max;
+			}
+		}
 		
 		public bool HasListenersRemoved;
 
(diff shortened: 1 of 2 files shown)


Source/Fuse.Common/UpdateManager.uno
```diff
@@ -82,8 +81,18 @@ namespace Fuse
...
-		public int PhaseDeferredActionsAt = -1;
+		public Dictionary<int, Queue<UpdateAction>> PhaseDeferredActions = new Dictionary<int, Queue<UpdateAction>>();
+
+		public int HighestPriority
+		{
+			get
+			{
+				int max = -1;
+				foreach (var p in PhaseDeferredActions)
+					if (p.Value.Count > 0 && p.Key > max) max = p.Key;
```
mortoray: This looks wrong. The items with lower priority values are actually to be executed first (it's an odd history that resulted in this).

Source/Fuse.Common/UpdateManager.uno
```diff
@@ -82,8 +81,18 @@ namespace Fuse
 		public List<UpdateListener> Onces = new List<UpdateListener>();
 		public List<UpdateListener> OncesPending = new List<UpdateListener>();
 		
-		public List<UpdateAction> PhaseDeferredActions = new List<UpdateAction>();
-		public int PhaseDeferredActionsAt = -1;
+		public Dictionary<int, Queue<UpdateAction>> PhaseDeferredActions = new Dictionary<int, Queue<UpdateAction>>();
```
mortoray: I wonder if we shoudln't change the values in the `LayoutPriority` enum then to be from 0...N instead, making it clear that you shouldn't do things like +/-1 on the priority (which it looks like you can do now based on the enum values).

Source/Fuse.Common/UpdateManager.uno
mortoray: I'd hope some tests would fail with this incorrect priority (possibly only those on master though).

Source/Fuse.Common/UpdateManager.uno
Duckers: That would be an orthogonal change. I'm just in it for the 60FPS :)

mortoray: It would actually make it faster. YOu don't need to compare priroity at all, you could simply take the first one. Indeed you could drop the `Dictionary` and just use a fixed array, which should be faster than iteration each time.

Duckers: Well, technically the API allows any integer, not just the LayoutPriorities. Are you sure it is fine to make that restriction?

mortoray: `foreach` may perofrm okay on DotNet, but it's aweful on the C++ backend. It creates a lot of needless objects.

mortoray: Yes, the restriction on the API is okay. It's structured such that it expects precse levels/groupings anyway. Passing random integers would basically result in undefined behaviour in some situations.

Source/Fuse.Common/UpdateManager.uno
```diff
@@ -82,8 +85,25 @@ namespace Fuse
 		public List<UpdateListener> Onces = new List<UpdateListener>();
 		public List<UpdateListener> OncesPending = new List<UpdateListener>();
 		
-		public List<UpdateAction> PhaseDeferredActions = new List<UpdateAction>();
-		public int PhaseDeferredActionsAt = -1;
+		// This size is hardcoded for performance reasons instead of using a dictionary
+		// to avoid using a foreach in the dispatch innerloop
+		public Queue<UpdateAction>[] PhaseDeferredActions = new Queue<UpdateAction>[] {
```
mortoray: Wouldn't a loop in the constructor make more sense?

Source/Fuse.Common/UpdateManager.uno
```diff
@@ -106,25 +126,11 @@ namespace Fuse
...
-		
-		public void ResetDeferredActions()
+
+		public void AddDeferredAction( Action pu, IUpdateListener ul, LayoutPriority priority = LayoutPriority.Layout)
 		{
-			PhaseDeferredActions.Clear();
-			PhaseDeferredActionsAt = -1;
+			var queue = PhaseDeferredActions[(int)priority];
+			queue.Enqueue(new UpdateAction{ action = pu, update = ul });
 		}
```
kusma: Could we still have a compatibility-shim like this, perhaps?

```uno
[Obsolete("Use AddDeferredAction(Action, IUpdateListener, LayoutPriority ) instead")]
public void AddDeferredAction(Action pu, IUpdateListener ul, int priority,
			[CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, 
			[CallerMemberName] string memberName = "")
{
	switch (priority)
	{
	case (int)LayoutPriority.Layout:
		AddDeferredAction(pu, ul, LayoutPriority.Layout);
		break;

	case (int)LayoutPriority.Placement:
		AddDeferredAction(pu, ul, LayoutPriority.Placement);
		break;

	case (int)LayoutPriority.Later:
		AddDeferredAction(pu, ul, LayoutPriority.Later);
		break;

	case (int)LayoutPriority.Post:
		AddDeferredAction(pu, ul, LayoutPriority.Post);
		break;

	default:
		Fuse.Diagnostics.UserError("Invalid priority", this, filePath, lineNumber, memberName);
	}
}
```

kusma: This way, we make the API breakage a bit more graceful, so people can update their code on their own schedule.

We could also use some different priority here, if needed.

kusma: Hmm, but people would probably call this like:
`AddDeferredAction(foo, bar, LayoutPriority.Layout + 10)`

And since C# doesn't allow implicit casts to int for an enum-type, this won't compile any more with the move to an enum in the first place. So maybe this won't work...

Issue #262, opened 2017-07-20
Crashing app in the last commit (47e524a)


ckarmy:
In the last commit **(47e524a)** my app crash when I export to a device (Android and IOS). Work fine on the preview and in the stable version.
Sorry for don't be specific but I get this log when I export the app:

```
fuse build --target=Android --run
Uno 1.1.3 (build 5510) macOS 10.12 i386 933ea5b

Configuring (6.2 s)
Installing dependencies (0.4 s)
Compiling syntax tree (2.6 s)
Generating code and data (10.9 s)
Building Android app
2/2: Ptei.apk
Note: Some input files use or override a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
(3:31)

Build completed in 231.37 seconds

---
Installing APK on 1 device(s)
Launching activity 'tei'
Running logcat on 'e21b5b8a'
E/Zygote  (  417): MountEmulatedStorage()
E/Zygote  (  417): v2
I/libpersona(  417): KNOX_SDCARD checking this for 10222
I/libpersona(  417): KNOX_SDCARD not a persona
I/SELinux (  417): Function: selinux_compare_spd_ram, SPD-policy is existed. and_ver=SEPF_GT-I9505_5.0.1-1 ver=51
I/SELinux (  417): Function: selinux_compare_spd_ram , priority [1] , priority version is VE=SEPF_GT-I9505_5.0.1-1_0051
E/SELinux (  417): [DEBUG] get_category: variable seinfo: default sensitivity: NULL, cateogry: NULL
I/ActivityManager(  905): Start proc com.tei for activity com.tei/.tei: pid=417 uid=10222 gids={50222, 9997, 1028, 1015, 3003} abi=armeabi-v7a
I/art     (  417): Late-enabling -Xcheck:jni
D/ResourcesManager(  417): creating new AssetManager and set to /data/app/com.tei-1/base.apk
D/tei (  417): SDK: 21
I/Adreno-EGL(  417): <qeglDrvAPI_eglInitialize:410>: EGL 1.4 QUALCOMM build:  ()
I/Adreno-EGL(  417): OpenGL ES Shader Compiler Version: E031.25.03.06
I/Adreno-EGL(  417): Build Date: 01/24/15 Sat
I/Adreno-EGL(  417): Local Branch: AF11_RB1_AU15
I/Adreno-EGL(  417): Remote Branch: 
I/Adreno-EGL(  417): Local Patches: 
[...]

PR #284, opened 2017-07-25
Fixes EvaluateJS crash

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/262

I'd like to add a test case to the ManualTestPlan for EvaluateJS before merging this.

This PR contains:
- [-] Changelog - hasn't made a release
- [-] Documentation - not relevant
- [x] Tests - a manual test (no WebView possible in unit tests)

### DIFF
--- a/Source/Fuse.Controls.WebView/EvaluateJS.uno
+++ b/Source/Fuse.Controls.WebView/EvaluateJS.uno
@@ -113,7 +113,7 @@ namespace Fuse.Triggers.Actions
 			get { return _target; }
 			set { _target = value; }
 		}
-		
+
 		[UXContent,UXVerbatim]
 		public string JavaScript
 		{
@@ -140,7 +140,7 @@ namespace Fuse.Triggers.Actions
 		{
 			var webView = _target ?? target.FindByType<IWebView>();
 
-			if (webView != null && _rawSource != null)
+			if (webView != null && !string.IsNullOrEmpty(_rawSource))
 			{
 				Execute(webView);
 			}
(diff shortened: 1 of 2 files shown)


Source/Fuse.Controls.WebView/EvaluateJS.uno
```diff
@@ -105,20 +105,24 @@ namespace Fuse.Triggers.Actions
...
 
 		public event JSEventHandler Handler;
 
+		IWebView _target;
 		public IWebView WebView {
 			get { return _target; }
 			set { _target = value; }
 		}
-
+
```
kusma: Adding these tabs seems pointless.

Source/Fuse.Controls.WebView/EvaluateJS.uno
```diff
@@ -136,7 +140,7 @@ namespace Fuse.Triggers.Actions
 		{
 			var webView = _target ?? target.FindByType<IWebView>();
 
-			if (webView != null && _rawSource != "")
+			if (webView != null && _rawSource != null)
```
kusma: `if (webView != null && string.IsNullOrEmpty(_rawSource))`?

kusma:
Do we need this on release-1.2?

Source/Fuse.Controls.WebView/EvaluateJS.uno
mortoray: There's no way to see these changes while editing. I also don't make them intentionally, they just happen as I move around the code.

kusma: @mortoray: `git add -p` is your friend here.

PR #286, opened 2017-07-25
Draw Rect Visualization

yupferris:
Most of the things we draw in fuselibs are rectangles, and mostly we draw them in a fairly straightforward way. However, drawing rectangles isn't always cheap - in fact the opposite: many rectangles, particularly those drawn on top of each other, can become quite expensive!

Wouldn't it be nice if we had a way to visualize which rects are being drawn and where? Wouldn't it be nice if we could see this without having to manually instrument fuselibs or my UX code each time? And wouldn't it be _even nicer_ if all of this was available built-in with Fuselibs with tests and behind a build flag so as to not affect performance when not using it?!?

I know what you're thinking: SIGN ME UP!

With this PR, building a fuse project with `-Dfuselibs_debug_draw_rects` and `-Dfuselibs_debug_draw_rects_default_enable_capture` will enable capturing and overlaid rendering of draw rects from all of your favorite Fuse visuals! (..unless your favorites involve fallback text rendering and some other corner cases, as those were skipped if I deemed the testing to be too annoying and/or the drawing to obscure. Also we only capture rects related to drawing to the screen/backbuffer/whatever..)

For example, here's what our [hikr example app](https://github.com/fusetools/hikr) looks like with these flags:

image
image
image

This tool should come in handy while profiling user apps, so that we can more quickly identify common mistakes/pitfalls involving large amounts of overdraw, oversized element measurements, etc.

Note that currently this is only meant to be a helpful tool internally, so it hasn't been productized/prettified in any way. As such, this contains no documentation/changelog except of course the comments in the implementation. Also there's surely some cleanup that needs to be done, beyond what I've originally thought of. Don't go easy on me during review :)

Things remaining before it's ready for merge:
 - [x] De-dupe some of the draw rect capturing code. Many of these draw statements apply blocks that result in [roughly] the same vertex transformations being done, so having some helpers for capturing those and sharing some more code would be nice. I suspect some of the transformations aren't entirely correct either (though they are in the common cases according to my own manual testing and the tests included here); this should also help normalize that a bit.
 - [x] Add some way to toggle this feature at run-time as well as behind the build flag. I was originally going to suggest this as a future improvement, but it seems since our tests run from the same project, only enabling this feature via a build flag will cause it to also be enabled for all other tests, which will break rendering tests elsewhere. Implementing dynamic toggle now should kill two birds with one stone.

Closes https://github.com/fusetools/fuselibs/issues/4244.

This PR contains:
- [x] Tests


Source/Fuse.Controls.Panels/GraphicsView.ux.uno
```diff
@@ -332,10 +333,14 @@ namespace Fuse.Controls
 					}
 
 					Internal.DrawManager.PrepareDraw(_dc);
+
 					_dc.PushViewport(this);
 					_dc.PushScissor( new Recti(0, 0, size.X, size.Y) );
 					_dc.Clear(Color);
 
+					if defined(DRAW_RECTS)
```
kusma: I think `DRAW_RECTS` is too vague of a name (and in a too global of a namespace) for us to use for this. For other functionality, we've used:
- `FUSELIBS_PROFILING`
- `FUSELIBS_GLDEBUG`
- `FUSELIBS_NO_TOASTS`
- `DISABLE_IMPLICIT_GRAPHICSVIEW`
- `USE_HARFBUZZ`
- `USE_JAVASCRIPTCORE` / `USE_V8` / `USE_DUKTAPE`
- `DEBUG_V8`

Even though we haven't been very consistent here, I would prefer if we used something that at least starts with `FUSELIBS_`, so we don't accidentally trample on defines that user-code uses. But perhaps we'd want to say that this is debug-drawing somehow, as opposed to "we're drawing rectangles!"...

So maybe something like `FUSELIBS_DEBUG_DRAW_RECTS` instead?

Source/Fuse.Elements/Caching/ElementBatch.uno
```diff
@@ -289,6 +289,7 @@ namespace Fuse.Elements
 				ClipPosition: Opacity > 0 ? prev : float4(0, 0, 0, -1);
 				PixelColor: sample(tex, TexCoord, SamplerState.LinearClamp) * Opacity;
 			};
+
```
kusma: This seems pointless.

Source/Fuse.Elements/Drawing/Rectangle.uno
```diff
@@ -19,4 +19,4 @@ namespace Fuse.Elements.Internal
 			};
 		}
 	}
-}
\ No newline at end of file
+}
```
kusma: This seems pointless.

Source/Fuse.Nodes/DrawRectVisualizer.uno
```diff
@@ -0,0 +1,175 @@
...
+		transformations that result in a non-rectangular shape, we store each of the 4 vertices of the
+		rectangle. In the future it will probably make sense to compare the DrawRects for a given frame
+		to determine which ones overlap to measure actual overdraw; this should be possible with this
+		configuration, even though it won't be as convenient as comparing simple rectangles necessarily.
+
+		We also store the scissor rect that was used to draw the original rectangle so we can clip the
+		draw rect accordingly for more precise info.
+
+		TODO: We'll probably want to grab the current viewport as well, although this won't
+		necessarily clip the rect like the scissor will.
```
kusma: Does this TODO still apply? If so, what's stopping us from addressing it?

Source/Fuse.Controls.Panels/GraphicsView.ux.uno
yupferris: `FUSELIBS_DEBUG_DRAW_RECTS` sounds great :)

kusma:
```
FAILED:  AllTests.DrawRectsTest.DrawRectsTest.Scale9VideoWithMarginDrawRectIsRendered
         Exception was thrown: Test timed out after 30.092709 seconds
```
Seems the video-test doesn't work on macOS :(

yupferris:
> Seems the video-test doesn't work on macOS :(

Yeah, it timed out, so looks like we need some sort of message pump for cocoa as well. From a quick google it looks even more sketchy than on Windows. Are we able to disable tests for specific platforms, eg. disable this one on macOS but leave it enabled for Windows etc?

yupferris:
Also, do you have any suggestions for how I might be able to toggle this at run-time? It should be as simple as having some `IsEnabled` state on the visualizer/capturing code technically, but I'm not entirely sure how we should expose that - perhaps as some weird swipe or something? How have we done similar things previously?

kusma:
> Are we able to disable tests for specific platforms, eg. disable this one on macOS but leave it enabled for Windows etc?

@yupferris: Yeah, you can either just mark the method as extern with a condition, or add an `[Ignore("Needs macOS message pump", OSX)]`-attribute. I tend to prefer the latter one when it's something that can be improved on in the future.

yupferris:
> Yeah, you can either just mark the method as extern with a condition, or add an [Ignore("Needs macOS message pump", OSX)]-attribute. I tend to prefer the latter one when it's something that can be improved on in the future.

Yeah, I like the latter option too here, as we can write down semantically why it's ignored. Feels like the right thing to do there.

yupferris:
@kusma I've addressed all of your previous feedback, and I've done the refactoring I wanted to do for the capture code. I'm still not sure how to expose the runtime enable/disable (see my comment above).

yupferris:
All boxes ticked etc, let's see how it fares with the tests :)

kusma:
Failed with these errors, trying to restart the tests:
```
FAILED:  AllTests.Fuse.Elements.Test.BoxSizingTest.Explicit
         Assertion failed in 'Explicit', at /Users/travis/build/fusetools/fuselibs-public/Source/Fuse.Elements/Tests/BoxSizing.Test.uno:45
           Expected: 0
           But got : 1
FAILED:  AllTests.Fuse.Elements.Test.ImageFillTest.ResourceBasic
         Assertion failed in 'ResourceBasic', at /Users/travis/build/fusetools/fuselibs-public/Source/Fuse.Elements/Tests/ImageFill.Test.uno:35
           Expected: 0
           But got : 1
FAILED:  AllTests.Fuse.Elements.Test.ImageFillTest.ResourceReplace
         Assertion failed in 'ResourceReplace', at /Users/travis/build/fusetools/fuselibs-public/Source/Fuse.Elements/Tests/ImageFill.Test.uno:66
           Expected: 0
           But got : 1
```

yupferris:
Yeah, I caught the same errors yesterday and restarted it, didn't seem to help, but let's see if it does this time. I'd be surprised if any of the changes I made were at all intrusive to those other tests, unless reading a property to capture the rects caused some weird cache behavior that changed something or something.. :)

yupferris:
Yeah they've failed again, looking into what those tests actually do..

Source/Fuse.Nodes/Tests/DrawRectsTest/DrawRects.Test.uno
```diff
@@ -0,0 +1,434 @@
...
+	{
+		[Test]
+		public void EmptyPanelIsBlack()
+		{
+			DrawRectVisualizer.IsCaptureEnabled = true;
+
+			var p = new Panel();
+			var root = TestRootPanel.CreateWithChild(p, int2(100, 100));
+
+			using (var fb = root.CaptureDraw())
```
kusma: This needs to be:
```uno
using (var root = TestRootPanel.CreateWithChild(p, int2(100, 100)))
using (var fb = root.CaptureDraw())
{
```
i.e, you need to dispose of TestRootPanel.

PR #288, opened 2017-07-25
Rewrote ZOrder implementation for speed & glory

Duckers:
The old ZOrder implementation was very expensive. It involved keeping a sorted list up to date at all times, and a lot of code sites depended on the ZOrder API when what they really wanted was just to iterate over the list of children (independent of ZOrder).

This new implementation:
- cleans up the code base to use the correct/intended APIs instead of the ZOrder ones. 
- reimplements ZOrder completely to be a lazy-evaluated and cached list, which is only sorted if needed
- adds test for the crucial `OnZOrderInvalidated` callback which is needed by `ITreeRenderer`
- makes sure `ITreeRenderer` doesn't dispatch zorder change on every event, but deferrs until the end of frame

This PR contains:
- [x] Changelog
- [-] Documentation - no relveant public APIs introduced or changed
- [x] Tests


kusma:
This failure seems relevant:
```
FAILED:  AllTests.Fuse.Controls.Test.PanelTest.ZOrder
         Assertion failed in 'ZOrder', at /Users/travis/build/fusetools/fuselibs-public/Source/Fuse.Controls/Tests/Panel.Test.uno:308
           Expected: Fuse.Controls.Panel, Name: B2
           But got : Fuse.Controls.Panel, Name: B1
```

Duckers:
On it

Duckers:
Hm, weird @kusma , that test doesn't fail locally...

Duckers:
Not able to reproduce this locally even when running all tests. Retrigger @kusma  ?

Duckers:
Curious. The test passes locally, AND on AppVeyor, but fails on Travis..

kusma:
@Duckers: Travis is on macOS, and run AllTests.unoproj rather than the tests individually. Perhaps that sheds some light on the problem?

Source/Fuse.Elements/Element.Batching.uno
```diff
@@ -47,24 +45,32 @@ namespace Fuse.Elements
...
 				// get rid of old element batcher
 				if (_elementBatcher != null)
 					_elementBatcher = null;
 
-				for (int i = _firstNonUnderlay; i < ZOrder.Count; ++i)
-					ZOrder[i].Draw(dc);
+				for (var i = 0; i < zOrder.Length; i++) 
+				{
+					var v = zOrder[i];
+					if (v.Layer == Layer.Underlay) continue;
```
mortoray: This is technically not the same algorithm as before, if for some reason there are two underlays (already hihgly unlikely) and they are in the wrong order (unlikely) this will skip both of them, whereas before it'd draw both.  I don't know if this warrants a corruption warning here if it happens.

Source/Fuse.Elements/Element.HitTest.uno
```diff
@@ -72,9 +72,9 @@ namespace Fuse.Elements
 		{
 			if (HasVisualChildren)
 			{
-				EnsureSortedZOrder();
-				for (int i = ZOrder.Count-1; i >=0; --i)
-					ZOrder[i].HitTest(htc);
+				var zOrder = GetCachedZOrder();
+				for (var i = zOrder.Length; i --> 0; )
```
mortoray: Please don't use this syntax, it's obtuse. Use the normal `for (var i = zOrder.Length-1; i >= 0; --i)`

Source/Fuse.Nodes/Tests/Fuse.Nodes.Test/Visual.Test.uno
```diff
@@ -9,8 +10,54 @@ using FuseTest;
 
 namespace Fuse.Test
```
mortoray: I've switched most test classe to the `FuseTest` namespace (without a dot) to distinguish them from being somehow part of `Fuse`.

Source/Fuse.Nodes/Visual.HitTest.uno
```diff
@@ -60,9 +60,9 @@ namespace Fuse
 		{
 			if (HasVisualChildren)
 			{
-				EnsureSortedZOrder();
-				for (int i = ZOrder.Count-1; i >=0; --i)
-					ZOrder[i].HitTest(htc);
+				var zOrder = GetCachedZOrder();
+				for (var i = zOrder.Length; i --> 0;)
```
mortoray: Again, use the non-obtuse syntax `for (var i = zOrder.Length-1; i >= 0; --i)`

Source/Fuse.Nodes/Visual.Layout.uno
```diff
@@ -417,9 +421,9 @@ namespace Fuse
 		protected virtual float2 OnArrangeMarginBox(float2 position, LayoutParams lp)
 		{
 			var sz = float2(0);
-			for (int i=0; i < ZOrderChildCount; ++i)
+			for (var v = FirstChild<Visual>(); v != null; v = v.NextSibling<Visual>())
 			{
-				var msz = GetZOrderChild(i).ArrangeMarginBox(position, lp);
+				var msz = v.ArrangeMarginBox(position, lp);
```
mortoray: Though this change is okay I wonder if we should mention it in the CHANGELOG. It may have an effect on elements that have layout relative to each other, like `LayoutMaster` or fucntions `width` and `height`. Those features are *supposed* to not care about ordering, but maybe we missed a spot.

mortoray: Or not CHANGELOG, since it's unlikely and not supposed to happen.

Source/Fuse.Nodes/Visual.ZOrder.uno
```diff
@@ -6,219 +6,123 @@ namespace Fuse
...
-			}
-		}
-
-		/** Returns the visual child with the given index. 
-		
-			For performance reasons, avoid using this function. 
+		float _zOffset = 0;
+		/**
+			Specifics a ZOffset, higher values are in front of other nodes. Only used by certain Node's,
+			such as `Panel`. The ZLayer has priority, then ZOffset, then ZOffsetNatural.
```
mortoray: I think this comment is now wrong. I belive you've elevated ZOrder handling completely to the Visual level -- it's not longer handled by the Layout system. THat might warrant a CHANGELOG entry, though I'm not sure it would affect any user (since only LayoutControl really used multiple children before).  THe comment should be correct in any case.

Source/Fuse.Nodes/VisualBounds.uno
```diff
@@ -307,6 +308,42 @@ namespace Fuse
...
+		{
+			Render,
+			HitTest
+		}
+
+		internal static VisualBounds Merge(IEnumerable<Visual> visuals, Type bt)
+		{
+			bool hasAnyBounds = false;
+			Box box = new Box(float3(0), float3(0));
+			foreach (var elm in visuals)
```
mortoray: Can you profile this on iOS, I'm still concerend when I see `foreach` in a critical code path (this will be called a lot per frame when something is animated).

Duckers:
@kusma : Unable to reproduce it on my mac either :/

Duckers:
Hum! I was able to reproduce it now. It seems to be discrepencies in how `Array.Sort` works in .NET vs C++

Duckers:
Tests were failing because the .NET and C++ implementations of Array.Sort coped differently with items with the equal sorting value. The previous implementation of ZOrder computed the natural Z-Order before sorting to disambiguate.

Fixed in latest commit.

Source/Fuse.Nodes/Tests/Fuse.Nodes.Test/Visual.Test.uno
Duckers: I didn't change that in this PR, so that's an orthogonal thing.

Source/Fuse.Elements/Element.Batching.uno
Duckers: I don't understand what you mean here. The `DrawNonUnderlayChildren` method is not supposed to draw underlays. That's what the code means.

mortoray: Yeah, but it did before if they happened to be out-of-order. So I guess this is more of a bugfix.

kusma:
@Duckers: Could you please squash that last commit into cb02d46? You being frustrated during development isn't useful information for the future ;)

Duckers:
@kusma: done

Duckers:
@kusma: done

Duckers:
Yay! Finally green!

PR #289, opened 2017-07-25
Fixes a bug in the ElementBatch.CalcRenderBounds implementation

Duckers:


kusma:
It's a little bit saddening that this didn't cause a test to fail... :/

Issue #301, opened 2017-07-27
newly introduced warnings


kusma:
The last couple of rounds of optimizations have marked APIs as obsolete without cleaning up their usage:

> Source\Fuse.Nodes\VisualBounds.uno(157.12): W4139: 'Fuse.VisualBounds.BoxTransform(Fuse.Box,float4x4)' is obsolete: 'Please use the other overload (for performance)'
> Source\Fuse.Nodes\VisualBounds.uno(167.12): W4139: 'Fuse.VisualBounds.BoxTransform(Fuse.Box,float4x4)' is obsolete: 'Please use the other overload (for performance)'
> Source\Fuse.Controls\ContentControl.uno(24.12): W4139: 'Fuse.Visual.get_FirstVisualChild()' is obsolete: 'Use FirstChild<Visual>() instead'
> Source\Fuse.Controls\Viewport.uno(126.13): W4139: 'Fuse.Visual.GetZOrderChild(int)' is obsolete: 'Iterate over ZOrder instead'
> Source\Fuse.Controls\Viewport.uno(133.23): W4139: 'Fuse.Visual.GetZOrderChild(int)' is obsolete: 'Iterate over ZOrder instead'
> Source\Fuse.Controls.Navigation\PageIndicator.uno(106.21): W4139: 'Fuse.Visual.get_LastVisualChild()' is obsolete: 'Use LastChild<Visual>() instead'

This needs to be fixed before we can release these.

kusma:
This has been fixed.

PR #302, opened 2017-07-27
Fix obsoletions

kusma:
These fix obsoletion-warnings in #301


Source/Fuse.Controls/Viewport.uno
```diff
@@ -123,14 +123,14 @@ namespace Fuse.Elements
 			get 
 			{ 
 				if (!HasVisualChildren) return null;
-				return  GetZOrderChild(0);
+				return FirstChild<Visual>();
```
mortoray: This may not be the same as the first child in Z order if there are multiple children. However, it does seem like it'd make more sense to have the actual first child instead.  And, since I think the code prevented multiple children this is probably fine.

kusma: Accidental improvement! ;)

PR #314, opened 2017-07-31
Fix fuselibs after DateTime changes in Uno

yupferris:
Fixes fuselibs impls after https://github.com/fusetools/uno/pull/1263. Only internal changes were made; no public interfaces were harmed during the making of this PR. :)

This PR contains:
- [-] Changelog
- [-] Documentation
- [-] Tests


yupferris:
👍

PR #321, opened 2017-08-01
ManualTestingApp: do not randomize with time

kusma:
`DateTime.Now` now returns a DateTime-struct, not a ZonedDateTime as
it did before. And our DateTime-struct doesn't (yet) support the
`Milliseconds`-property.

Having this test randomize based on time serves no real purpose, we
can just randomize with some constant instead. Having different
colors per run isn't a big plus.

Another alternative would be to use the LocalDateTime-class, but the
plan is to make that one obsolete at some later point. So it doesn't
feel like it's worth the hassle just for a rather useless feature.

### DIFF
--- a/Tests/ManualTests/ManualTestingApp/PointerAndMouse/PointerAndMouse.ux.uno
+++ b/Tests/ManualTests/ManualTestingApp/PointerAndMouse/PointerAndMouse.ux.uno
@@ -11,7 +11,7 @@ public partial class PointerAndMouse
     {
         InitializeUX();
         this.Title = "Pointer & Mouse";
-        _random = new Random(DateTime.Now.Millisecond);
+        _random = new Random(1337);
     }
 
 
@@ -53,4 +53,3 @@ public partial class PointerAndMouse
         return result;
     }
 }
-


Tests/ManualTests/ManualTestingApp/PointerAndMouse/PointerAndMouse.ux.uno
```diff
@@ -53,4 +53,3 @@ public partial class PointerAndMouse
         return result;
     }
 }
-
```
yupferris: Nit: needless whitespace change

kusma:
This fixes this warning:
> PointerAndMouse/PointerAndMouse.ux.uno(14.43): E3104: Uno.DateTime does not contain a member called 'Millisecond'. Could you be missing a package reference?

yupferris:
> This fixes this warning:
*error

Tests/ManualTests/ManualTestingApp/PointerAndMouse/PointerAndMouse.ux.uno
kusma: Yeah. It's something my editor did by default, and I considered dropping it. I decided to leave it in, because this is on master where whitespace-cleanups aren't generally a problem. If it was a release-branch, I'd be more skeptical. Having two newlines at the end of a file seems just silly.

That being said, I'm fine with dropping it if people care.

yupferris: I don't think it's at all a big deal in this case, especially at the end of the file. But I'll admit there's a small amount of satisfaction in pointing out white space "errors" from you specifically :) . I think it's fine!

PR #322, opened 2017-08-01
ManualTestingApp: use ZonedDateTime.Now instead of DateTime.Now

kusma:
We've changed the Uno API for DateTime.Now from returning a
ZonedDateTime to returning a struct DateTime, which is severely
limited. This is so we can add DateTime to our public UX-visible
API without pulling in a lot of timezone-stuff for all future.

So, until DateTime is a bit more capable, we need to use
ZonedDateTime instead.

In the great future, this will probably have to get reverted.


kusma:
Turns out #321 was not quite enough. I'm not 100% sure why I was able to build ManualTestingApp locally without this, but I was. Oh well, maybe some stale libraries or something.

PR #304, opened 2017-07-27
Feature upgrade to Xamarin Mac Universal

Tapped:
By upgrading to Xamarin Mac Universal we'll support 64-bit executables on macOS. 
NOTE: Depends on https://github.com/fusetools/uno/pull/1255 and fusetools/Fuse#4052

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

PR #326, opened 2017-08-01
README: add a requirements section

kusma:
This should make it a bit easier to get things up and running on a
clean machine (or docker-image).

This PR contains:
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,6 +7,22 @@ Fuselibs is the [Uno](https://www.fusetools.com/docs/uno/uno-lang)-libraries tha
 the UI framework used in [Fuse](https://www.fusetools.com/) apps.
 
 
+## Requirements
+
+In order to use Uno / Fuselibs, the following software must be installed:
+
+### Windows
+
+* VCRedist 2010: [x86](https://www.microsoft.com/en-us/download/details.aspx?id=5555), [x64](https://www.microsoft.com/en-US/Download/confirmation.aspx?id=14632)
+* [VCRedist 2013](https://www.microsoft.com/en-gb/download/details.aspx?id=40784)
+
+### macOS
+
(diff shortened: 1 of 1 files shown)


kusma:
@Tapped: You might be interested in this, as #304 probably changes the required Mono-version.

kusma:
By the way, this was inspired by a discussion with @mdavidsen, when setting up new Docker-based TeamCity agents, as well as recently setting up a new MacBook. I isolated the Windows requirements myself, using [this Docker-container](https://github.com/kusma/fuselibs-docker) on Windows. For macOs, I just documented what I had to install. I might set up a macOS Docker-container for testing requirements in the future as well, as that turned out to be a useful hack.

kusma:
@yupferris: If you have some concrete suggestions, sure. I think as-is the list is useful, and this is geared towards developers. But I guess I could add a line saying "The following software is required to be installed to use uno / fuselibs:" or something like that. I guess we could also say that it requires Windows or macOS itself also? :P

yupferris:
Haha fair, yeah, this is fine as-is I think.

kusma:
Just to be clear: That smiley was really just meant for the last part, I wasn't trying to make fun of your feedback. I'm fine with adding a sentence, and I do agree that it looks a bit bare as-is.

kusma:
I guess I could also consolidate the VCRedist 2010-lines into one line, and give separate x86 / x64 links instead.

Something like:
* VCRedist 2010: [x86](https://www.microsoft.com/en-us/download/details.aspx?id=5555), [x64](https://www.microsoft.com/en-US/Download/confirmation.aspx?id=14632)
* [VCRedist 2013](https://www.microsoft.com/en-gb/download/details.aspx?id=40784)

I guess that would make the list a bit less "in your face".

PR #327, opened 2017-08-01
Feature ImageFill uv offset and zoom

Sunjammer:
Looking at some of the fun squishy UI in games like Splatoon last night I started thinking about how I would replicate some of the effects in Fuse, and it seems to me we lack features for doing pretty simple stuff like infinitely scrolling images (even though our imagefills are repeating textures by default) or doing image scaling within a mask like circle/rect with rounded corners.  This PR adds ScaleFactor and ScaleOrigin for zooming on an imagefill, as well as Offset for translating UVs. I don't think the added complexity to the shader should hurt, as it adds an add, a subtract and a multiply. 

UV offset is applied, then scaling around an origin. 

This PR contains:
- [x] Changelog
- [x] Documentation

### DIFF
--- a/Source/Fuse.Elements/Drawing/ImageFill.uno
+++ b/Source/Fuse.Elements/Drawing/ImageFill.uno
@@ -117,6 +117,46 @@ namespace Fuse.Drawing
 				}
 			}
 		}
+		
+		static Selector _scaleOriginName = "ScaleOrigin";
+		float2 _scaleOrigin;
+		/**
+			A normalized offset used to adjust the position on which ScaleFactor is based.
+			
+			Default is 0.5,0.5 (center of the image)
+		*/
+		public float2 ScaleOrigin
+		{
+			get { return _scaleOrigin; }
+			set
+			{
+				if(_scaleOrigin != value)
+				{
+					_scaleOrigin = value;
+					OnPropertyChanged(_scaleOriginName);
+				}
+			}
+		}
+		
+		static Selector _scaleFactorName = "ScaleFactor";
+		float _scaleFactor = 1.0f;
+		/**
+			A nor
+
+			Useful for animating scrolling effects. 
+		*/
+		public float ScaleFactor
+		{
+			get { return _scaleFactor; }
+			set
+			{
+				if(_scaleFactor != value)
+				{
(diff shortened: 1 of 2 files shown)


Source/Fuse.Elements/Drawing/ImageFill.uno
```diff
@@ -110,7 +170,8 @@ namespace Fuse.Drawing
 			public float2 Origin, Size;
 			public float4 UVClip;
 			public Texture2D Texture;
-			public float2 TexCoordBias1, TexCoordBias2, TexCoordScale1, TexCoordScale2;
+			public float2 TexCoordBias1, TexCoordBias2, TexCoordScale1, TexCoordScale2, TexCoordOffset, ScaleOrigin;
+			public float ScaleFactor;
```
kusma: Can't these new bias / scales be baked into the pre-existing ones?

Sunjammer: Maybe, I didn't want to intrude on things I didn't entirely understand and keep the pr features as separate as I could

Sunjammer:
That just sounds like something that needs doing ;) curious about the uv flow on paths, would be very cool if it was oriented

Sunjammer:
Hoo boy, this hole runs pretty deep. Did not know we used System.Drawing stuff so much on dotnet

Source/Fuse.Elements/Drawing/ImageFill.uno
kusma: Well, without doing that, everyone who uses ImageFill has to pay for the cost of potentially being able to scale. I don't think that's very reasonable.

Sunjammer: It's not clear to me what TexCoordBias1/2 and Scale1/2 do. 

As for reasoning, the primary reason you'd want to use ImageFill for anything at all is to get properly masked non-rectangular images. Animations are common for Image but cannot easily be done on elements masked through ImageFills. It seems to me, at least, to be a pretty reasonable _need_ rather than a weird edge case.

Sunjammer:
Not going to pursue this approach, going to try another by extending SizingContainer

PR #330, opened 2017-08-02
Fuse.Text: fix problem with ICU-clashing symbols

kusma:
When using both -DUSE_HARFBUZZ and -DCOCOAPODS and some particular
pods, we end up linking to the wrong, AppStore banned symbols
instead of our bundled copy of ICU.

The approach of providing multiple copies of the same symbols, is
fragile, and some Cocoapods project re-writing ends up biting us
here. So let's get out of the "same symbols"-business, and use
ICUs built-in symbol-renaming mechanism instead.

Since we're using the system-provided copy of ICU on macOS, we
need to only do this for iPhone-builds.

The scripts to build and package ICU can be found here (sorry in
advance to people outside the fusetools organization, this repo
is currently private):

https://github.com/fusetools/uno-base/tree/build-icu

This PR contains:
- [x] Changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 # 1.2
 
+## Fuse.Text
+- Fixed an issue where the combination of `-DUSE_HARFBUZZ`, `-DCOCOAPODS` *and* certain Pods (in particular Firebase.Database has been identified) caused an app to link to symbols that the AppStore disallows.
+
 ## Each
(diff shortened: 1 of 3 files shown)


kusma:
I'm not entirely sure that this is urgent enough to warrant another 1.1-hotfix, but at the same time I don't really know how soon we'll ship 1.2. If needed, I can change the PR to point to 1.2.

PR #331, opened 2017-08-02
Implement basic marshalling between JS' Date and uno's DateTime

yupferris:
This allows us to do a few things:
 - Databind `Uno.DateTime` properties in UX to `Date` objects with automatic conversion both ways
 - Expose functions/properties for `NativeModule`s and `ScriptClass`s of type `Uno.DateTime` and be able to interact with them from JS as if they were JavaScript `Date`s
 - Use `Context.Wrap/Unwrap` with `Date`/`DateTime` just like we would other common types

This is a big step towards pushing out official native `DatePicker` and `TimePicker` components.

This PR contains:
- [x] Changelog
- [x] Tests


Source/Fuse.Scripting/Date.uno
```diff
@@ -0,0 +1,70 @@
...
+			return DateTimeConverterHelpers.ConvertDateToDateTime(date);
+		}
+
+		static DateTimeConverter()
+		{
+			Marshal.AddConverter(new DateTimeConverter());
+		}
+	}
+
+	class Date
```
Duckers: You don't need this type, an you don't need the `IConverter`. Convert directly to `DateTime` in `ThreadWorker.Mirror`

Source/Fuse.Scripting/Date.uno
```diff
@@ -0,0 +1,70 @@
+using Uno;
+using Uno.Time;
+
+namespace Fuse.Scripting
+{
+	static class DateTimeConverterHelpers
+	{
+		const long dotNetTicksInJsTick = 10000L;
+		const long unixEpochInDotNetTicks = 621355968000000000L;
```
kusma: Style-nit: We usually use upper-case first letter for constants like these.

yupferris:
@Duckers @kusma feedback addressed in latest commit

yupferris:
The tests that fail all fail due to a nullref on mac only it seems - my first guess is to assume mirroring/wrapping isn't handled the same way exactly in different script VM's. Perhaps it's related to this change?

https://github.com/fusetools/fuselibs-public/pull/331/files#diff-ae4b8f60af99e2159a6972f359a303e1L21

Admittedly I don't really understand what the original logic was trying to accomplish here; what does `external_object` mean and how is that different from any other scripting object?

Issue #274, opened 2017-07-24
Array of NVP's being converted to an object at root-level expression


mortoray:
The code that converts a list of NVP's into a single object is also triggering at a root level expression such as this:

```<GotoRoute Path=" 'four': (id: 13), 'inner': (a:1,b:2) "/>```

This unfortunately parses as an object with properties `four` and `inner` rather than a list of NVPs.

The NVP conversion should be limited to when an explicit parenthised list of given.

`Path=" 'four': (id: 13), 'inner': (a:1,b:2) "` (Array of NVP)
`Path=" ('four': (id: 13), 'inner': (a:1,b:2)) "` (Object)

`modifyRoute( path: "one", transition: "None" )` (Array of NVP -- assuming a VarArg function)
`modifyRoute( (path: "one", transition: "None") )` (Object)

mortoray:
See also the parsing issues: https://github.com/fusetools/fuselibs-public/issues/254 and https://github.com/fusetools/fuselibs-public/issues/253

Duckers:
Hmm, to me this isn't a straight forward bug. 

What about this expression producing an object that is both `IArray` and `IObject` ? this would make all parties happy

mortoray:
That will work if we agree those two `Path=` constructs are logically the same.  I think we did say this before. You can wrap an expression at the top-level in as many parens as you want and it makes no difference.

 The `modifyRoute` however must be distinct (one is varargs, one is a single argument).

Duckers:
@mortoray : I believe the above PR fixes this, but I don't have a test case. Please verify (and add a test case if applicable)

Issue #253, opened 2017-07-20
String cannot be converted to an IArray


mortoray:
`<FuseTest.DudElement ArrayValue=" 'one' " ux:Name="a"/>`

Where

`public IArray ArrayValue { get; set; }`

Results in: `E8001: There is nothing named ''one'' in this scope, and no global resources with that alias.`

Test branch: [expr_string_list](https://github.com/mortoray/fuselibs-public/tree/expr_string_list). Test Fuse.Reactive.Expressions / ExpressionFormat.StringList

Duckers:
Fixed in https://github.com/fusetools/fuselibs-public/pull/346

PR #337, opened 2017-08-03
NameValuePair as object fix

Duckers:
Fixes https://github.com/fusetools/fuselibs-public/issues/274

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Common/NameValuePair.uno
+++ b/Source/Fuse.Common/NameValuePair.uno
@@ -8,10 +8,13 @@ namespace Fuse
 		Implements `IObject`, which means the `NameValuePair` can be viewed as an object
 		with a single property.
 
+		Implements `IArray`, which means the `NameValuePair` can be viewed as an array
+		with a single element.
+
 		An `IArray` containing some `NameValuePairs` can be converted to an `IObject`
 		implementation containing all those properties using the `ObjectFromArray` method.
 	*/
-	public sealed class NameValuePair : IObject
+	public sealed class NameValuePair : IObject, IArray
 	{
 		public string Name { get; private set; }
 		public object Value { get; private set; }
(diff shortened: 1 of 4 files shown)


kusma:
How about getting a test for this?

Duckers:
@kusma : Yes, I'm assuming @mortoray has tests on the branch that prompted this fix, see #274

Source/Fuse.Reactive.Expressions/VarArgFunction.uno
```diff
@@ -10,7 +10,7 @@ namespace Fuse.Reactive
 	public abstract class VarArgFunction: Expression
 	{
 		/** Holds information about an argument to a `VarArgFunction`. */
-		protected class Argument 
+		internal class Argument
```
mortoray: This needs to be `internal protected` -- removing the `protected` is the cause of several of the test errors.

Source/Fuse.Common/NameValuePair.uno
```diff
@@ -8,10 +8,13 @@ namespace Fuse
...
 		with a single property.
 
+		Implements `IArray`, which means the `NameValuePair` can be viewed as an array
+		with a single element.
+
 		An `IArray` containing some `NameValuePairs` can be converted to an `IObject`
 		implementation containing all those properties using the `ObjectFromArray` method.
 	*/
-	public sealed class NameValuePair : IObject
+	public sealed class NameValuePair : IObject, IArray
```
mortoray: I'm not sure this is actually needed. I don't think the issue requires a single NameValuePair to behave as an array. The implementation also feels wrong as it strips away information, you just get the value, not the key.

mortoray: In particular, if a property takes an `IArray` and a NVP is provided the resulting array *must* be the NVP, not just the value.  So I think this is wrong here to make `NameValuePair` implement `IArray`.

Duckers: well, this means it will be impossible to denote an array of length 1

mortoray: Isn't this supposed to be something handled by marshalling? Couldn't any individual object, which isn't already an array, be marshallable to an array?

This isn't special with NVP here, a normal integer, string, or otherwise should be assignable to an array. There is another issue about that: https://github.com/fusetools/fuselibs-public/issues/253

mortoray:
Replaced wtih https://github.com/fusetools/fuselibs-public/pull/354

PR #338, opened 2017-08-03
README: urlencode parens in TeamCity badge

kusma:
Some Markdown parsers don't deal well with parens inside an URL, so
let's just URL-encode these (like they should have been in the first
place).


kusma:
In particular, this makes Atom's [Markdown Preview package](https://atom.io/packages/markdown-preview) render this link properly.

PR #344, opened 2017-08-07
ScrollViewPager -- simplified infinite/paged scrolling

mortoray:
Implements ScrollViewPager

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


mortoray:
@Duckers There's also a `ReachedStart`. The user could in theory use both in they want to unload data from the other end. However, my assumption is that the actual JS objects involved are not a significant size and there's no real need to discard them.

Unrooting/discarding of the actual node's is handled within the Each. There's only ever a small window of items that are rooted.

PR #346, opened 2017-08-08
Fixes several defects with UX string literals, arrays and expression serialization

Duckers:
Fixes several defects reported by @mortoray 

This PR contains:
- [x] Changelog
- [x] Tests

Needs new Uno stuff from https://github.com/fusetools/uno/pull/1281 before it will pass


kusma:
But not to the tests:

> Source\Fuse.Reactive.Expressions\Tests\UX\ExpressionFormat.StringList.ux(2): E8001: There is nothing named ''one'' in this scope, and no global resources with that alias. 
> Source\Fuse.Reactive.Expressions\Tests\UX\ExpressionFormat.StringList.ux(3): E8001: There is nothing named ''one', 'two'' in this scope, and no global resources with that alias. 
> Source\Fuse.Reactive.Expressions\Tests\Fuse.Expressions.Test.unoproj: E0000: Object reference not set to an instance of an object.

PR #358, opened 2017-08-11
Wrap JS Date time value when marshalling to DateTime

yupferris:
Previously, this code would assume the result of `getTime` on a JS `Date` object would be a `double`. However, one critical value does not satisfy this assumption: the Unix epoch. This is because the result is 0, which is representable as an integer.

The fix is to wrap the value from `getTime`. We can assume the wrapped value is a `double`, so we cast to that from the resulting object, then cast its value to `long`.

I was sure I had tested this initially, and I probably did when writing the code, but since I wrote tests after, that case was likely not covered when adjusting the casting logic a bit while cleaning up. Whoops!

This PR contains:
- [x] Tests

### DIFF
--- a/Source/Fuse.Reactive.JavaScript/Tests/DateMarshalTest.uno
+++ b/Source/Fuse.Reactive.JavaScript/Tests/DateMarshalTest.uno
@@ -60,12 +60,6 @@ namespace Fuse.Reactive.Test
 			{
 				e.RoundTripComponent.CallReadDate.Perform();
 				root.StepFrameJS();
-
-				// Expected date/time (UTC): November 1 1997 10:15:00
-				const long expectedTicks = 630139761000000000L;
-				var expectedDateTime = new DateTime(expectedTicks, DateTimeKind.Utc);
-
-				Assert.AreEqual(e.DateTimeComponent.TheDateTime, expectedDateTime);
 			}
 		}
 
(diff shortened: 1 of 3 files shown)


yupferris:
Yeah, that's due to me removing a part of one test that was redundant as well (as that exact same thing is covered in isolation in the following test). Without that we get a cleaner diff of the test file:

```diff
index e88f82f..0683065 100644
--- a/Source/Fuse.Reactive.JavaScript/Tests/DateMarshalTest.uno
+++ b/Source/Fuse.Reactive.JavaScript/Tests/DateMarshalTest.uno
@@ -86,6 +86,33 @@ namespace Fuse.Reactive.Test
                }

                [Test]
+               public void JSRoundTripUnixEpochDateMatches()
+               {
+                       var e = new UX.DateMarshal.UnixEpochDateRoundTrip();
+                       using (var root = TestRootPanel.CreateWithChild(e))
+                       {
+                               e.RoundTripComponent.CallReadDate.Perform();
+                               root.StepFrameJS();
+                       }
+               }
+
+               [Test]
+               public void JSRoundTripUnixEpochDateTimeMatches()
+               {
+                       var e = new UX.DateMarshal.UnixEpochDateRoundTrip();
+                       using (var root = TestRootPanel.CreateWithChild(e))
+                       {
+                               root.StepFrameJS();
+
+                               // Expected date/time (UTC): January 1 1970 00:00:00 (Unix Epoch)
+                               const long expectedTicks = 621355968000000000L;
+                               var expectedDateTime = new DateTime(expectedTicks, DateTimeKind.Utc);
+
+                               Assert.AreEqual(e.DateTimeComponent.TheDateTime, expectedDateTime);
+                       }
+               }
+
+               [Test]
                public void NativeModuleDateRoundTripAddsAnHour()
                {
                        var e = new UX.DateMarshal.NativeModuleDateRoundTrip();

```

PR #363, opened 2017-08-11
travis: build manual testing app

kusma:
So far, we haven't been building manual testing app on pull-requests, because we had issues with too long build times. Now that this has been solved, we should be able to do this.

Note that building ManualTestingApp doesn't actually work right now, due to a uno-bug. But another uno-bug cancels that one out, so uno doesn't report an error. Both of these bugs are in the process of being fixed in uno master.

This PR contains:
- [x] Tests

### DIFF
--- a/.travis.xcconfig
+++ b/.travis.xcconfig
@@ -1,0 +1,1 @@
+SDKROOT = iphonesimulator
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,10 @@ compiler: clang
 
 cache: ccache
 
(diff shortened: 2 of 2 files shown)


mortoray:
At the end of the build we get this error `/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found`  Is this related to this PR?

Should we merge despite the other pending issues you mention?

kusma:
@mortoray: This is fixed on master now, so I think this is ready to merge.

PR #371, opened 2017-08-14
Upgrade uno

kusma:
There's some incompatible / obosolete changes in Uno, making this update not entirely trivial. So here's a separate PR to fix that.


kusma:
@mortoray: Indeed. I was really just applying the changes that the uno-compiler produced, but I was also slightly involved in the PR on Uno that did this change, so I was kind of aware that this was coming ;)

yupferris:
Other than the `InternalsVisibleTo` that was missing, is there anything that's actually incompatible, or just deprecated stuff we're changing in here for later?

kusma:
@yupferris: The rest is just about removing new warnings.

yupferris:
gotcha

PR #373, opened 2017-08-14
Make Fuse.Marshal.TryConvertTo able to convert to self

yupferris:
Add support to the marshal for supporting conversion to whatever type the object already is, and add a small optimization for reference types that can skip many of the value type-specific typechecks.


yupferris:
@mortoray feedback addressed

yupferris:
:D

PR #378, opened 2017-08-16
Cleanup caching-rects in batcher

kusma:
This is just cleanups that makes this code IMO much more sane.


kusma:
@mortoray: there's no failing case, this just cleans up the fix from the release branch,which has a test case.

yupferris:
oh whoops, didn't notice it was already merged :)

PR #390, opened 2017-08-23
GL pack/unpack alignment

kusma:
We need to make sure that unpack/pack alignments are set to 1 when uploading / downloading images to GL, otherwise corruption can occur.

Seems we we've been lucky and had one of the code-sites that sets these ran before every upload, but I don't think this is guaranteed.

Fixes fusetools/fuselibs#3888.


yupferris:
Good call :)

PR #402, opened 2017-08-28
iOS: Remove SetActionStyle on MultiLineTextEdit

nekronos:
ActionStyle is not a part of the TextView public API. This was here due to legacy code before the split into TextInput and TextView

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Native/iOS/TextEdit.uno
+++ b/Source/Fuse.Controls.Native/iOS/TextEdit.uno
@@ -652,18 +652,7 @@ namespace Fuse.Controls.Native.iOS
 
 		TextInputActionStyle ITextEdit.ActionStyle
 		{
-			set
-			{
-				switch(value)
-				{
-					case TextInputActionStyle.Default: SetActionStyle(Handle, extern<int>"UIReturnKeyDefault"); break;
-					case TextInputActionStyle.Done: SetActionStyle(Handle, extern<int>"UIReturnKeyDone"); break;
-					case TextInputActionStyle.Next: SetActionStyle(Handle, extern<int>"UIReturnKeyNext"); break;
-					case TextInputActionStyle.Go: SetActionStyle(Handle, extern<int>"UIReturnKeyGo"); break;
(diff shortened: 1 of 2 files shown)


mortoray:
I know this fix is only for iOS, but the reason may not be valid. Some Android tablets offer "Action" buttons even on multiline text edits. Is this never available on iPads?

Source/Fuse.Controls.Native/iOS/TextEdit.uno
```diff
@@ -652,18 +652,7 @@ namespace Fuse.Controls.Native.iOS
...
-					case TextInputActionStyle.Default: SetActionStyle(Handle, extern<int>"UIReturnKeyDefault"); break;
-					case TextInputActionStyle.Done: SetActionStyle(Handle, extern<int>"UIReturnKeyDone"); break;
-					case TextInputActionStyle.Next: SetActionStyle(Handle, extern<int>"UIReturnKeyNext"); break;
-					case TextInputActionStyle.Go: SetActionStyle(Handle, extern<int>"UIReturnKeyGo"); break;
-					case TextInputActionStyle.Search: SetActionStyle(Handle, extern<int>"UIReturnKeySearch"); break;
-					case TextInputActionStyle.Send: SetActionStyle(Handle, extern<int>"UIReturnKeySend"); break;
-				}
-			}
+			set { /* Does not apply to MultilineTextEdit */ }
 		}
```
kusma: Keeping this feels a bit silly, as a setter which does nothing is kinda unexpected. Can't we remove this from the ITextEdit, and just check if the ITextEdit is `SingleLineTextEdit`, and call the method directly instead?

nekronos: Have to keep the setter around due to legacy reasons. This will be fixed when I get around to split the `ITextEdit` interface into `ITextInput` and `ITextView`. But that must be done in a separate PR as it would involve a lot more code changes

Issue #405, opened 2017-08-29
example-docs pages-using-js fails to locate router


mortoray:
`example-docs/examples/pages-using-js/code` the `WhileCanGoBack` and `GoBack` nodes no longer locate a Router and thus don't work. This did work in release-1.2, and logically it seems valid.

mortoray:
Actually, it might be unclear, since the `Navigator` is competing with the `Router`.  It might actually be an invalid example that happened to work before.

mortoray:
Looks like https://github.com/fusetools/fuselibs-public/commit/10e6b50f50b4c64711f64cefcefe5daee7f248e2#diff-38fbecb6487c32fe7bbeeba84109b27b actaully changed the behaviour.

PR #406, opened 2017-08-29
Fixes a IBaseNavigation location defect

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/405

This PR contains:
- [-] Changelog - fixes an unreleased regression
- [-] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Navigation/Navigation.Locators.uno
+++ b/Source/Fuse.Navigation/Navigation.Locators.uno
@@ -26,10 +26,11 @@ namespace Fuse.Navigation
 			if (t != null)
 				return t;
 				
-			for (var x = node.FirstChild<Behavior>(); x != null; x = x.NextSibling<Behavior>())
+			for (var x = node.FirstChild<Node>(); x != null; x = x.NextSibling<Node>())
 			{
 				var c = x as IBaseNavigation;
-				if (c != null) return c;
+				//we excluded Visual's before, so continue to do so (this logical prevents a trigger from matching a sibling Visual navigation control
+				if (c != null && !(c is Visual)) return c;
 			}
 			return null;
 		}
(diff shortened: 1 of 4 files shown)


Source/Fuse.Navigation/Navigation.Locators.uno
```diff
@@ -26,10 +26,11 @@ namespace Fuse.Navigation
 			if (t != null)
 				return t;
 				
-			for (var x = node.FirstChild<Behavior>(); x != null; x = x.NextSibling<Behavior>())
+			for (var x = node.FirstChild<Node>(); x != null; x = x.NextSibling<Node>())
 			{
 				var c = x as IBaseNavigation;
-				if (c != null) return c;
+				//we excluded Visual's before, so continue to do so (this logical prevents a trigger from matching a sibling Visual navigation control
+				if (c != null && !(c is Visual)) return c;
```
kusma: We excluded *everything* that wasn't Behaviors before. Are you sure this change is safe? it seems a bit dodgy to me, but I haven't thought carefully about this...

mortoray: The comment was wrong. We actually only excluded Visual's previously (which is also questionable, but two examples depend on that logic).  See https://github.com/fusetools/fuselibs-public/commit/10e6b50f50b4c64711f64cefcefe5daee7f248e2#diff-38fbecb6487c32fe7bbeeba84109b27b

kusma: Well, that had the comment saying "mean to check only behaviours (as was a distinct list before), so exclude visuals". So it's unclear to me why we want to change away from this... But I guess you've thought this properly through.

mortoray: Logic is wacky, but hard to change since it breaks examples. Duckers change is what actually broke the examples, I'm basically just reverting his change back to the wackiness that was before.

kusma: @mortoray: Right. Thanks for clarifying :)

PR #412, opened 2017-08-30
Base Route history on individual controls, as Model requires

mortoray:
This includes cleanup of several bits of the routing and navigation. This is all primarily preparation work to complete the Model based routing.

This PR contains:
- [x] Changelog - though very few user visible changes for the signigicantly changed bits
- [-] Documentation - no new APIs
- [x] Tests


Source/Fuse.Navigation/IRouterOutlet.uno
```diff
@@ -34,37 +35,70 @@ namespace Fuse.Navigation
 		/** The path/parameter was not valid and thus no change occurred */
 		Invalid,
 	}
+
+	class RouterPage
+	{
+		public string Path;
+		public string Parameter;
+		//[WeakReference] //UNO: this is causing segfaults There is a memory loop without this WeakReference
```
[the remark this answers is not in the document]
mortoray: Ah right, forgot about that. I'll double-check it still happens and add an issue.

kusma:
These warnings seems to be new:
> Source\Fuse.Navigation\Router.uno(184.10): W0000: The variable 'current' is assigned but its value is never used
> Source\Fuse.Navigation\Router.uno(308.8): W0000: The variable 'c' is assigned but its value is never used
> Source\Fuse.Controls.Navigation\Navigator.uno(348.10): W0000: The variable 'ignore' is declared but never used

mortoray:
@kusma Cleaned up the warnings, and code bits. Rebased to master. Updated changelog.

Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
```diff
@@ -95,13 +95,13 @@ namespace Fuse.Controls
 				
 			var trans = NavigationGotoMode.Transition;
 			
-			Visual v;
-			(this as IRouterOutlet).Goto( ref path, ref param, trans, op, "", out v );
-			if (v != null)
+			RouterPage rPage = new RouterPage{ Path = path, Parameter = param };
+			(this as IRouterOutlet).Goto( rPage, trans, op, "" );
```
kusma: I know that the code already did this, so just take this as a thought along the road:

This should use a real cast instead of `as`. If the cast would fail, the cast would return null, which would give us a NRE. An InvalidCastException would be much more helpful in such a case. A real cast conveys the information that this isn't supposed to fail much clearer than `as` does.

kusma: But reading the code further, how *is* `NavigationControl` a `IRouterOutlet`? I can't see either it nor any of it's base-classes implementing `IRouterOutlet`... What am I missing?

mortoray: Yeah, I just never thought of this, copying what the code already did. It seems weird as a whole that I need to cast at all to call functions on `this`, but oh well.

mortoray: Both of it's derived classed implement `IRouterOutlet`.

It appears to be a C# (not necessarily Uno) problem that it's difficult to get this interface on the base class when you have derived classes implementing it.

kusma: OK, so we depend on derived classes to implement IRouterOutlet?

mortoray: Yes.

Source/Fuse.Controls.Navigation/NavigationControl.uno
```diff
@@ -255,13 +303,10 @@ namespace Fuse.Controls
 			}
 		}
 		
-		static readonly PropertyHandle _pageDataProperty = Fuse.Properties.CreateHandle();
-	
 		//UNO: should be protected but the compiler says it isn't accessible for protected functions then
-		public class PageData
+		public class ControlPageData
```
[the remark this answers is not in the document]
mortoray: Let me figure out if I can make it private instead. It's not meanat to be public.

PR #415, opened 2017-08-31
PageControl.Pages

mortoray:
Adds the `PageControl.Pages` feature

Blocked/PR required for this merge: https://github.com/fusetools/fuselibs-public/pull/412
I will likely rebase once that one is merged

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
+++ b/Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
@@ -6,8 +6,8 @@ namespace Fuse.Controls
 	public partial class NavigationControl
 	{
 		//TODO: Change to IObservableArray once Model feature is merged
-		IArray _pages;
-		Uno.IDisposable _pagesSubscription;
+		IArray _pageHistory;
+		Uno.IDisposable _pageHistorySubscription;
 		/**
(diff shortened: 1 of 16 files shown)


Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
```diff
@@ -6,43 +6,43 @@ namespace Fuse.Controls
...
+		Uno.IDisposable _pageHistorySubscription;
 		/**
 			Pages is a stack of pages that controls the local history for a NavigationControl.
 			
 			It should be bound to a JavaScript observable array. The highest index page will always be the active page for the control. As pages are added/removed from this array the navigation state will change.
 			
 			The items in the array are objects, either explicitly created or via the Model feature. They should contain the the `$path` property which specifies the path to use. The object itself will be added to the data context for the page, allowing lookups from within the object.
 		*/
-		public IArray Pages
+		public IArray PageHistory
```
kusma: This is a public API change, and should be mentioned in the changelog, at least. But perhaps we should provide some backwards compatibility here also?

mortoray: This wasn't released yet. `Pages` is only on master.

PR #419, opened 2017-08-31
ScriptPromise: Fix randomly failing test

nekronos:
They could fail due to test-setup after calling TestRootPanel.CreateWithChild(). That could end up running the JS code before the whole test was hooked up

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
+++ b/Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
@@ -61,9 +61,9 @@ namespace Fuse.Scripting.Test
 		public void StringTest()
 		{
 			var child = new UX.BasicTest();
+			child.customPanel.StringFuture = new Promise<string>("this is a string");
 			using (var root = TestRootPanel.CreateWithChild(child))
 			{
-				child.customPanel.StringFuture = new Promise<string>("this is a string");
 				root.StepFrameJS();
 				Assert.AreEqual("this is a string", child.Properties.StringResult);
 			}
@@ -73,9 +73,9 @@ namespace Fuse.Scripting.Test
 		public void NumberTest()
 		{
 			var child = new UX.BasicTest();
+			child.customPanel.NumberFuture = new Promise<double>(13.37);
 			using (var root = TestRootPanel.CreateWithChild(child))
 			{
-				child.customPanel.NumberFuture = new Promise<double>(13.37);
 				root.StepFrameJS();
 				Assert.AreEqual(13.37, child.Properties.NumberResult);
 			}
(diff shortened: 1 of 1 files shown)


Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
```diff
@@ -85,11 +85,11 @@ namespace Fuse.Scripting.Test
 		public void FailTest()
 		{
 			var child = new UX.BasicTest();
+			var p = new Promise<double>();
+			p.Reject(new Exception("fail"));
```
mortoray: Maybe you should do this inside the `using` block, to better simulate when it might happen?

Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
```diff
@@ -99,12 +99,12 @@ namespace Fuse.Scripting.Test
 		public void ObjectTest()
 		{
 			var child = new UX.BasicTest();
+			var o = new TestObject() { Value = "Hello, World!" };
+			var p = new Promise<TestObject>();
+			p.Resolve(o);
```
mortoray: Here to?

Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
mortoray: I mean just the call to `Reject`, not the whole setup.

mortoray:
FAILED:  AllTests.FuseJS.Test.BundleTests.TestRead
         Assertion failed in 'TestRead', at /Users/travis/build/fusetools/fuselibs-public/Source/FuseJS/Tests/Bundle.Test.uno:41

This isn't related to your code, is it?

kusma:
@mortoray: No, that's not related. This is another spurious test, sadly.

Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
nekronos: Yep 👍

Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
nekronos: 👍

Source/Fuse.Scripting/Tests/ScriptClass/ScriptPromiseTests.uno
```diff
@@ -99,12 +99,12 @@ namespace Fuse.Scripting.Test
 		public void ObjectTest()
 		{
 			var child = new UX.BasicTest();
+			var o = new TestObject() { Value = "Hello, World!" };
+			var p = new Promise<TestObject>();
+			child.customPanel.ObjectFuture = p;
 			using (var root = TestRootPanel.CreateWithChild(child))
 			{
-				var o = new TestObject() { Value = "Hello, World!" };
```
kusma: It doesn't seem like moving `o` out of here is very useful...

PR #420, opened 2017-09-01
Optimize IsFlat to avoid traversal

mortoray:
Trying to see if duckers branch can be fixed.

This PR contains:
- [-] Changelog - no public api change
- [-] Documentation
- [x] Tests


Source/Fuse.Nodes/Visual.uno
```diff
@@ -95,7 +95,7 @@ namespace Fuse
 			UpdateIsContextEnabledCache();
 			UpdateIsVisibleCache();
 			UpdateContextSnapToPixelsCache();
-			WTIRooted();
```
nekronos: Why is this call removed? Is it called elsewhere?

Source/Fuse.Nodes/Visual.uno
```diff
@@ -139,7 +139,7 @@ namespace Fuse
 				foreach (var c in Children) c.UnrootInternal();
 			}
 
-			WTIUnrooted();
```
nekronos: Same as above

Source/Fuse.Nodes/Visual.uno
mortoray: I think was part of a revert on duckers branch. I'll double check why it was done. There is clearly still a `WTIRooted` function and it's not called anywhere else.  Maybe just a merge gone bad somewhere.

I'll add a test that either proves we don't need this or fails.

mortoray: Odd, I can't figure out which commit actually makes that change (I viewed each commit on its own, each the merge from master).

mortoray: I've fixed this and added tests to make sure it'd fail if they go missing again.

kusma:
@mortoray: Perhaps we could get a benchmark-app page that shows the performance benefit from this?

mortoray:
@kusma I don't have that app. It was profiling that @Duckers was doing on DotNet.  Or do you mean create a page in your new benchmark test to track the performance here?

kusma:
@mortoray: Yes, that's exactly what I mean. That app exists in the repo now.

mortoray:
@kusma Can we merge withtout and I create a high-priroity ticket to add a test that would cover this? It'd be covered by the other test I promised for each, simply by adding a few pointer events (which force IsFlat to be evaluated).

kusma:
@mortoray: I would kinda like to see evidence that this helps, it's kinda hard to tell without... I've multiple times seens "optimizations" that didn't actually help...

Source/Fuse.Nodes/Visual.IsFlat.uno
```diff
@@ -8,86 +9,75 @@ namespace Fuse
...
-				_areChildrenFlatCached = true;
-				return _areChildrenFlat;
-			}
+			if (_localNonFlat > 0)
+				DecrementNonFlat();
+
+			_flatRooted = false;
+
+			if (_nonFlat != 0)
+				throw new Exception(); // Should never happen
```
kusma: Please add some error message here. Unexplained exceptions are extremely unhelpful when they trigger...

Source/Fuse.Nodes/Visual.IsFlat.uno
```diff
@@ -8,86 +9,75 @@ namespace Fuse
...
 		}
 
-		bool _areChildrenFlat
+		bool _flatRooted;
+		void FlatRooted()
 		{
-			get { return HasBit(FastProperty1.AreChildrenFlat); }
-			set { SetBit(FastProperty1.AreChildrenFlat, value); }
+			if (_nonFlat != 0)
+				throw new Exception(); // Should never happen
```
kusma: Please add some error message here. Unexplained exceptions are extremely unhelpful when they trigger...

Source/Fuse.Nodes/Visual.IsFlat.uno
```diff
@@ -8,86 +9,75 @@ namespace Fuse
...
+		internal void DecrementNonFlat()
 		{
-			for (var t = FirstChild<Transform>(); t != null; t = t.NextSibling<Transform>())
-				if (!t.IsFlat) return false;
+			_nonFlat--;
+			if (Parent != null) Parent.ParentDecrementNonFlat();
+		}
 
-			return true;
+		int _localNonFlat = 0;
```
kusma: It's not clear to me what the difference between _nonFlat and _localNonFlat is (and why _localNonFlat isn't a boolean)...

mortoray:
We need the tests for this as well. I've got the branches if I decide to complete this.

PR #424, opened 2017-09-01
Cleanup tests

kusma:
Here's a bunch of cleanups of minor things that annoy me in our test-suite.

This PR contains:
- [x] Tests


kusma:
Just to try to sell this one a bit more:
After this:
- All test-sources are named the same way, e.g `Foo.Test.uno`.
- Most (if not all) test-classes are named the same way, e.g `FooTest`.
- No tests-sources have the same name any more. We for instance used to have *thee* sources called `Image.Test.uno`.

This is quite helpful when trying to find a particular test after a failure, as there's no more having to guess how the source is spelled, nor having to look through several sources.

PR #425, opened 2017-09-01
Fix circle precision

nekronos:
This PR contains:
- [x] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## ViewHandle
 - Fixed issue where Images with Mask could end up not displaying. This happend due to unnecessary invalidation of the implicit native GraphicsView in the app root. This invalidation was introduced when the Surface API was implemented for native. Invalidation is now opt-in on ViewHandle
 
+## Circle
+- Fixed issue where Circles could draw incorrect due to floating point precision
+
 ## Rectangle
 - Fixed issue where Rectangles could render incorreclty due to FP16 precision limitation.
 
--- a/Source/Fuse.Drawing.Primitives/Circle.uno
+++ b/Source/Fuse.Drawing.Primitives/Circle.uno
@@ -78,6 +78,9 @@ namespace Fuse.Drawing.Primitives
 		internal void Draw(DrawContext dc, Element visual, float radius, Brush brush,
 			Coverage cover, LimitCoverage limit, float extend, float2 center, float smoothness )
 		{
+			if (radius <= 0)
+				return;
+
 			if (_bufferVertex == null)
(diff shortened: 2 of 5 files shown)


Source/Fuse.Drawing.Primitives/Circle.uno
```diff
@@ -80,7 +80,8 @@ namespace Fuse.Drawing.Primitives
 		{
 			if (_bufferVertex == null)
 				InitBuffers();
-				
+
+
```
kusma: This part seems pointless

Source/Fuse.Drawing.Primitives/Circle.uno
```diff
@@ -96,8 +97,9 @@ namespace Fuse.Drawing.Primitives
 				VertexCount: _bufferIndex.Count();
 				
 				float2 VertexPosition: V0 * (radius + extend*2);
+				float2 VertexPositionScaled: (V0 * (radius + extend*2)) / radius;
```
kusma: I know I suggested this, but maybe `float2 VertexPositionScaled: VertexPosition / radius;` would be a bit clearer?

Source/Fuse.Drawing.Primitives/Circle.uno
```diff
@@ -96,8 +96,9 @@ namespace Fuse.Drawing.Primitives
 				VertexCount: _bufferIndex.Count();
 				
 				float2 VertexPosition: V0 * (radius + extend*2);
+				float2 VertexPositionScaled: VertexPosition / radius;
```
yupferris: What if radius is 0 ?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Circle
+- Fixed issue where Circles could draw incorrect due to floating point precision
```
yupferris: Nit: incorrect -> incorrectly

kusma:
@yupferris: It's not really about precision, it's about range. FP16 will overflow if the circle is too big otherwise.

Source/Fuse.Drawing.Primitives/Circle.uno
kusma: Good point. Perhaps earlying out in such cases would be a good move.

kusma:
Also, about a comment, that's probably a good idea. It'd be easy to remove this fix in the future in an effort to "optimize" the code...

yupferris:
> It'd be easy to remove this fix in the future in an effort to "optimize" the code...
right, that's exactly what I thought.

Issue #430, opened 2017-09-05
Double children rooting with resolved bindings


mortoray:
Bindings resolve during rooting prior to the children. Setting a binding value however could result in children being added, which since it is rooting will result in them being rooted. The rooting loop doesn't check for this, thus we get an error.

The recreation of this is kind of difficult with existing features, but quite easy with model based navigation.

PR #431, opened 2017-09-05
fixing issue with double rooting with known bindings

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/430

This PR contains:
- [-] Changelog
- [-] Documentation
- [x] Tests


mortoray:
A quick discussion with @nekronos confirms the underlying bug is more problematic than what I fixed. I'm going to make a more complete fix instead.

PR #437, opened 2017-09-11
Misc Fuse.Controls.Native fixes

nekronos:
This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Native/NativeRenderer.uno
+++ b/Source/Fuse.Controls.Native/NativeRenderer.uno
@@ -7,7 +7,7 @@ using OpenGL;
 namespace Fuse.Controls.Native
 {
 	extern(Android || iOS)
-	internal interface IViewHandleRenderer : IDisposable
+	public interface IViewHandleRenderer : IDisposable
 	{
 		void Draw(ViewHandle viewHandle, float4x4 localToClipTransform, float2 position, float2 size, float density);
 		void Invalidate();
--- a/Source/Fuse.Controls.Panels/SingleViewHost.uno
+++ b/Source/Fuse.Controls.Panels/SingleViewHost.uno
@@ -7,7 +7,7 @@ using Fuse.Drawing;
 namespace Fuse.Controls
 {
 	using Fuse.Controls.Native;
-	extern(Android || iOS) internal class SingleViewHost : Control, ITreeRenderer, IDisposable
+	extern(Android || iOS) public class SingleViewHost : Control, ITreeRenderer, IDisposable
 	{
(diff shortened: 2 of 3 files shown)


Source/Fuse.Controls.Native/NativeRenderer.uno
```diff
@@ -7,7 +7,7 @@ using OpenGL;
 namespace Fuse.Controls.Native
 {
 	extern(Android || iOS)
-	internal interface IViewHandleRenderer : IDisposable
+	public interface IViewHandleRenderer : IDisposable
```
mortoray: Why does this need to be public?

Source/Fuse.Controls.Panels/SingleViewHost.uno
```diff
@@ -52,6 +52,11 @@ namespace Fuse.Controls
 
 			void NextFrame()
 			{
+				UpdateManager.PerformNextFrame(NextNextFrame);
```
mortoray: It'd be cleaner to modify PerformNextFrame to allowing skipping an extra frame -- it already has the ability to wait an arbitrary number of frames internally (`deferFrame`)

I think a comment here would also be helpful as to why waiting two frames is necessary.

Source/Fuse.Controls.Native/NativeRenderer.uno
nekronos: Users cannot implement this interface if it is internal

mortoray: Are there docs that describe how to use this stuff? (I won't hold up the PR for this reason, just a question)

nekronos: No docs. This is for advanced users that can read the code.

nekronos:
@mortoray bump

mortoray:
@nekronos Looks good, but your editor changed a lot of whitespace in https://github.com/fusetools/fuselibs-public/pull/437/commits/2e967d4a4e9933660998920c87b903e2c9945a84
Bad setting somewhere?
Shall I accept anyway (I reviewed the actual change, which is fine)

mortoray:
@nekronos Did you still want to get this merged?  It's just the whitespace changes (lots of them) that are an issue.

kusma:
@mortoray: In general, I don't think the white-space changes are a problem on master. I generally object to changes like that on release-branches, due to the needless divergence and risk of merge-conflicts.

Ideally they should come in a separate commit, though.

PR #440, opened 2017-09-12
Handle AndroidSystemFont Default being null

cbaggers:
`Default` returns null when `_families.Count == 0`, however
`GetFallback` always passed the result of `Default` straight to
`AddRange` which on some devices (Android Oreo in this case) caused
the app to crash.

Havent added a changelog as this as a bugfix. However it may be worth one as it fixes Fuse builds for android 8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Fonts
+- Fixed bug where the default font on Android could end up being null.
+
 ## ViewHandle
 - Fixed issue where Images with Mask could end up not displaying. This happend due to unnecessary invalidation of the implicit native GraphicsView in the app root. This invalidation was introduced when the Surface API was implemented for native. Invalidation is now opt-in on ViewHandle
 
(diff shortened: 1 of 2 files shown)


nekronos:
@cbaggers good stuff :) Please add a changelog entry 🌮

nekronos:
@cbaggers bump

kusma:
I must say, this PR gives me a chilling feeling of curing the symptom rather than the decease. Why is _families.Count == 0 in the first place? Did Android Oreo move where they store their font-lists, for instance?

cbaggers:
Very much agreed. This only lets `GetFallback` handle a case that `Get` allows to happen, however it dealing with a code issue, not an actual one

kusma:
After diffing a bit, I actually think this is the correct fix; it seems that in Android O, we have full unicode-fonts, and don't need fallback-fonts any more.

I'm still not 100% sure about this, but it's enough to me to want to defer any other changes until we see some actual problems.

cbaggers:
ah ok. Holding off on ticket

PR #446, opened 2017-09-13
ViewHandle invalidation opt-in

nekronos:
This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests


mortoray:
Does this need to go onto 1.2?  PErhaps squash to one commit if cherry picking is necessary.

nekronos:
If we are doing a patch release for 1.2 we can backport it. I'll squash

Issue #213, opened 2017-07-12
Backpaddle on DataBinding needing a NameTable


Duckers:
`DataBinding` now requires a `NameTable` in order to be able to resolve `Fuse.Reactive.Name` nodes dynamically. This is a quite expensive requirement. Let's rewrite it so names are resolved UX-compile time instead, at the expense of these syntax trees not being reusable. This will generally make it easier to work with expressions in other contexts, too.

mortoray:
When resolved look at (upcoming from my branch) RouterModify.Test.uno/DynamicPath I've commented out a part of that test.

Duckers:
Fixed in the above PR. @mortoray please revisit your RouterModify thing

PR #310, opened 2017-07-29
Get rid if NameTable as depenency to DataBinding

Duckers:
Fixes https://github.com/fusetools/fuselibs-public/issues/213

Will need new Uno with https://github.com/fusetools/uno/pull/1265

This PR contains:
- [x] Changelog
- [-] Documentation - no new APIs
- [-] Tests - covered by existing tests

PR #459, opened 2017-09-15
Remove NameTable dependency from DataBinding

mortoray:
Fixes https://github.com/fusetools/fuselibs-public/issues/213

Replaces https://github.com/fusetools/fuselibs-public/pull/310

This PR contains:
- [x] Changelog
- [-] Documentation - no new APIs
- [-] Tests - covered by existing tests

### DIFF
--- a/Source/Fuse.Navigation/RouterModify.uno
+++ b/Source/Fuse.Navigation/RouterModify.uno
@@ -98,18 +98,10 @@ namespace Fuse.Navigation
 		internal enum Flags
 		{	
 			None = 0,
-			DontNeedNameTable = 1<<0,
 		}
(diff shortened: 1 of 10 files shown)


Source/Fuse.Navigation/RouterModify.uno
```diff
@@ -98,18 +98,10 @@ namespace Fuse.Navigation
 		internal enum Flags
```
nekronos: Is this in use still? Only one member and internal?

mortoray: I had removed it at first and replaced it with a `bool`, but the flags looked better, so I replaced it. We still need the two constructors, just so we can emit a deprecation warning on the one.

PR #461, opened 2017-09-15
Extend Fuse.PushNotifications

Clebal:
Modified version of Android implementation to add BigTextStyle and three new parameters (longTitle, longBody and notificationStyle). These parameters are sended from server side like title or body.

### DIFF
--- a/Source/Fuse.PushNotifications/Android/Impl.uno
+++ b/Source/Fuse.PushNotifications/Android/Impl.uno
@@ -24,6 +24,7 @@ namespace Fuse.PushNotifications
 					"android.os.Bundle",
 					"android.support.v4.app.NotificationCompat",
 					"com.fuse.PushNotifications.PushNotificationReceiver",
+					"com.fuse.PushNotifications.BigPictureStyleHttp",
 					"com.google.android.gms.gcm.GcmListenerService",
 					"com.google.android.gms.gcm.GoogleCloudMessaging",
 					"com.google.android.gms.common.ConnectionResult",
@@ -263,24 +264,39 @@ namespace Fuse.PushNotifications
 
 			Class cls = alertObj.getClass();
 
-			if (cls == String.class) {
+			if (cls == String.class)
+			{
 				String s = (String)alertObj;
-				@{SpitOutNotification(Java.Object,string,string,string,Java.Object):Call(listener, s, "", json.optString("sound"), bundle)};
-			} else {
+				@{SpitOutNotification(Java.Object,string,string,string,string,string,string,string,Java.Object):Call(listener,
+						s,
+						"",
+						json.optString("bigTitle"),
+						json.optString("bigBody"),
+						json.optString("notificationStyle"),
+						json.optString("featuredImage"),
+						json.optString("sound"),
+						bundle)};
+			}
+			else
+			{
 				JSONObject alert = (JSONObject)alertObj;
-				if (alertObj!=null) {
-					@{SpitOutNotification(Java.Object,string,string,string,Java.Object):Call(listener,
-																				   alert.optString("title"),
-																				   alert.optString("body"),
-																				   alert.optString("sound"),
-																				   bundle)};
+				if (alertObj!=null)
+				{
+					@{SpitOutNotification(Java.Object,string,string,string,string,string,string,string,Java.Object):Call(listener,
+						alert.optString("title"),
+						alert.optString("body"),
+						alert.optString("bigTitle"),
+						alert.optString("bigBody"),
+						alert.optString("notificationStyle"),
+						alert.optString("featuredImage"),
+						alert.optString("sound"),
+						bundle)};
 				}
 			}
 		@}
 
-
 		[Foreign(Language.Java)]
-		static void SpitOutNotification(Java.Object _listener, string title, string body, string sound, Java.Object _payload)
+		static void SpitOutNotification(Java.Object _listener, string title, string body, string bigTitle, string bigBody, string notificationStyle, string featuredImage, string sound, Java.Object _payload)
 		@{
 			Context context = (Context)_listener;
 			Bundle payload = (Bundle)_payload;
@@ -288,9 +304,8 @@ namespace Fuse.PushNotifications
 			intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
 			intent.setAction(PushNotificationReceiver.ACTION);
 			intent.replaceExtras(payload);
-			android.app.PendingIntent pendingIntent =
-				android.app.PendingIntent.getActivity(context, 0, intent, android.app.PendingIntent.FLAG_ONE_SHOT);
-
+			android.app.PendingIntent pendingIntent = android.app.PendingIntent.getActivity(context, 0, intent, android.app.PendingIntent.FLAG_ONE_SHOT);
+			android.app.NotificationManager notificationManager = (android.app.NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE);
 
 			NotificationCompat.Builder notificationBuilder = new NotificationCompat.Builder(context)
 				.setSmallIcon(@(Activity.Package).R.mipmap.notif)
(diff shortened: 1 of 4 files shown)


Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -299,6 +302,15 @@ namespace Fuse.PushNotifications
 				.setAutoCancel(true)
 				.setContentIntent(pendingIntent);
 
+
+				if(notificationStyle=="bigtextstyle" || notificationStyle.equals("bigtextstyle")){
+
+					notificationBuilder.setStyle(new NotificationCompat.BigTextStyle()
+									.bigText(longBody)
```
mortoray: Inconsistent indendation style. Just one tab per line.

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -299,6 +302,15 @@ namespace Fuse.PushNotifications
 				.setAutoCancel(true)
 				.setContentIntent(pendingIntent);
 
+
+				if(notificationStyle=="bigtextstyle" || notificationStyle.equals("bigtextstyle")){
```
mortoray: Use a const string.

Source/Fuse.PushNotifications/Fuse.PushNotifications.unoproj
```diff
@@ -13,11 +13,9 @@
     "Fuse.Elements",
     "Fuse.Reactive",
     "Fuse.Scripting",
-    "Fuse.Common"
-  ],
-  "Projects": [
-    "../Experimental.TextureLoader/Experimental.TextureLoader.unoproj",
-    "../Fuse.Platform/Fuse.Platform.unoproj"
+    "Fuse.Common",
```
mortoray: I'm not sure how this change is relevant to the feature?

Clebal: I'm sorry, this is something I had to change to make it works in my PC, I forgot to add it before to pull.

cbaggers:
Also Google seems to be pushing hard to use Firebase instead of GCM so we should probably get these changes over too [Fuse.Firebase](https://github.com/fuse-compound/Fuse.Firebase/tree/master/src/Firebase.Notifications.Android) too

Source/Fuse.PushNotifications/Android/Impl.uno
Clebal: I don't understand. Do you mean that I should create a constant string with the value "bigtextstyle" and then compare it with notificationStyle value?

Source/Fuse.PushNotifications/Android/Impl.uno
Clebal: True. I don't know why Github marks as changed when there's no change, probably I made a mistake.

Source/Fuse.PushNotifications/Android/Impl.uno
mortoray: Yes. Anytime a string is used more than once it should be a constant.

Clebal:
I've fixed everything you said. I hope it's correct now.

Clebal:
I've added an option to use BigPicturesStyle which include a pic, this pic can be send from backend as a URL that will be downloaded and converted to a bitmap using the ImageDownloaderTask class.

kusma:
@Clebal: Nice work. But the commit history is now a bit messy. Would you mind squashing things a bit, so we get something like this:

1. A commit adding ImageDownloaderTask.java
2. A commit with the rest, with a subject along the lines of "Fuse.PushNotifications: respect notificationStyle"

Commits goes into the permanent history, and there's no good reason to keep back-and-forth changes in there, it's just harder to wrap ones head around for future readers.

In addition, it would be nice if you made the subject of the PR a bit more to the point. "Modified Fuse.PushNotifications - Android implementation" makes it sound like it's just churn, but the contents of the PR looks nice to me ;)

Clebal:
@kusma Thanks 😄 But I would need your help because I'm totally new making these pull requests. What should I do to squashing things? I should checkout the files to first version and then commit again or it would be better if I close the pull request and open a new one with what you said?

Clebal:
@kusma Finally done 😃 Hope you agree with the changes.

kusma:
@mortoray, @cbaggers: What do you think? :)

Clebal:
I've forgotten to add the file Impl.cpp.uxl, I think it's perfect now.

Clebal:
@kusma Really? 😮 Oh man, I've not touch that, I'll fix it

Clebal:
@kusma I think it's done done 😃

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -299,6 +308,25 @@ namespace Fuse.PushNotifications
...
+				if(notificationStyle == BIGTEXTSTYLE || notificationStyle.equals(BIGTEXTSTYLE)){
+
+					notificationBuilder.setStyle(new NotificationCompat.BigTextStyle()
+						.bigText(bigBody)
+						.setBigContentTitle(bigTitle));
+
+				}else if(notificationStyle == BIGPICTURESTYLE || notificationStyle.equals(BIGPICTURESTYLE)){
+
+					notificationBuilder.setStyle(new NotificationCompat.BigPictureStyle()
+						.bigPicture(new ImageDownloaderTask().doInBackground(featured_image))
```
cbaggers: I'm a bit confused by this line. You make a ASync task (which is ace as we don't want to block) but then you call the `doInBackground` method directly rather than using `execute`.
I think that, when using a http image you will want to delay the creation of the notification until `onPostExecute`. Also as images are often local it would make sense to support those.

Maybe use uno api to download http/https paths to the apps internal storage and then this java code only has to handle local paths (and then it's fine to just block). This is also nice as we don't need to duplicate download functionality we already have.

Thoughts?

Clebal: I'm sorry, I'm not good using Uno then I've got some doubts.
What do you mean about `execute`? Is there a method for that?
How can I delay the creation of the notification?

I'm gonna investigate how to uso more Uno than Java for the download of the images and the local support. Where should the images be saved? Is there a specific folder for that as cache in build folder?

cbaggers: Hi Clebal, no worries about not knowing so much uno. You're doing grand. The issue in this case is in the Java code. `ASyncTask` is for running things on a separate thread. You instantiate it and then call `execute` on it which then starts the new thread and runs the code, you can find more details [here](https://developer.android.com/reference/android/os/AsyncTask.html). 

I was thinking that maybe its best to just assume the path is always to a local file and then user can just use `fetch` to `XMLHttpRequest` to get the file. 
Another reason for this is that every time we send a notification it would download the file again as there is no obvious place to cache the image data here.

Clebal: But if the user has to download the image using `fetch` and in `featured_image` parameter he send the path, what about if the user doesn't know the URL? 

For example, I'm using this to make a notification with BigPictureStyle that use an image from URL, and everytime a blog publish an article, I receive a notification. I don't want to cached those image, because it's just for one use.

I think that it would be better if I add the option to use the path. But URL option should me mantained (using Uno, for sure).

cbaggers: Hmm if the address is coming from the notifications then in that case we will have to use the java downloading code because when the notification is recieved the app may not be running so Uno wont be started yet.

So I guess it's back to the AsyncTask issue.

Clebal:
@cbaggers I think it's all done.
PushNotifications supports images using path (always that they are in Bundle) and I've fixed everything about AsyncTask in ImageDownloaderTask. I've renamed that class to make it more descriptive, because I've put some part of the BigPictureStyle code inside a method in that class to be called when `OnPostExecute` is called.

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -335,50 +333,53 @@ namespace Fuse.PushNotifications
...
 						int iconResourceID = com.fuse.R.get(featuredImage);
 
 						if (iconResourceID!=-1)
 						{
-							bitmap = android.graphics.BitmapFactory.decodeResource(context.getResources(), iconResourceID);
+							style.bigPicture(android.graphics.BitmapFactory.decodeResource(context.getResources(), iconResourceID));
 						}
 						else
 						{
 							String packageName = "@(Project.Name)";
```
kusma: Looks like this one is unused now...

Clebal: Why @kusma?

kusma: Hmm, seems I messed up, sorry about that.

kusma:
This seems to have broken building the tests: 
```
Source/build/Fuse.PushNotifications/1.3.0-rc1/Android/Impl.cpp.uxl(4.18): E0000: File /usr/local/Teamcity/Agents/tcagent-osx-agent02-teamcity-dyson/work/452c9bcd4266d21e/Source/build/Fuse.PushNotifications/1.3.0-rc1/Android/BigPictureStyleHttp.java does not exist
Source/build/Fuse.PushNotifications/1.3.0-rc1/Android/Impl.cpp.uxl(5.18): E0000: File /usr/local/Teamcity/Agents/tcagent-osx-agent02-teamcity-dyson/work/452c9bcd4266d21e/Source/build/Fuse.PushNotifications/1.3.0-rc1/Android/BundleFiles.java does not exist
```

I'm guessing this is due to some missing files from the project-file.

cbaggers:
Ah yes they are included in the build by `Impl.cpp.uxl` but not referenced by the unoproj. I'll fix that

kusma:
@cbaggers: I think @sebbert was already looking into it, so I guess you two should somehow coordinate ;)

Issue #382, opened 2017-08-17
Timer.delete(null) crash on Android


eksperts:
Fuse 1.2.0 build 13961 (bundled libs), macOS. Likely other Fuse versions affected too.

[Reported on community](https://www.fusetools.com/community/forums/bug_reports/data_loaded_the_application_crash_after_updating_f) with a larger reproduction.

Only a problem on Android, other targets appear to be fine (remains to be tested).

Minimal repro:
```
<App>
    <JavaScript>
    var Timer = require('FuseJS/Timer');
    setTimeout(function() {
        Timer.delete(null);
    }, 3000);
    </JavaScript>
</App>
```
Causes a fatal crash in Android app. Throws something like this in Android Studio:
<img width="1423" alt="screen shot 2017-08-17 at 13 30 25" src="https://user-images.githubusercontent.com/7188878/29408292-4e5de836-8350-11e7-85f5-02e63a1ed5ae.png">

fusebuild:
The forum thread [Data loaded, the application crash after updating Fuse 1.2.1](https://www.fusetools.com/community/forums/bug_reports/data_loaded_the_application_crash_after_updating_f) posted by Jae Woong Kang was linked to this issue.

fusebuild:
The forum thread [socket.io!](https://www.fusetools.com/community/forums/show_and_tell/socketio) posted by Bjørn-Olav Strand was linked to this issue.

fusebuild:
The forum thread [how to use socket io with fusetools?](https://www.fusetools.com/community/forums/howto_discussions/how_to_use_socket_io_with_fusetools) posted by Pasindu Dineth Peiris was linked to this issue.

eksperts:
This issue also causes socket.io library to fail hard, since their implementation apparently also ends up trying to delete a NULL timer someplace. That is 3rd party code, and not easily fixable.

I believe we should add some safeguarding here, so that this doesn't cause a fatal crash on Android.

igorls:
I've managed to fix the socket.io lib to solve the issues on android, actually the issue lies on the engine.io dependency, exactly on this file (https://github.com/socketio/engine.io/blob/master/lib/socket.js).

There are several clearTimeout call that are not protected from receiving null

Here's my fixed version based on socket.io 2.0.3 client library for anyone interested in testing, it's not uglyfied for debugging.
I've tested superficially on desktop preview and android
https://gist.github.com/igorls/5009b33230bb635590bebf364f14cc45

[...]

PR #469, opened 2017-09-19
Add JavaScriptCore support on Android

Tapped:
Add support for JSC on Android, which is the JavaScript engine used on iOS. It minimizes the inconsistency between JavaScript on iOS vs. Android. It also happens to fix a crash on Android https://github.com/fusetools/fuselibs-public/issues/382

This PR contains:
- [x] Changelog


Source/Fuse.Scripting.JavaScriptCore/Context.uno
```diff
@@ -5,6 +5,11 @@ using Uno;
 
 namespace Fuse.Scripting.JavaScriptCore
 {
+	[extern(Android) Require("Gradle.Dependency.Compile", "org.webkit:android-jsc:r174650")]
+	[extern(Android) Require("LinkLibrary", "jsc")]
+	[extern(Android) Require("IncludeDirectory", "@(PACKAGE_DIR:Path)/3rdparty/JavaScriptCore/Headers")]
+	[extern(Android) Require("LinkDirectory", "${CMAKE_CURRENT_SOURCE_DIR}/app/build/intermediates/exploded-aar/org.webkit/android-jsc/r174650/jni/${ANDROID_ABI}")]
```
Tapped: I would love to get some tips here... from all the resources I found, Gradle, CMake and NDK doesn't cooperate very well, when handling Gradle packages containing native libraries. The best approach I found was to find the library in the exploded-aar folder.

kusma:
@Tapped: I just tried using this for running test, and got this problem:
```
[350/350] Linking CXX shared library C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.JavaScript\Tests\build\Test\Android\app\.externalNativeBuild\cmake\debug\obj\armeabi-v7a\libFuseReactiveJavaScriptTest.so
FAILED: cmd.exe /C "cd . && C:\ProgramData\Uno\SDKs\AndroidSDK\ndk-bundle\toolchains\llvm\prebuilt\windows-x86_64\bin\clang++.exe  --target=armv7-none-linux-androideabi --gcc-toolchain=C:/ProgramData/Uno/SDKs/AndroidSDK/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/windows-x86_64 --sysroot=C:/ProgramData/Uno/SDKs/AndroidSDK/ndk-bundle/platforms/android-9/arch-arm -fPIC -g -DANDROID -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16 -fno-integrated-as -mthumb -Wa,--noexecstack -Wformat -Werror=format-security  -g -DANDROID -ffunction-sections -funwind-tables -fstack-protector-strong -no-canonical-prefixes -march=armv7-a -mfloat-abi=softfp -mfpu=vfpv3-d16 -fno-integrated-as -mthumb -Wa,--noexecstack -Wformat -Werror=format-security   -std=c++11 -Wno-unused-value -Wno-dangling-else -Wno-switch -Wno-invalid-offsetof -fexceptions -frtti -Os -DNDEBUG -Os -DNDEBUG  -Wl,--build-id -Wl,--warn-shared-textrel -Wl,--fatal-warnings -Wl,--fix-cortex-a8 -Wl,--no-undefined -Wl,-z,noexecstack -Qunused-arguments -Wl,-z,relro -Wl,-z,now -Wl,--build-id -Wl,--warn-shared-textrel -Wl,--fatal-warnings -Wl,--fix-cortex-a8 -Wl,--no-undefined -Wl,-z,noexecstack -Qunused-arguments -Wl,-z,relro -Wl,-z,now  -s -shared -Wl,-soname,libFuseReactiveJavaScriptTest.so -o C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.JavaScript\Tests\build\Test\Android\app\.externalNativeBuild\cmake\debug\obj\armeabi-v7a\libFuseReactiveJavaScriptTest.so @CMakeFiles/FuseReactiveJavaScriptTest.rsp  && cd ."
C:/ProgramData/Uno/SDKs/AndroidSDK/ndk-bundle/toolchains/arm-linux-androideabi-4.9/prebuilt/windows-x86_64/lib/gcc/arm-linux-androideabi/4.9.x/../../../../arm-linux-androideabi/bin\ld: error: cannot find -ljsc
```

Do you have any idea what the problem is?

kusma:
Looks kinda like the `LinkDirectory`-thing didn't make it to the command-line... :/

kusma:
If I hard-code the full path, this works for me:
```patch
diff --git a/Source/Fuse.Scripting.JavaScriptCore/Context.uno b/Source/Fuse.Scripting.JavaScriptCore/Context.uno
index e77a79eac8..7113b046a7 100644
--- a/Source/Fuse.Scripting.JavaScriptCore/Context.uno
+++ b/Source/Fuse.Scripting.JavaScriptCore/Context.uno
@@ -8,7 +8,7 @@ namespace Fuse.Scripting.JavaScriptCore
 	[extern(Android) Require("Gradle.Dependency.Compile", "org.webkit:android-jsc:r174650")]
 	[extern(Android) Require("LinkLibrary", "jsc")]
 	[extern(Android) Require("IncludeDirectory", "@(PACKAGE_DIR:Path)/3rdparty/JavaScriptCore/Headers")]
-	[extern(Android) Require("LinkDirectory", "${CMAKE_CURRENT_SOURCE_DIR}/app/build/intermediates/exploded-aar/org.webkit/android-jsc/r174650/jni/${ANDROID_ABI}")]
+	[extern(Android) Require("LinkDirectory", "C:/Users/erik/src/fuselibs/Source/Fuse.Reactive.JavaScript/Tests/build/Test/Android/app/build/intermediates/exploded-aar/org.webkit/android-jsc/r174650/jni/armeabi-v7a")]
 	[extern(iOS) Require("Xcode.Framework", "JavaScriptCore")]
 	[Require("Header.Include", "JavaScriptCore/JavaScript.h")]
 	public extern(USE_JAVASCRIPTCORE) class Context: Fuse.Scripting.Context
```

Obviously, that's not a viable solution, though. So there's something wrong with the library-path...

Tapped:
@cbaggers fixed that in https://github.com/cbaggers/fuselibs-public/commit/8f7a7575aded01f530b5b90a3a2a6a7503758117

Stuff have been moved since I made this PR, and a change in Uno affected the library path

kusma:
@Tapped: Sadly, that doesn't fix the problem for me.

kusma:
Actually, it seems this is the real problem after that change:
```
CMake Error at CMakeLists.txt:6 (link_directories):
  Syntax error in cmake code at

    C:/Users/erik/src/fuselibs/Source/Fuse.Reactive.JavaScript/Tests/build/Test/Android/app/src/main/CMakeLists.txt:6

  when parsing string

    C:\Users\erik\src\fuselibs\Source\Fuse.Reactive.JavaScript\Tests\build\Test\Android/app/build/intermediates/exploded-aar/org.webkit/android-jsc/r174650/jni/${ANDROID_ABI}

  Invalid character escape '\U'.
```
So, it seems we're mixing Windows and Unix path separators here. CMake seems to want Unix-style.

Tapped:
Yeah, that's fixed by doing `@(OutputDirectory:Path)` as following
`[extern(Android) Require("LinkDirectory", "@(OutputDirectory:Path)/app/build/intermediates/exploded-aar/org.webkit/android-jsc/r174650/jni/${ANDROID_ABI}")]`

kusma:
Yeah, that does the trick. Thanks a lot!

kusma:
@ollef: I'm a bit skeptical to the requires being moved to context, but apart from that most of the code-stuff seems OK to me. I suspect we could perhaps make requires a bit easier to maintain through an UXL-file or something, what do you think?

kusma:
I also think the last commit (allowing passing in context) should be dropped from this PR. It can be included in the PR that depends on it instead.

Source/Fuse.Scripting.JavaScript/JavaScriptCore/3rdparty/README.md
```diff
@@ -0,0 +1,3 @@
+### Update JavaScriptCore headers
+
+Currently it can only be done on macOS by replacing `Headers/JavaScriptCore` with `/System/Library/Frameworks/JavaScriptCore.framework/Headers`.
```
ollef: Do these headers necessarily match the Gradle library?

Source/Fuse.Scripting.JavaScript/JavaScriptCore/JSTypedArrayInclude.h
```diff
@@ -1,11 +1,14 @@
...
-	#define __MAC_10_12 101200
-#endif
-#ifndef __IPHONE_10_0
-	#define __IPHONE_10_0 100000
-#endif
-#if __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_10_0 || __MAC_OS_X_VERSION_MAX_ALLOWED >= __MAC_10_12
-	#include <JavaScriptCore/JSTypedArray.h>
-	#define JAVASCRIPTCORE_ARRAYBUFFER_SUPPORT
-#endif
+#ifdef ANDROID
```
ollef: If at all possible, i.e. we're using a recent enough version of JSC on Android, we should define `JAVASCRIPTCORE_ARRAYBUFFER_SUPPORT`, as otherwise some operations on byte arrays and the like are very slow.

ollef:
@kusma: I'm guessing it's inconsequential where the requires are since `Array`, `Object`, etc depend on `Context` anyway. This way avoids some repetition, so I'm for it. An UXL file would also be alright if that's preferred.

Agreed on passing in the context. That's not needed now.

Source/Fuse.Scripting.JavaScript/JavaScriptCore/JSTypedArrayInclude.h
ollef: If you're not sure, try it and see if the scripting tests pass.

kusma: Sadly, scripting-tests is not in their best shape yet with this (nor with V8). I'm working on fixing what I can on master as we speak...

ollef: Aha, I wasn't aware. [These tests](https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Scripting/Tests/Scripting.Test.uno) should be enough in this case. Are they also in bad shape?

Source/Fuse.Scripting.JavaScript/JavaScriptCore/3rdparty/JavaScriptCore.cpp.uxl
```diff
@@ -0,0 +1,11 @@
+<Extensions Backend="CPlusPlus">
+	<CopyFile HeaderFile="JavaScriptCore/JavaScript.h" />
```
ollef: Does this mean that we're copying these headers on iOS as well? Do we want that and not use the ones from the framework? For example the `JSTypedArray` header isn't present on some older versions.

kusma: I think we do, and we probably shouldn't...

Source/Fuse.Scripting.JavaScript/JavaScriptCore/JSTypedArrayInclude.h
kusma: Sadly, yes. They crash on JSC context creation after some tests :(

ollef: Ouch :(

cbaggers: We get a linker error when trying to use the arraybuffersupport from android, so I'm guessing the version we pull from gradle is old.

I think we should (maybe as a seperate task) use this git@github.com:facebook/android-jsc.git to build jscore as an aar and then use https://stackoverflow.com/a/23326397/574033 to include it in our android builds.

cbaggers:
I've remove the context passing and rebased on master. Currently the android tests fail when run on device with `-DUSE_JAVASCRIPTCORE` so there's work to do there

cbaggers:
Wow. Trying to get tests working with jscore on android is disturbing

Sometimes `TwoWayIdentity()` passes, somethings it hangs until the tests timeout and sometimes I get this:

```
FAILED:  Fuse.Scripting.JavaScript.Test.Fuse.Reactive.Test.ObservableTest.TwoWayIdentity
         Exception was thrown: FuseTest.DiagnosticException: Deprecated: `.inner().twoWayMap()` is deprecated. Use `.innerTwoWay()`, possibly combined with `.mapTwoWay()` or another two-way function
```

The key difference between the cases where it hangs and where it doesnt seems to be what other tests I have commented out :|. There is no pattern i've been able to see that tells me when I'll get that exception and when I won't

I have to guess this is thread related and my worry is this isnt going to make sense until we get through the thread safety fixes

kusma:
@cbaggers: Nah, that's not thread-safety related, really. It's just state not being cleaned up properly from a test before. I don't think there's any work in progress that will fix this, so I don't think waiting is going to help.

cbaggers:
@kusma good to know, thanks

cbaggers:
@kusma I'm stuck on this issue now. 

The tests reliably fail on `TwoWayIdentity` however it's not likely that itself as removing some of the tests just moves the failure

The failure on the Uno side is in the `ManualResetEvent`, however given that the JS side has crashed this is likely a spurious error

I think this is the last valid stacktrace from the js thread before the crash:
```
g::Fuse::Scripting::JavaScriptCore::JSObjectRef::CallAsFunction(OpaqueJSValue*, OpaqueJSContext const*, OpaqueJSValue*, uArray*, uDelegate*) Fuse.Scripting.JavaScriptCore.g.cpp:1460
g::Fuse::Scripting::JavaScriptCore::Object__CallMethod_fn(g::Fuse::Scripting::JavaScriptCore::Object*, g::Fuse::Scripting::Context*, uString*, uArray*, uObject**) Fuse.Scripting.JavaScriptCore.g.cpp:2270
g::Fuse::Scripting::Object::CallMethod(g::Fuse::Scripting::Context*, uString*, uArray*) Fuse.Scripting.Object.h:37
g::Fuse::Scripting::JavaScript::ClassInstance::CallMethod(g::Fuse::Scripting::Context*, g::Fuse::Scripting::Function*, uArray*) Fuse.Scripting.JavaScript.g.cpp:752
g::Fuse::Scripting::JavaScript::RootableScriptModule__CallModuleFunc_fn(g::Fuse::Scripting::JavaScript::RootableScriptModule*, g::Fuse::Scripting::Context*, g::Fuse::Scripting::Function*, uArray*) Fuse.Scripting.JavaScript.g.cpp:4413
g::Fuse::Scripting::ScriptModule::CallModuleFunc(g::Fuse::Scripting::Context*, g::Fuse::Scripting::Function*, uArray*) Fuse.Scripting.ScriptModule.h:77
g::Fuse::Scripting::ScriptModule__Evaluate_fn(g::Fuse::Scripting::ScriptModule*, g::Fuse::Scripting::Context*, g::Fuse::Scripting::ModuleResult*) Fuse.Scripting.g.cpp:4184
g::Fuse::Scripting::JavaScript::RootableScriptModule__Evaluate_fn(g::Fuse::Scripting::JavaScript::RootableScriptModule*, g::Fuse::Scripting::Context*, g::Fuse::Scripting::ModuleResult*) Fuse.Scripting.JavaScript.g.cpp:4426
g::Fuse::Scripting::Module::Evaluate(g::Fuse::Scripting::Context*, g::Fuse::Scripting::ModuleResult*) Fuse.Scripting.Module.h:43
g::Fuse::Scripting::Module::Evaluate1(g::Fuse::Scripting::Context*, uString*) Fuse.Scripting.g.cpp:1897
g::Fuse::Scripting::JavaScript::ModuleInstance::EvaluateModule(g::Fuse::Scripting::Context*) Fuse.Scripting.JavaScript.g.cpp:2460
```

I'm not sure about enough of the observable/test logic here to get any further. 

My branch is here https://github.com/cbaggers/fuselibs-public/tree/feature-android-javascriptcore

Please let me know if there is anything else I can provide

Source/Fuse.Scripting.JavaScript/JavaScriptCore/JSTypedArrayInclude.h
cbaggers: The problem with updating is that we don't have this option when cohabiting with RN.

cbaggers:
Since models merged[0] I am getting spurious hangs during testing. The first case I have been able to get a breakpoint on is:

```
		[Test]
		public void FailTest()
		{
			var p = new Promise<double>();
			child.customPanel.NumberFuture = p;
			using (var root = TestRootPanel.CreateWithChild(child))
			{
				p.Reject(new Exception("fail"));

				while (string.IsNullOrEmpty(child.Properties.NumberError))
					root.StepFrameJS();

				Assert.AreEqual("fail", child.Properties.NumberError);
			}
		}
```

where we get into that while loop and keep stepping until we are killed by the testrunner

[0] I think

kusma:
@cbaggers: Hmm, interesting. That looks like my attempt to make errors less spurious, and it's probably the wrong approach...

kusma:
35c1984b1ba91d63696f60ae97dadd84981bf7c7

Source/Fuse.Scripting.JavaScript/Tests/UX/ClearBindingTest.ux
```diff
@@ -8,19 +8,19 @@
...
+		module.exports = { p: p, remove: remove };
 	</JavaScript>
 
 	<FuseTest.Invoke Handler="{remove}" ux:Name="CallRemove"/>
 
 	<Each Items="{p}">
 		<Page ux:Class="UX.ClearBindingTestPage">
 			<JavaScript>
-				module.exports = { foo : "123" }
+				module.exports = { foo : "123" };
```
kusma: This shouldn't be necessary

Source/Fuse.Scripting.JavaScript/Tests/UX/ClearBindingTest.ux
```diff
@@ -8,19 +8,19 @@
 			p.removeAt(0);
 		}
 
-		module.exports = { p, remove }
+		module.exports = { p: p, remove: remove };
```
kusma: Adding the semicolon shouldn't be necessary

Source/Fuse.Scripting.JavaScript/Tests/UX/DataBinding.BehaviorDataContext.ux
```diff
@@ -16,7 +16,7 @@
 	                console.log("toggled");
 	                searchMode.value = !searchMode.value;
 	            }
-	        }
+	        };
```
kusma: This shouldn't be necessary

Source/Fuse.Scripting.JavaScript/Tests/UX/DataBinding.Failed.ux
```diff
@@ -1,19 +1,19 @@
...
 		})
 		
 		exports.fail = function() {
-			exports.text.failed("nope")
+			exports.text.failed("nope");
 		}
 
 		exports.restore = function() {
-			exports.text.value = "bye"
+			exports.text.value = "bye";
```
kusma: I don't think any of these changes should be necessary

Source/Fuse.Scripting.JavaScript/Tests/UX/ClearBindingTest.ux
cbaggers: Not necessary but less chance of getting bitten by js' auto-semicolon rules when refactored in future. Seemed like a painless improvement given that I was having to review all the code anyway

kusma: Please don't. This is not the right PR to discuss JS code-style.

cbaggers: No probs, I'll remove them :)

cbaggers:
Seeing the hanging issue on master, have made separate issue for tracking https://github.com/fusetools/fuselibs-public/issues/840

cbaggers:
@kusma I've cleaned up the semicolon sillyness. Should be ready for next round of review

Source/FuseJS/Tests/Environment.Test.uno
```diff
@@ -2,31 +2,32 @@ using Uno;
...
 	{
-		new FuseJS.Environment();
-		using (var jsRecorder = new JsTestRecorder())
-		{
-			var context = jsRecorder.Begin();
-			var iOS = defined(iOS) ? "true" : "false";
-			var android = defined(android) ? "true" : "false";
-			var mobile = defined(mobile) ? "true" : "false";
-			var designmode = defined(designmode) ? "true" : "false";
+		var test = new JSTest(EnvironmentInner);
```
kusma: This is confusing. The new helper is called `JSTest`, and this class is called `JsTest`. Notice the lower-case `s` in the latter. I think we should fix this up by renaming this class to `EnvironmentTest` instead, to match the normal filename/test-name convention here.

Source/FuseJS/Tests/Base64.Test.uno
```diff
@@ -11,16 +12,16 @@ namespace FuseJS.Test
...
 		{
-			new FuseJS.Base64();
-			using (var jsRecorder = new JsTestRecorder())
-			{
-				var context = jsRecorder.Begin();
-				var moduleResult = new FileModule(import("Base64Tests.js")).Evaluate(context, "Base64Tests");
-				if (moduleResult.Error != null)
-					throw moduleResult.Error;
+			var test = new JSTest(JavaScriptTestsInner);
+			test.WaitOnResults();
```
kusma: Looks like
```
var test = new JSTest(foo);
test.WaitOnResults();
```
is a common pattern now. How about making a helper in JSTest, along the lines of 
```uno
static void RunTest(Action<Fuse.Scripting.Context> method)
{
	var test = new JSTest(method);
	test.WaitOnResults();
}
```
...to reduce the code replication a bit?

cbaggers: True, I'll do that now

Source/FuseJS/Tests/Environment.Test.uno
cbaggers: agreed, on it

Source/Fuse.Scripting.JavaScript/JSTest.uno
```diff
@@ -0,0 +1,75 @@
...
+		public JSTest(Action<Fuse.Scripting.Context> test)
+		{
+			Fuse.Reactive.JavaScript.EnsureVMStarted();
+			_test = test;
+			Fuse.Reactive.JavaScript.Worker.Invoke(TestBodyWrapper);
+		}
+
+		void AddTestToJSContext(Fuse.Scripting.Context context)
+		{
+			var f = context.Evaluate("", "(function(obj, assert) { obj['test'] = { assert: function(exp, msg) { try { assert(Boolean(exp ? 1 : 0), msg); } catch(e) { assert(0, 'Error: ' + e); } } }; } )") as Fuse.Scripting.Function;
```
kusma: `as Fuse.Scripting.Function` is not the appropriate cast here, if it fails we're just going to crash with an NRE on the next line. Please do
```uno
var f = (Fuse.Scripting.Function)context.Evaluate("", "(function(obj, assert) { obj['test'] = { assert: function(exp, msg) { try { assert(Boolean(exp ? 1 : 0), msg); } catch(e) { assert(0, 'Error: ' + e); } } }; } )");
```
instead.

kusma: OK, seems you're just moving code here, so I guess that'd be an inappropriate time to sneak in a change like that. I'll just clean this up afterwards instead.

Source/FuseJS/Tests/Base64.Test.uno
cbaggers: done

Source/FuseJS/Tests/Environment.Test.uno
cbaggers: done

Source/Fuse.Scripting.JavaScript/JSTest.uno
cbaggers: 👍

cbaggers:
WOOO!

PR #474, opened 2017-09-19
ScriptClass: implement ScriptReadonlyProperty

nekronos:
This feature enables users to expose readonly data from a scriptclass

This PR contains:
- [x] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,8 @@
 - Allowed `GoBack` and `WhileCanGoBack` on the router to properly interact with bound observable/model `PageHistory`
 
 ## ScriptClass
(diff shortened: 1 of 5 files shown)


yupferris:
Though I must admit I'm struggling a bit to come up with a good use case for this, do you have an example?

nekronos:
I need this for exposing constants in JS.

```
<CameraView ux:Name="Camera">
	<JavaScript>
		Camera.SetCaptureMode(Camera.CAPTURE_MODE_PHOTO);
		Camera.SetFlashMode(Camera.FLASH_MODE_AUTO);
	</JavaScript>
</CameraView>
```

yupferris:
Ah right, nice one :)

Issue #270, opened 2017-07-24
Drawing curves in straight lines results in invisible curves


COCPORN:
Tested on Windows Preview, Fuse version 1.1.0.

This:

```
   <Panel Alignment="Center" Width="300" Height="300">
        <Curve StrokeColor="#000" StrokeWidth="10">
            <CurvePoint At="0.4,0.5" />
            <CurvePoint At="0.6,0.5" />
        </Curve>
    </Panel>
```
...and...

```
   <Panel Alignment="Center" Width="300" Height="300">
        <Curve StrokeColor="#000" StrokeWidth="10">
            <CurvePoint At="0.5,0.4" />
            <CurvePoint At="0.5,0.6" />
        </Curve>
    </Panel>
```

...result in the curve not being rendered at all. Changing the values that are the same (0.5) to something very similar (4.999 or 5.0001) makes the curve render as expected.

mortoray:
I can't reproduce this on another platform.

It's defintiely a floating point issue, though I'm not sure it's with our code.  Try putting `Type="Straight"` on the `Curve` and see if it goes away. If it does then it points towards our tangent code, if it doesn't go away then it points towards a backend/dotnet issue.

COCPORN:
`Style="Straight"` also results in an invisible curve.

mortoray:
It doesn't render on the DotNet target on OSX either. So there's a way to debug it at least.

fusebuild:
The forum thread [Curve don't draw line](https://www.fusetools.com/community/forums/bug_reports/curve_dont_draw_line) posted by SvenWS was linked to this issue.

mortoray:
This only seems to happen with a line that is either horizontal or vertical, and only on DotNet. There is possibly something that doesn't calculate any size in this case.

mortoray:
We have a check for `IsEmptyPath` that checks the bounds of the path. If either was 0 it'd not draw. This is because the backend API seems to crash on 0x0 bounds. This of coruse filters out a vertical/horz line since once dimension is 0. I'm fixing by only skipping the draw if both dimensions are 0, which seems to work. We can't draw a single stroked point, but I'm willing to live with that limitation.

kusma:
I have a feeling that this check is a bit confused. While it might crash on 0x0 bounds, I suspect that due to stroking, we do not actually have 0x0 bounds here.

[...]

PR #470, opened 2017-09-19
Fixes DotNet not drawing horz/vert lines

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/270

This PR contains:
- [x] Changelog
- [ ] ~Documentation~ no API change
- [ ] ~Tests~ Visual only result, no automation

PR #475, opened 2017-09-19
Fix axis-aligned lines

kusma:
This is an alternative to #470 to fix issue #270. I believe this takes the bull by its horns a bit more than the original fix, as it avoids the specific thing that cause a crash under Mono instead of caring about zero-bounds paths before expanding.

This PR contains:
- [x] Changelog
- [ ] ~Documentation~ no API changes
- [x] Tests


kusma:
@mortoray: I've tested the case of a single point path separately, and SurfaceTestApp works fine with this change as well.

kusma:
In fact, I also have a local patch to add a test for a single point. The reason I didn't include it, is that this seems to render differently on Win32 and macOS, but the reason seems to be unrelated to this patch (because the same thing can be triggered with more complicated paths).

PR #482, opened 2017-09-21
Native DatePickers

yupferris:
Introduces the `Fuse.Controls.DatePicker` package, which includes a `DatePicker` class that proxies native views on Android and iOS. On desktop, a simple placeholder is shown. The public API is in place and the implementations appear to be stable and robust.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


Source/Fuse.Controls.DatePicker/DatePicker.Docs.uno
```diff
@@ -0,0 +1,66 @@
...
+							someDate.value = new Date(Date.parse("1984-06-08T00:00:00.000Z"));
+						}
+					};
+				</JavaScript>
+
+				<NativeViewHost>
+					<DatePicker Value="{someDate}" MinValue="{minDate}" MaxValue="{maxDate}" />
+				</NativeViewHost>
+
+				<Button Text="Who you gonna call?" Clicked="{whoYouGonnaCall}" Margin="5" />
```
LuisRodriguezLD: Ghostbusters!

yupferris: check what it sets the date to ;)

yupferris:
@kusma just so you're aware, I'd like to look into those feedback issues, so this shouldn't be merged yet

yupferris:
@kusma all issues fixed, ready for merge. :)

Source/Fuse.Controls.DatePicker/iOS/DatePicker.uno
```diff
@@ -0,0 +1,188 @@
...
+
+			return [utcCalendar dateFromComponents:utcComponents];
+		@}
+	}
+
+	extern(!iOS) class DatePickerView
+	{
+		[UXConstructor]
+		public DatePickerView([UXParameter("Host")]IDatePickerHost host) { }
+	}
```
kusma: Do you really need this? Having two classes with the same name is kind of nasty for the doc-generator etc. These classes doesn't have docs, so it probably isn't an issue for that reason, but it seems a bit pointless...

Source/Fuse.Controls.DatePicker/Android/DatePicker.uno
```diff
@@ -0,0 +1,235 @@
...
+
+			return msSince1970InUtc;
+		}
+	}
+
+	extern(!Android) class DatePickerView
+	{
+		[UXConstructor]
+		public DatePickerView([UXParameter("Host")]IDatePickerHost host) { }
+	}
```
kusma: Do you really need this? Having two classes with the same name is kind of nasty for the doc-generator etc. These classes doesn't have docs, so it probably isn't an issue for that reason, but it seems a bit pointless...

Source/Fuse.Controls.DatePicker/iOS/DatePicker.uno
yupferris: Good call, I think I added that when testing only one platform impl and forgot to remove it. I'll see if it's needed and remove it if not.

yupferris: aha, now I remember why I have that. Otherwise the UX compiler complains about missing types in `DatePicker.ux`:

```ux
<Fuse.Controls.DatePickerBase ux:Class="Fuse.Controls.DatePicker" ux:Name="self">
	<Panel ux:Template="GraphicsAppearance" Margin="4" Background="#eee" Padding="30">
		<Text Value="GraphicsAppearance for DatePicker not implemented!" Alignment="Center" TextWrapping="Wrap" />
		<Rectangle Layer="Background">
			<Stroke Color="#000" />
		</Rectangle>
	</Panel>
	<Fuse.Controls.Native.iOS.DatePickerView ux:Condition="iOS" ux:Template="iOSAppearance" Host="self" />
	<Fuse.Controls.Native.Android.DatePickerView ux:Condition="Android" ux:Template="AndroidAppearance" Host="self" />
</Fuse.Controls.DatePickerBase>
```

I would've assumed `ux:Condition` would have solved that, but perhaps that's only evaluated at runtime.

This pattern comes from [our current docs](https://www.fusetools.com/docs/native-interop/native-ux-components) btw.

kusma: OK, that's a sucky implementation of the UX compiler in this regard, but OK. It is what it is ;)

Source/Fuse.Controls.DatePicker/Android/DatePicker.uno
```diff
@@ -0,0 +1,235 @@
...
+			var dotNetTicks = dotNetTicksRelativeToUnixEpoch + UnixEpochInDotNetTicks;
+
+			return new DateTime(dotNetTicks, DateTimeKind.Utc);
+		}
+
+		public static long ConvertDateTimeToMsSince1970InUtc(DateTime dt)
+		{
+			dt = dt.ToUniversalTime();
+
+			var dotNetTicks = dt.Ticks;
```
kusma: micro-nit, this could have been written like this
```uno
var dotNetTicks = dt.ToUniversalTime().Ticks;
```
...to avoid the temporary. But it's IMO not worth fixing ATM.

Source/Fuse.Controls.DatePicker/iOS/DatePicker.uno
yupferris: agreed, https://github.com/fusetools/uno/issues/1348

kusma:
I'd like to point out that this is a beautiful PR. The code is very clear and to the point, and most strange details seems to be properly justified and commented. Very nice work!

yupferris:
thanks 😊

Issue #396, opened 2017-08-24
Video.Position property does not work with Observable bindings.


bstrr:
It would be great to have the position time in seconds reported back using the Position property on the Video element. This would allow me to show the playback time as well as calculate progress on live video, where the Progress data is incorrect (as the video duration isn't static).

PR #486, opened 2017-09-22
Feature video scriptclass

kristianhasselknippe:
An attempt at fixing #396

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ## Video
 - Fixed bug in Video where playback actions, like `Play`, used before the video was initialized would end up getting swallowed.
+- Added some JavaScript methods to `Video` to make it easier to control playback from JavaScript, as well as obtaining information the video duration.
+- Made `Video.Duration` and `Video.Position` property-bindable.
 
 ## Fuse.Reactive framework changes (Uno-level)
 - These are breaking changes, but very unlikely to affect your app:
--- a/Source/Fuse.Controls.Video/Video.ScriptClass.uno
+++ b/Source/Fuse.Controls.Video/Video.ScriptClass.uno
@@ -1,0 +1,82 @@
+using Uno;
+using Uno.UX;
+
+using Fuse.Scripting;
+
+namespace Fuse.Controls
+{
+	public partial class Video
+	{
+		static Video()
+		{
+			ScriptClass.Register(typeof(Video),
+				new ScriptMethod<Video>("getDuration", getDuration, ExecutionThread.Any),
+				new ScriptMethod<Video>("resume", resume, ExecutionThread.MainThread),
+				new ScriptMethod<Video>("pause", pause, ExecutionThread.MainThread),
+				new ScriptMethod<Video>("stop", stop, ExecutionThread.MainThread));
+		}
+
+		object _durationMutex = new object();
+
+		// private fields used by ScriptClass for retrieving the position from JavaScript
+		double _outDuration;
+		void UpdateScriptClass(double duration)
+		{
+			lock (_durationMutex)
+				_outDuration = duration;
+		}
+
+		/**
+			Gets the duration of the Video in seconds. getDuration will return 0 until the video has initialized.
+
+			@scriptmethod getDuration()
+		*/
+		static object getDuration(Context c, Video v, object[] args)
+		{
+			if (args.Length != 0)
+				Fuse.Diagnostics.UserError("getDuration takes 0 arguments, but " + args.Length + " was supplied", v);
+
+			lock (v._durationMutex)
+				return (object)v._outDuration;
+		}
+
+		/**
+			Resumes playback from the current position.
+
+			@scriptmethod resume()
+		*/
+		static void resume(Context c, Video v, object[] args)
+		{
+			if (args.Length != 0)
+				Fuse.Diagnostics.UserError("resume takes 0 arguments, but " + args.Length + " was supplied", v);
+
+			v.Resume();
+		}
+
+		/**
+			Pauses playback, leaving the current position as-is.
+
+			@scriptmethod pause()
+		*/
+		static void pause(Context c, Video v, object[] args)
+		{
+			if (args.Length != 0)
+				Fuse.Diagnostics.UserError("pause takes 0 arguments, but " + args.Length + " was supplied", v);
+
+			v.Pause();
+		}
+
+		/**
+			Stops playback and returns to the beginning of the video.
+
+			@scriptmethod stop()
+		*/
+		static void stop(Context c, Video v, object[] args)
+		{
+			if (args.Length != 0)
+				Fuse.Diagnostics.UserError("stop takes 0 arguments, but " + args.Length + " was supplied", v);
+
+			v.Stop();
+		}
+	}
+}
--- a/Source/Fuse.Controls.Video/Video.uno
+++ b/Source/Fuse.Controls.Video/Video.uno
@@ -122,7 +122,7 @@ namespace Fuse.Controls
 			</DockPanel>
 			
 	*/
-	public class Video : Panel, IMediaPlayback
+	public partial class Video : Panel, IMediaPlayback
 	{
 
 		VideoSource _source;
@@ -485,6 +485,8 @@ namespace Fuse.Controls
 
 		void OnProgressChanged(object sender, EventArgs args)
 		{
+			UpdateScriptClass(Duration);
+
 			OnPropertyChanged(_progressName);
 			if (ProgressChanged != null)
 				ProgressChanged(this, new ValueChangedArgs<double>(Progress));
--- a/Source/Fuse.Controls.Video/Video.uno
+++ b/Source/Fuse.Controls.Video/Video.uno
@@ -481,15 +481,24 @@ namespace Fuse.Controls
 			}
 		}
 
+		static Selector _positionName = "Position";
+		static Selector _durationName = "Duration";
 		static Selector _progressName = "Progress";
 
 		void OnProgressChanged(object sender, EventArgs args)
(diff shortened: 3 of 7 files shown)


Source/Fuse.Controls.Video/Video.ScriptClass.uno
```diff
@@ -0,0 +1,76 @@
...
+			return (object)ret;
+		}
+
+		/**
+			Resumes playback from the current position.
+
+			@scriptmethod resume()
+		*/
+		static void resume(Context c, Video v, object[] args)
+		{
```
mortoray: Please put checks on the `args` here (and other functions). If it's not 0-length emit a Diagnostics.UserError.

Source/Fuse.Controls.Video/Video.ScriptClass.uno
```diff
@@ -0,0 +1,76 @@
...
+using Fuse.Scripting;
+
+namespace Fuse.Controls
+{
+	public partial class Video
+	{
+		static Video()
+		{
+			ScriptClass.Register(typeof(Video),
+				new ScriptMethod<Video>("getPosition", getPosition, ExecutionThread.Any),
```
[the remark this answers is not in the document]
mortoray: That may only be for OBservable returns though, so if you decide it's not suitable here that's fine as well.

CHANGELOG.md
```diff
@@ -77,6 +77,7 @@ which will stop push notifications registering (and potentially asking for permi
 
 ## Video
 - Xamarin Mac was upgraded to support 64-bit executables on macOS.
+- Added some JavaScript methods to `Video` to make it easier to control playback from JavaScript, as well as obtaining information about playback position and video duration.
```
kusma: This is now in the wrong section of the changelog since we did 1.3. Would you mind rebasing, on master and moving this to the "unreleased"-section, @kristianhasselknippe?

Source/Fuse.Controls.Video/Video.ScriptClass.uno
```diff
@@ -0,0 +1,91 @@
...
+		*/
+		static object getPosition(Context c, Video v, object[] args)
+		{
+			if (args.Length != 0)
+				Fuse.Diagnostics.UserError("getPosition takes 0 arguments, but " + args.Length + " was supplied", v);
+
+			var ret = 0.0;
+			lock (v)
+				ret = v._outPosition;
+			return (object)ret;
```
kusma: why not the simpler
```uno
lock (v)
	return (object)v._outPosition;
```
instead?

Also, [`lock (this)` is generally frowned upon in C#](https://stackoverflow.com/questions/251391/why-is-lockthis-bad). I don't think this case is particularly bad, but perhaps you could add a `object _positionMutex = new object();`  just to prevent problem if someone else tries to lock `Fuse.Controls.Video`?

Source/Fuse.Controls.Video/Video.ScriptClass.uno
```diff
@@ -0,0 +1,91 @@
...
+		*/
+		static object getDuration(Context c, Video v, object[] args)
+		{
+			if (args.Length != 0)
+				Fuse.Diagnostics.UserError("getDuration takes 0 arguments, but " + args.Length + " was supplied", v);
+
+			var ret = 0.0;
+			lock (v)
+				ret = v._outDuration;
+			return (object)ret;
```
kusma: The questions for `getPosition` applies here as well...

Source/Fuse.Controls.Video/Video.uno
```diff
@@ -485,6 +487,12 @@ namespace Fuse.Controls
 
 		void OnProgressChanged(object sender, EventArgs args)
 		{
+			lock (this)
+			{
+				_outPosition = Position;
+				_outDuration = Duration;
+			}
```
kusma: I would kind of prefer if this passed Position and Duration off to a set-method `Video.ScriptClass.uno`, so the state and mutex-locking is contained together in there.

Source/Fuse.Controls.Video/Video.ScriptClass.uno
kristianhasselknippe: should i add this to Video? or in getPosition?

Source/Fuse.Controls.Video/Video.uno
kristianhasselknippe: Not sure i understand what you mean by this..

kusma: I mean something like this:
```patch
diff --git a/Source/Fuse.Controls.Video/Video.ScriptClass.uno b/Source/Fuse.Controls.Video/Video.ScriptClass.uno
index ff12d13..0ba31ea 100644
--- a/Source/Fuse.Controls.Video/Video.ScriptClass.uno
+++ b/Source/Fuse.Controls.Video/Video.ScriptClass.uno
@@ -17,6 +17,19 @@ namespace Fuse.Controls
 				new ScriptMethod<Video>("stop", stop, ExecutionThread.MainThread));
 		}
 
+ 		// private fields used by ScriptClass for retrieving the position from JavaScript
+		double _outPosition;
+		double _outDuration;
+		void UpdateScriptClass(double position, double duration)
+		{
+			lock (this)
+			{
+				_outPosition = position;
+				_outDuration = duration;
+			}
+		}
+
+
 		/**
 			Gets the current position in the Video in seconds.
 
diff --git a/Source/Fuse.Controls.Video/Video.uno b/Source/Fuse.Controls.Video/Video.uno
index 09b0786..93379b0 100644
--- a/Source/Fuse.Controls.Video/Video.uno
+++ b/Source/Fuse.Controls.Video/Video.uno
@@ -446,7 +446,6 @@ namespace Fuse.Controls
 			set { if (Playback != null) (Playback as IPlayback).Progress = value; }
 		}
 
-		double _outPosition; //private field used by ScriptClass for retrieving the position from JavaScript
 		/*** The position of the video in seconds */
 		public double Position
 		{
@@ -454,7 +453,6 @@ namespace Fuse.Controls
 			set { if (Playback != null) Playback.Position = value; }
 		}
 
-		double _outDuration; //private field used by ScriptClass for retrieving the duration from JavaScript
 		public double Duration
 		{
 			get { return Playback != null ? Playback.Duration : 0.0; }
@@ -487,12 +485,7 @@ namespace Fuse.Controls
 
 		void OnProgressChanged(object sender, EventArgs args)
 		{
-			lock (this)
-			{
-				_outPosition = Position;
-				_outDuration = Duration;
-			}
-
+			UpdateScriptClass(Position, Duration);
 			OnPropertyChanged(_progressName);
 			if (ProgressChanged != null)
 				ProgressChanged(this, new ValueChangedArgs<double>(Progress));
```

Source/Fuse.Controls.Video/Video.ScriptClass.uno
```diff
@@ -17,6 +17,20 @@ namespace Fuse.Controls
...
+
+		// private fields used by ScriptClass for retrieving the position from JavaScript
+		double _outPosition;
+		double _outDuration;
+		void UpdateScriptClass(double position, double duration)
+		{
+			lock (_positionMutex)
+				_outPosition = position;
+			lock (_durationMutex)
+				_outDuration = duration;
```
kusma: I don't think we need two mutexes for these, one should be enough... It's not a huge contention, two assigns vs one.

kristianhasselknippe: aaight 👍

kusma:
Yeah, so thinking about this whole thing a bit more, it seems like a bad idea to expose position to JavaScript, as javascript runs out of sync with the UI thread, so this gets inherently racy, and might lead to people trying to animate things from JavaScript.

And going back and reading #396 again, it seems like it's actually asking about UX expressions, and not JavaScript. So I think this should go back to the drawing-board.

kristianhasselknippe:
@kusma : Ref our discussion about making getPosition() a bindable property instead. I was hoping that the delay ux-function would basically allow me to reduce how often the position property changing would result in the bound value (Text.Value in this case) getting invalidated. Seems like the delay function would cause the value to just lag behind instead of reducing its update frequency.

kusma:
@kristianhasselknippe: Hm, yeah, I think you're right. But if you do `x * floor(foo / x)`, you should get the result you want.

Tests/ManualTests/ManualTestingApp/Singles/VideoPage.ux
```diff
@@ -38,7 +38,7 @@
 
 			</Video>
 			<Grid Dock="Bottom" ColumnCount="2" Rows="Auto">
-				<Text Value="Position: {position}" />
+				<Text Value="Position: {=delay({ReadProperty _video.Position},1)}" />
```
kusma: Yeah, I doubt we want this one now ;)

CHANGELOG.md
kusma: OK, so this description is going to need updating as well now. We no longer expose position to JS.

Source/Fuse.Controls.Video/Video.uno
```diff
@@ -481,11 +481,18 @@ namespace Fuse.Controls
...
+		static Selector _positionName = "Position";
+		static Selector _durationName = "Duration";
 		static Selector _progressName = "Progress";
 
 		void OnProgressChanged(object sender, EventArgs args)
 		{
+			UpdateScriptClass(Duration);
+
+			OnPropertyChanged(_positionName);
+			OnPropertyChanged(_durationName);
```
kristianhasselknippe: I realize it doesn't make much sense to notify a Duration change in `OnPropertyChanged`... will fix on monday

PR #487, opened 2017-09-25
Fix/navcontext

mortoray:
Fixes some defects relating to dynamic pages in a PageControl.

This PR contains:
- [ ] ~Changelog~ No released feature changed. Fixes defects found internally.
- [ ] ~Documentation~ No API changes
- [x] Tests

### DIFF
--- a/Source/Fuse.Common/Tests/FuseTest/InstanceCounter.uno
+++ b/Source/Fuse.Common/Tests/FuseTest/InstanceCounter.uno
@@ -1,0 +1,26 @@
+using Uno;
+
+using Fuse;
+
+namespace FuseTest
+{
+	/** Maintains a static count of how often this has been initialized. */
+	public sealed class InstanceCounter : Behavior
+	{
+		static int _count = 0;
+		public static int Count 
+		{
+			get { return _count; }
+		}
+		
+		public static void Reset()
+		{
+			_count = 0;
+		}
+		
+		public InstanceCounter()
+		{
+			_count++;
+		}
+	}
+}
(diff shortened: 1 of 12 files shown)


kusma:
Yikes, this is an uncomfortably large change to come this late... Is there no simpler way of dealing with the problem?

mortoray:
The majority of the new code is test cases. The functional code change isn't as big as the lines changed might suggest. Most of that, if not all, is also in code that touches only the new `PageControl.Pages`, and slightly the `Navigator.Pages` features -- not previous features.

kusma:
@mortoray: I did look at the code, and it's not the size of the tests that worry me. In particular, the changes in `Source/Fuse.Controls.Navigation/PageControl.Pages.uno` worry me; they change a lot of code. Yes, it's new features, but that isn't *very* settling to me, at least not on its own. If it's functionality not currently in use, why do we need to patch it on the release-branch in the first place?

Just to be clear: this isn't me saying "no, we can't do this", I'm just trying to understand the situation better...

mortoray:
The fix is to `PageControl.Pages` in how it creates pages. The feature does work, but poorly (excessive instantiation/creation of children) without these changes. It seems like it'd make more sense to have this fixed than rather to issue a patch that attempts to remove the feature it fixes  (which probably isn't possible).

It's because it's a newer feature I also think it's okay a bit late in the cycle, since no examples, nor tool testing, would pick up any problems in this part of the code. It's covered only by test cases now. But I'd like to see at least some minimal use of this feature in the next iteration, as it may help to address issues we'll have with Model based navigation (this is an essential part of that feature set).

kusma:
OK, so if I got your point: You're saying that fixing this feature in release-1.3 will enable adding examples using this feature? If it's only for models, then I think a fix can land in master instead. It's no problem to keep the feature in the tree until it's fixed, even if there's no users.

mortoray:
The feature isn't just for models, but models does depend on it. There's some use-cases I've seen in the forums that it would help right now; the reason why I didn't make it depends on models.

PR #489, opened 2017-09-25
Add native support for MultiDensityImageSource

kristianhasselknippe:
Meaning it can be used within a NativeViewHost.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Native/Android/ImageView.uno
+++ b/Source/Fuse.Controls.Native/Android/ImageView.uno
@@ -72,22 +72,44 @@ namespace Fuse.Controls.Native.Android
 	extern(Android) internal class ImageView : View, IImageView
 	{
 
+		ImageSource _imageSource;
 		public ImageSource ImageSource
 		{
 			set
 			{
+				if (ImageSource is MultiDensityImageSource)
+					((MultiDensityImageSource)ImageSource).ActiveChanged -= OnMultiDensityImageSourceActiveChanged;
+
 				if (value == null)
 					ImageHandle = null;
-				else if (value is FileImageSource)
+
+				_imageSource = value;
+
+				if (value is FileImageSource)
 					UpdateImage((FileImageSource)value);
 				else if (value is HttpImageSource)
 					UpdateImage((HttpImageSource)value);
 				else if (value is MultiDensityImageSource)
+				{
+					((MultiDensityImageSource)ImageSource).ActiveChanged += OnMultiDensityImageSourceActiveChanged;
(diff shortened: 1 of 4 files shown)


Source/Fuse.Controls.Native/Android/ImageView.uno
```diff
@@ -72,22 +72,34 @@ namespace Fuse.Controls.Native.Android
...
+			{
+				return _imageSource;
+			}
+		}
+
+		void OnImageSourceChanged()
+		{
+			SourceChanged;
+
+				var mds = (MultiDensityImageSource)ImageSource;
```
mortoray: Soemthing with the spacing is wrong here.

mortoray: There's a chunk of code missing too it seems, when compared to the iOS code.

Source/Fuse.Controls.Native/iOS/ImageView.uno
```diff
@@ -13,22 +13,44 @@ namespace Fuse.Controls.Native.iOS
...
+		{
+			if (ImageSource == null)
+				return;
+			else if (ImageSource is FileImageSource)
+				UpdateImage((FileImageSource)ImageSource);
+			else if (ImageSource is HttpImageSource)
+				UpdateImage((HttpImageSource)ImageSource);
+			else if (ImageSource is MultiDensityImageSource)
+			{
+				if (ImageSource != null && ImageSource is MultiDensityImageSource)
```
mortoray: This cleanup won't happen if the user switches from a MultiDensityImageSource to a different type of ImageSource.
A common pattern in fuselibs is on settting a new value, cleanup the old one first, regardless of what it is, then handle the new value.  You've kind of mixed the two steps together here.

Source/Fuse.Elements/Resources/MultiDensityImageSource.uno
```diff
@@ -92,6 +110,7 @@ namespace Fuse.Resources
 			}
 
 			SwapActive(use);
+			OnSourcesChanged();
```
mortoray: The name of the event seems wrong, shouldn't it be `OnActiveChanged`?  It probably belongs in the `SwapActive` function.

Source/Fuse.Controls.Native/Android/ImageView.uno
kristianhasselknippe: yeah, no idea how i missed that, but that's indeed what happened :S

PR #490, opened 2017-09-25
Fuse.Reactive cleanup

sebbert:
Large refactoring of Fuse.Reactive interfaces, implementations and related tets.

This introduces IObservableArray and IObservableObject which allows for a more general purpose interop with the datacontext. Changes the IWriteable interface slightly.

Adds a lot of documentation for those interfaces.

Fuse.Reactive.Observable no longer pushes initial value on subscribe. Consumers are fixed to cope with the new reality.

The IRaw interface gains an ReflectedRaw property to accommodate change detection engines to reflect a different object from what will be seen in JS callbacks.

Fuse.Reactive.JavaScript gains a reference to Fuse.Reactive.Expressions which requires some disambiguation of namespaces (Uno.UX also has a Property class).

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

PR #499, opened 2017-09-26
Feature TreeObservable

sebbert:
~~Requires #490~~ (now merged)

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Reactive.JavaScript/ArrayMirror.uno
+++ b/Source/Fuse.Reactive.JavaScript/ArrayMirror.uno
@@ -27,6 +27,9 @@ namespace Fuse.Reactive
 
 		public override void Unsubscribe()
 		{
+			if (_isDisposed) return;
+			_isDisposed = true;
+
 			for (int i = 0; i < _items.Count; i++)
 			{
 				var d = _items[i] as ValueMirror;
--- a/Source/Fuse.Reactive.JavaScript/ArrayMirror.uno
+++ b/Source/Fuse.Reactive.JavaScript/ArrayMirror.uno
@@ -25,10 +25,11 @@ namespace Fuse.Reactive
 
 		internal object[] ItemsReadonly { get { return _items.ToArray(); } }
 
+		bool _hasUnsubscribed;
 		public override void Unsubscribe()
 		{
-			if (_isDisposed) return;
-			_isDisposed = true;
+			if (_hasUnsubscribed) return;
+			_hasUnsubscribed = true;
 
 			for (int i = 0; i < _items.Count; i++)
 			{
--- a/Source/Fuse.Reactive.JavaScript/ObjectMirror.uno
+++ b/Source/Fuse.Reactive.JavaScript/ObjectMirror.uno
@@ -26,10 +26,11 @@ namespace Fuse.Reactive
 			}
 		}
 
+		bool _hasUnsubscribed;
 		public override void Unsubscribe()
 		{
-			if (_isDisposed) return;
-			_isDisposed = true;
+			if (_hasUnsubscribed) return;
+			_hasUnsubscribed = true;
 
 			foreach (var p in _props)
 			{
--- a/Source/Fuse.Reactive.JavaScript/ObjectMirror.uno
+++ b/Source/Fuse.Reactive.JavaScript/ObjectMirror.uno
@@ -28,6 +28,9 @@ namespace Fuse.Reactive
 
 		public override void Unsubscribe()
 		{
+			if (_isDisposed) return;
+			_isDisposed = true;
+
 			foreach (var p in _props)
 			{
 				var d = p.Value as ValueMirror;
--- a/Source/Fuse.Reactive.JavaScript/Tests/UX/TreeObservable.Set.ux
+++ b/Source/Fuse.Reactive.JavaScript/Tests/UX/TreeObservable.Set.ux
@@ -1,0 +1,29 @@
+<Panel ux:Class="UX.TreeObservable.Set">
+	<JavaScript>
+		var TreeObservable = require("FuseJS/TreeObservable");
+		function Tree() {}
+		Tree.prototype = Object.create(TreeObservable.prototype);
+		let tree = new Tree();
+
+		tree.__fuse_id = 0;
+		tree.foo = {
+			__fuse_id: 1,
+			bar: {
+				__fuse_id: 2,
+				baz: "initial"
+			}
+		};
+
+		module.exports = {
+			tree: tree,
+			setDeep: function() {
+				tree.foo.bar.baz = "changed";
+				TreeObservable.set.call(tree, "foo", "bar", "baz", "changed");
+			}
+		};
+	</JavaScript>
+
+	<Text ux:Name="t" Value="{tree.foo.bar.baz}" />
+
+	<FuseTest.Invoke ux:Name="setDeep" Handler="{setDeep}" />
+</Panel>
--- a/Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
+++ b/Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
@@ -1,0 +1,88 @@
+using Uno.Collections;
+using Uno;
+
+namespace Fuse.Reactive
+{
+	class TreeArray : ArrayMirror, IObservableArray
+	{
+		internal TreeArray(Scripting.Array arr): base(arr) {}
+
+		public IDisposable Subscribe(IObserver observer)
+		{
+			return new ArraySubscription(this, observer);
+		}
+
+		class ArraySubscription: Subscription
+		{
+			readonly IObserver _observer;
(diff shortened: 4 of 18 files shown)


Source/Fuse.Reactive.JavaScript/ArrayMirror.uno
```diff
@@ -27,6 +27,9 @@ namespace Fuse.Reactive
 
 		public override void Unsubscribe()
 		{
+			if (_isDisposed) return;
+			_isDisposed = true;
```
kusma: This feels a bit like abuse. Unsubscribe and Dispose isn't quite the same... Perhaps they should be, though?

Source/Fuse.Reactive.JavaScript/Tests/UX/TreeObservable.Set.ux
```diff
@@ -0,0 +1,29 @@
+<Panel ux:Class="UX.TreeObservable.Set">
+	<JavaScript>
+		var TreeObservable = require("FuseJS/TreeObservable");
+		function Tree() {}
+		Tree.prototype = Object.create(TreeObservable.prototype);
+		let tree = new Tree();
+
+		tree.__fuse_id = 0;
```
mortoray: What's with the explicit setting of `__fuse_id` ? Isn't this automatic, or is it something the model differ does later?

Source/Fuse.Reactive.JavaScript/ObjectMirror.uno
```diff
@@ -28,6 +28,9 @@ namespace Fuse.Reactive
 
 		public override void Unsubscribe()
 		{
+			if (_isDisposed) return;
+			_isDisposed = true;
```
kusma: Same comment as above, this feels like an abuse...

Source/Fuse.Reactive.JavaScript/ValueMirror.uno
```diff
@@ -22,6 +22,7 @@ namespace Fuse.Reactive
 		readonly object _raw;
 		public object Raw { get { return _raw; } }
 		public virtual object ReflectedRaw { get { return _raw; } }
+		protected bool _isDisposed = false;
```
kusma: I don't think this is the right place to introduce this. I think it's better to introduce these in ObjectMirror and ArrayMirror. And then we can maybe even name it a bit better...

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+					var newValuesJs = ctx.NewArray(nv);
+
+					var replaceAllFn = (Scripting.Function) ctx.Evaluate("replaceAll",
+						"(function(array, values) {" +
+							"if('__fuse_replaceAll' in array) array.__fuse_replaceAll(values);" +
+							"else {"+
+								"array.length = 0;"+
+								"Array.prototype.push.apply(array, values);"+
+							"}"+
+						"})");
```
kusma: Looks like this one is constant, do we really need to re-create it every time?

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeObject.uno
```diff
@@ -0,0 +1,111 @@
...
+					else
+					{
+						_obj[_key] = _value;
+					}
+				}
+			}
+
+			bool IPropertySubscription.TrySetExclusive(string key, object newValue)
+			{
+				var t = SubscriptionSubject as TreeObject;
```
kusma: Please do
```uno
var t = (TreeObject)SubscriptionSubject;
```

This is clearer. We're going to be dereferencing `t`  on the next statement anyway, so it needs to be non-null. Doing a proper cast gives us a better error-message if this was every wrong.

Source/Fuse.Reactive/ObserverMap.uno
```diff
@@ -54,14 +54,16 @@ namespace Fuse.Reactive
 		{
 			Detach();
 			_observable = obs;
-			_slave = slave;
```
mortoray: This bit is likely to conflict with https://github.com/fusetools/fuselibs-public/pull/487
That PR also fixes this issue, in the same way, but with other changes as well.  Maybe we can remove the changes to ObserverMap from this PR and get that other PR merged to 1.3, then merged to master first? Or just deal with conflict later?

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+
+			_items.Clear();
+
+			for(var i = 0; i < newValues.Length; ++i)
+			{
+				_items.Add(newValues[i]);
+			}
+
+			var sub = Subscribers as ArraySubscription;
+			if(sub != null)
```
kusma: space between `if` and opening paren.

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+			public void OnRemoveAt(int index)
+			{
+				_observer.OnRemoveAt(index);
+				var next = Next as ArraySubscription;
+				if (next != null) next.OnRemoveAt(index);
+			}
+
+			public void OnReplaceAll(IArray values, ArraySubscription exclude)
+			{
+				if(this != exclude) _observer.OnNewAll(values);
```
kusma: space between `if` and opening paren.

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+
+				var next = Next as ArraySubscription;
+				if(next != null) next.OnReplaceAll(values, exclude);
```
kusma: space between `if` and opening paren.

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+					_target = target;
+					_newValues = newValues;
+				}
+
+				public void Perform()
+				{
+					var ctx = _worker.Context;
+
+					var nv = new object[_newValues.Length];
+					for(var i = 0; i < _newValues.Length; ++i) {
```
kusma: space between `for` and opening paren.

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+					{
+						return _values[index];
+					}
+				}
+			}
+		}
+
+		internal void ReplaceAll(IArray newValues, ArraySubscription exclude)
+		{
+			for(var i = 0; i < _items.Count; ++i)
```
kusma: space between `for` and opening paren.

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -0,0 +1,193 @@
...
+			{
+				ValueMirror.Unsubscribe(_items[i]);
+			}
+
+			_items.Clear();
+
+			for(var i = 0; i < newValues.Length; ++i)
```
kusma: space between `for` and opening paren.

Source/Fuse.Reactive.JavaScript/Tests/UX/TreeObservable.Set.ux
sebbert: These are automatically generated by Model system, yes. It's an ugly workaround for the fact that `Scripting.Object` doesn't guarantee object equality/identity on the Uno side.

Source/Fuse.Reactive/ObserverMap.uno
sebbert: I think that sounds like a good approach. Please ping me when it's merged and I'll rebase.

Source/Fuse.Reactive.JavaScript/TreeObservable/TreeArray.uno
sebbert: We definitely don't, although it needs to be evaluated in the right context. What's the correct place for this in your opinion? Builtins?

kusma: TBH, I'm not really sure.

kusma: Lets just do that as a follow-up to get things moving.

kusma:
Seems like this broke master :(

kusma:
Reverted. Please find the problem and send a new PR.

PR #500, opened 2017-09-26
Fix TextWrapping="NoWrap" on Android TextView

kristianhasselknippe:
The previous behavior was to set max lines to 1 in the case of
Wrapping="NoWrap". This proved to not give the right behavior, as the
TextView should still be multiline even though it shouldn't wrap
automatically. Changing this to setting horizontally scrolling to false
gives the correct behavior.

Fixes fusetools/fuselibs#4010

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## TextView
+
+- Fixed bug on Android where setting `TextWrapping="NoWrap"` would force the `TextView` to be single line. New behavior is to instead allow the view to scroll horizontally instead of automatically wrapping the text.
+
 # 1.3
 
 ## 1.3.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,9 +1,12 @@
 # Unreleased
 
+## TextView
+- Fixed bug on Android where setting `TextWrapping="NoWrap"` would force the `TextView` to be single line. New behavior is to instead allow the view to scroll horizontally instead of automatically wrapping the text.
+
 ## MultiDensityImageSource
 - Added native support, meaning it can be used by images inside a `NativeViewHost`.
 
-### Fuse.Reactive framework changes (Uno-level)
+## Fuse.Reactive framework changes (Uno-level)
```
nekronos: Remove this change? @kristianhasselknippe

kusma: @nekronos: This was my change, and it's simply just a lazy version of a rebase. This change is already on master.

PR #503, opened 2017-09-27
ViewHandle.iOS: fix translation bug

nekronos:
The properties frame, bounds and center should all point to the same property behind the scenes. Looks like this is not the case in thirdparty libraries. For example setting a frame on Ad views from firebase has no effect on the view position.

Fixing this by always setting center to 0,0.

Fixes https://github.com/fuse-compound/Fuse.Firebase/issues/51

This PR contains:
- [x] Changelog
- [ ] ~~Documentation~~
- [ ] ~~Tests~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@
 
 ## 1.3.0
 
+### Native UI:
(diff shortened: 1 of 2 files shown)


kusma:
@nekronos: Should we consider including this in release-1.3, perhaps?

nekronos:
@kusma done

PR #504, opened 2017-09-27
State migration for Preview

mortoray:
Implements the state migration interface to allow Preview to save/restore the state.

For preview you need to call the `SavePreviewState` prior to unrooting the old nodes and then `RestorePreviewState` prior to creating/adding the new nodes:

- `var store = app.RootViewport.SavePreviewState()`
- Unroot/cleanup
- `app.RootViewport.RestorePreviewState(store)`
- add new children

This PR contains:
- [ ] ~Changelog~ - no user visible API changes
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Nodes/PreviewState.uno
+++ b/Source/Fuse.Nodes/PreviewState.uno
@@ -4,8 +4,6 @@ namespace Fuse
 {
 	/** 
 		State management in Preview for reloading/reifying. This has not use other than when running in preview.
-		
-		TODO: Reduce as much as possible if not defined(DESIGNMODE)
 	*/
 	class PreviewState
 	{
@@ -101,15 +99,9 @@ namespace Fuse
 			return null;
 		}
 	}
-	
 
 	interface IPreviewStateSaver
 	{
 		void Save( PreviewStateData data );
-	}
-	
-	interface IStateSource
-	{
-		
 	}
 }
--- a/Source/Fuse.Nodes/PreviewState.uno
+++ b/Source/Fuse.Nodes/PreviewState.uno
@@ -19,6 +19,13 @@ namespace Fuse
 		
 		public void SetState(PreviewStateData data)
 		{
+			if (data.Consumed)
+			{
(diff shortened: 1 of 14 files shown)


Source/Fuse.Nodes/PreviewState.uno
```diff
@@ -0,0 +1,107 @@
...
+		Dictionary<string, Entry> _data = new Dictionary<string, Entry>();
+		
+		/** Sets the data for a key. Note that a null value will not be considered a value, and will erase an existing value. */
+		public void Set( string key, object data )
+		{
+			_data[key] = new Entry{ Data = data, Consumed = false };
+		}
+		
+		/** Returns true if data has been stored for this key. Note that a null value is not considered to be actual data. */
+		public object Has( string key )
```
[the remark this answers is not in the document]
mortoray: Yes. I guess I missed a test for that as well.

mortoray:
It works in Preview so this should be good for merging.

Source/Fuse.Nodes/PreviewState.uno
```diff
@@ -0,0 +1,107 @@
...
+				n = n.Parent;
+			}
+			return null;
+		}
+	}
+	
+	class PreviewStateData
+	{
+		/** Has this state data already been consumed/restored? */
+		public bool Consumed;
```
sebbert: Although there is a check for this above, I can't see it being set anywhere

kusma: Yeah, this seems to be unused.

kusma: Yeah, this seems to be unused. Please remove :)

mortoray: It's not meant to be unused. I forget to set it. It's a safety check.

mortoray:
@kusma Can you recheck based on my changes? I fixed the mistake.

Source/Fuse.Nodes/PreviewState.uno
kusma: @mortoray: Yeah, looks good.

PR #513, opened 2017-09-29
Fuse.Drawing.Primitives: avoid differing uniform precision

kusma:
On Mali-400, the fragment-shader does not support highp floats, but
the vertex-shader does.

And since we set the default-precision to highp for all floats when
supported, we end up with different default-precision in the shader
stages. This cause issues when two uniforms have the same name in
both stages.

To side-step this, pre-calculate 1 / radius on the CPU, and use that
instead in the vertex-shader. This is a very minor optimization, but
is worth it simply due to avoiding using the same uniform-name.

Fixes fusetools/fuselibs#4266

### DIFF
--- a/Source/Fuse.Drawing.Primitives/Circle.uno
+++ b/Source/Fuse.Drawing.Primitives/Circle.uno
@@ -100,7 +100,7 @@ namespace Fuse.Drawing.Primitives
 				
 				float2 VertexPosition: V0 * (radius + extend*2);
 				LocalPosition: VertexPosition + center;
-				// Mali-400 hax GP16 max precision, which cannot square big numbers without overflowing.
+				// Mali-400 has FP16 max precision, which cannot square big numbers without overflowing.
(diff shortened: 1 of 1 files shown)


kusma:
@mortoray: the precision difference is about GLSL ES's `lowp` / `mediump` / `highp`, not `float` / `double`.

kusma:
In particular, [this](https://github.com/fusetools/uno/blob/323973482084bfbdfb7853255e42c2621486f633/Library/Core/UnoCore/Source/Uno/Runtime/Implementation/ShaderBackends/OpenGL/GLProgram.uno#L46) code.

mortoray:
I don't understand how changing from division to multiplication changes the shader's precision though? Or is it just because you're using a different name from `radius` -- and didn't want to do `var radius2 = radius;`?

kusma:
Yes, the division -> multiplication is just an optimization (GPUs generally perform `a / b` by doing `a * (1 / b)`), it's the different name that fixes the bug. I just saw a way of fixing the problem without adding an obscure hack.

PR #517, opened 2017-10-02
Fuse.Drawing.Surface: re-introduce IsEmptyPath-hack

kusma:
In 6bed36f, we dropped the incorrect IsEmptyPath-hack because we
had a more directed fix. However, this doesn't seem to cover all
cases. So let's reintroduce a less broken version of this check
(e.g one that checks if *both* width and height is zero) just to
be safe.

We should probably try to get rid of this at the benefit of a more
targetted fix later, but let's not try to deal with that right now.

This fixes a crash on startup in ManualTestingApp on dotnet-targets.

### DIFF
--- a/Source/Fuse.Drawing.Surface/DotNetSurface.uno
+++ b/Source/Fuse.Drawing.Surface/DotNetSurface.uno
@@ -156,6 +156,8 @@ namespace Fuse.Drawing
 			var actualPath = (DotNetCanvasPath)path;
 
 			var graphicsPath = actualPath.Path.GetGraphicsPath();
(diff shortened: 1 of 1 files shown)


kusma:
Just to be clear, this triggered in the `ImageFill`-block inside `DotNetSurface.StrokePath`. The failure happens while trying to create the Bitmap that were to be inserted into `_tiledImages`. So to do this cleanly, we need a dedicated guard inside that code-path.

This approach here is just an attempt to be as conservative as we can, by going back to what we did before (modulo that single-dimension change, which would make us start failing the new tests).

PR #523, opened 2017-10-03
Add support for exif orientation from byte[] 

kristianhasselknippe:
This is implemented for android,ios and dotnet, but there seems to be a
problem with some android devices and their ability to get the correct
orientation information. In particular, the images added in the tests in
this commit was not handled well by the samsung and two nexus phones i
tested on. iOS and DotNet did report correct information however. These
test images in particular were included since they were very small. This
implementation was tested with more image for which all implementations
reported the correct orientation.

Need to add an entry here https://docs.google.com/spreadsheets/d/1mYt9jve20DH2Zaxa03l8GSqZuMcqC-cUZoUKyvegxWM/edit#gid=0
when this is merged, as it uses https://drewnoakes.com/code/exif/

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests


kristianhasselknippe:
looking into whether there might be some alternative libraries to make the android implementation more reliable

kristianhasselknippe:
Im reporting ExifRotation.Undefined if run on an unsupported platform, like native, since i imagine this class being used by types that are supported by native. Should i instead throw an exception or emit a diagnostics message?

mortoray:
Does WIP mean this isn't ready yet?  You can use the "In Progress" tag for that.

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+				CGDataProviderRelease(dataProvider);
+			}
+
+			int rotation = 0;
+			NSString* tagTarget = @"Orientation";
+			for (id tag in metadataArray) {
+				CFStringRef tagName = CGImageMetadataTagCopyName((CGImageMetadataTagRef)tag);
+				if ([tagTarget isEqualToString: (__bridge NSString*)tagName]) {
+					CFTypeRef rot = CGImageMetadataTagCopyValue((CGImageMetadataTagRef)tag);
+					rotation = [((__bridge NSNumber*)rot) intValue];
```
nekronos: Can you `break` here?

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+			}
+			return (int)_buffer[_position++];
+		}
+
+		[Foreign(Language.Java)]
+		Java.Object CreateStream(Func<int> readByte)
+		@{
+			return new java.io.InputStream() {
+				public int read() {
+					return (int)readByte.run();
```
nekronos: Is this cast really needed?

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+		}
+
+		int _position = 0;
+		int ReadByte()
+		{
+			if (_position >= _buffer.Length)
+			{
+				return -1;
+			}
+			return (int)_buffer[_position++];
```
nekronos: You dont need this cast, `byte` to `int` is implicit

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+						CFRelease(metadata);
+					}
+					CFRelease(source);
+				}
+				CGDataProviderRelease(dataProvider);
+			}
+
+			int rotation = 0;
+			NSString* tagTarget = @"Orientation";
+			for (id tag in metadataArray) {
```
nekronos: What happens if `metadataArray` is `NULL`?

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+		@{
+
+			CGDataProviderDirectCallbacks callbacks = {
+				.getBytePointer = 0,
+				.getBytesAtPosition = getBytesAtPosition
+			};
+			UnoCallbackBox box = { readByte };
+
+			CGDataProviderRef dataProvider = CGDataProviderCreateDirect(&box, length, &callbacks);
+			CGImageSourceRef source = CGImageSourceCreateWithDataProvider(dataProvider, NULL);
```
kusma: You don't need the readByte-stuff, that's just overhead.
Here's a routine to create a CGImageSourceRef directly from a byte-array:
https://github.com/kusma/fuselibs-public/blob/bitmap/Source/Fuse.Common/Internal/Bitmap.uno#L110

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+			if (_position >= _buffer.Length)
+			{
+				return -1;
+			}
+			return (int)_buffer[_position++];
+		}
+
+		[Foreign(Language.Java)]
+		Java.Object CreateStream(Func<int> readByte)
+		@{
```
kusma: Similarly to on iOS, you shouldn't need this. You can create an InputStream from a byte-buffer using this helper:
https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Android/AndroidInterop.uno#L66

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+					return -1;
+				}
+				try {
+					int orientation = directory.getInt(ExifIFD0Directory.TAG_ORIENTATION);
+					return orientation;
+				} catch (MetadataException e) {
+				}
+			} catch (ImageProcessingException e) {
+			} catch (IOException e) {
+			}
```
kusma: I'm unable to parse the indentation here...

But even so, do we really want to ignore all these exceptions?! That sounds dubious to me...

Source/Fuse.Elements/Resources/Exif/ExifHelper.h
```diff
@@ -0,0 +1,9 @@
+#include <uObjC.Foreign.h>
+
+typedef unsigned char (^GetByteFromUno)(int);
+
+typedef struct {
+	GetByteFromUno readByte;
+} UnoCallbackBox;
+
+size_t getBytesAtPosition(void *info, void *buffer, off_t pos, size_t cnt);
```
kusma: This should go away with the suggestion above...

Source/Fuse.Elements/Resources/Exif/ExifHelper.mm
```diff
@@ -0,0 +1,13 @@
...
+	UnoCallbackBox* callback_box = (UnoCallbackBox*)info;
+	size_t bytes_read = 0;
+	for (int i = 0; i < cnt; i++) {
+		GetByteFromUno reader = callback_box->readByte;
+		char b  = reader(pos + i);
+		((char*)buffer)[i] = b;
+		bytes_read += 1;
+	}
+	return bytes_read;
+}
```
kusma: This should go away with the suggestion above...

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,242 @@
...
+		Undefined,
+		Normal,
+		Rotate90,
+		Rotate180,
+		Rotate270,
+		Transpose,
+		Transverse,
+		FlipHorizontal,
+		FlipVertical
+	}
```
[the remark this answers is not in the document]
kristianhasselknippe: I was trying to reflect the exif spec here. Doesn't this go against that (perhaps not a bad thing?)?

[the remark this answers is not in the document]
kusma: The spec doesn't use these definitions. It defines it in terms of memory-storage, not transformations.

Source/Fuse.Elements/Resources/Exif/Exif.uno
kusma: I would say that we try to prevent them from happening in the first place.

I'm guessing you're getting compile-errors from Java because of missing exception-handlers here? If that's the whole deal, then just wrap the whole thing in `try { .... } catch (Exception e) { e.printStackTrace(); }`.

If you actually want to convey some information, you might want to pass an onFailed-Action in so you can report things back in a clean way.

Source/Fuse.Elements/Resources/Exif/Exif.uno
kristianhasselknippe: can't find this on the fusetools remote. Should i just copy the whole method?

kusma: Yeah, it's not merged yet. You could just inline it into your function, if you want.

Source/Fuse.Elements/Resources/Exif/Exif.uno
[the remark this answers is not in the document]
kusma: Hm, thinking about it a bit more, it seems like some of those ImageOrientation-modes that I suggest give the same result. `FlipVertical` gives the same result as `Rotate180 | FlipHorizontal`, for instance.  So if we go this path, I think we should only have one of those flip-flags, so it's possible to compare orientation and get what you expect.

Source/Fuse.Elements/Resources/Exif/Exif.uno
kusma: If you use `Directory.getInteger` rather than `Directory.getInt`, you won't have to have the catch. 

So, I guess you could write it like this to make it (IMO) a bit clearer?
```java
InputStream s = (InputStream)stream;
try {
	Metadata metadata = ImageMetadataReader.readMetadata(s);

	ExifIFD0Directory directory = metadata.getFirstDirectoryOfType(ExifIFD0Directory.class);
	if (directory != null) {
		Integer orientation = directory.getInteger(ExifIFD0Directory.TAG_ORIENTATION);
		if (orientation != null)
			return (int)orientation;
	}
} catch (Exception e) {
	e.printStackTrace();
}
return -1;
```

Source/Fuse.Elements/Resources/Exif/Exif.uno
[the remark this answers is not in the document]
kristianhasselknippe: Good point. Should i still put this enum in the Exif namespace btw?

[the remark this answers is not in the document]
kusma: If we call it ImageOrientation, then I think it should be in the Fuse.Resources namespace, but I don't really have strong feelings here.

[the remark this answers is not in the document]
kristianhasselknippe: We also have to define which operation comes first, right? Flip or Rotate...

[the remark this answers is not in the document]
kusma: Yeah. It seems to me like y-flipping first will give the easiest pattern to deal with.

Source/Fuse.Elements/Resources/Exif/Tests/OrientationTest/Exif.OrientationTest.uno
```diff
@@ -0,0 +1,104 @@
+using Uno.Testing;
+using Fuse.Resources.Exif;
+
+public class OrientationTest
+{
+	public void AssertImageOrientation(byte[] bytes, ImageOrientation expected)
+	{
+		var ori = ExifData.FromByteArray(bytes).Orientation;
+		Assert.AreEqual(expected, ori);
+	}
```
kusma: First off, please make `expected` the first argument here, this is what all other Assert-methods do.

Secondly, if this fails, it's going to fail with the location being inside this call, which isn't particularly helpful. If you add `[CallerFilePath] string filePath = "", [CallerLineNumber] int lineNumber = 0, [CallerMemberName] string memberName = ""` to the signature, and pass these to Assert.AreEqual as well, the correct location should be blamed instead.

Third, it seems to me like all call-sites call BundleFile.ReadAllBytes() as the input here. Maybe we could just make this helper take BundleFile as an argument instead? If so, the call-sites would read:

```uno
AssertImageOrientation(ImageOrientation.Identity, import("bird0.jpg"));
```
...which I think is pretty neat!

Lastly (and this is extremely unimportant at this point), but this method could also be marked as static ;)

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,190 @@
...
+				return ExifIOSImpl.FromByteArray(buffer);
+			if defined(DotNet)
+				return ExifDotNetImpl.FromByteArray(buffer);
+			else
+				throw new Exception("Exif is not supported by this platform");
+		}
+
+		internal ExifData(int orientation)
+		{
+			//This is based on the vertical flip being applied before the rotation
```
kusma: I think this would be better written as proper documentation on the ImageOrientation enum instead, no?

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,190 @@
...
+
+	extern(CIL) static class ExifDotNetImpl
+	{
+		const int OrientationTagId = 274; //as defined by the exif spec www.exif.org/Exif2-2.PDF (page 16)
+
+		internal static ExifData FromByteArray(byte[] bytes)
+		{
+			var memStream = new MemoryStream(bytes);
+			var img = Image.FromStream(memStream);
+			return new ExifData(GetOrientation(img));
```
kusma: I guess `return new ExifData(GetOrientation(Image.FromStream(new MemoryStream(bytes))));` got a bit too hard to read? (I lost track of the parens myself while writing it :P)

How about 
```
var image = Image.FromStream(new MemoryStream(bytes));
return new ExifData(GetOrientation(image));
```

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,190 @@
...
+		[Foreign(Language.Java)]
+		static int GetOrientation(Java.Object buf)
+		@{
+			InputStream s = new com.fuse.android.ByteBufferInputStream((com.uno.UnoBackedByteBuffer)buf);
+			try {
+				Metadata metadata = ImageMetadataReader.readMetadata(s);
+				ExifIFD0Directory directory = metadata.getFirstDirectoryOfType(ExifIFD0Directory.class);
+
+				if (directory != null && directory.containsTag(ExifIFD0Directory.TAG_ORIENTATION)) {
+					return directory.getInt(ExifIFD0Directory.TAG_ORIENTATION);
```
kusma: I think this still risks throwing in the case where `ExifIFD0Directory.TAG_ORIENTATION` isn't an int. But I guess that's such a bizarre case, that it's not worth worrying about. We'd just let it throw. In other words, this looks like a nice compromise.

kusma:
Sadly, this broke building Android-tests, so it had to be reverted :(

PR #529, opened 2017-10-04
Native TimePickers

yupferris:
Introduces the `Fuse.Controls.TimePicker` package, which includes a `TimePicker` class that proxies native views on Android and iOS. On desktop, a simple placeholder is shown.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.TimePicker/TimePicker.Docs.uno
+++ b/Source/Fuse.Controls.TimePicker/TimePicker.Docs.uno
@@ -1,0 +1,60 @@
+namespace Fuse.Controls
+{
+	/**
+		Displays a component to select a time.
+
+		Currently, the TimePicker only has native implementations, so it should be contained in a @NativeViewHost.
+
+		A `TimePicker` can be used to select a specific time value. The type of its `Value` property is `Uno.DateTime`,
+		which is marshalled automatically to and from the JavaScript `Date` type. This makes interaction between JavaScript
+		and the `TimePicker` type seamless via databinding. If you plan to wrap a `TimePicker` in a UX component and use a
+		UX property to hook up this value, the `Uno.DateTime` type should be used.
+
+		Both `Uno.DateTime` and JS' `Date` type represent a specific timestamp. These types have both date and time
+		components, and their interpretation depends on a given time zone, which can cause a great deal of confusion. To
+		simplify all of this and ensure consistent behavior accross different time zones and locales, `TimePicker` will assume
+		incoming values are relative to UTC, and truncate the date component to the Unix epoch (1 Jan 1970), effectively
+		ignoring the date component altogether. Similarly, values read from `TimePicker` properties will only consist of a time
+		component at on 1 Jan 1970. This makes values going to/from the `TimePicker` control easy to create and interpret
+		consistently, but also means that if a value with a date component other than the unix epoch is written to TimePicker`'s
+		`Value` property, subsequent values read from the property may not match the written value, as the date component will
+		have been truncated.
+
+		You should avoid modifying the `TimePicker` values programmatically while the control has focus, as this is known to
+		have some issues on some Android devices (particularly ones which use the new `clock` appearance prior to Android 7).
+
+		## Example
+
+		The following example shows how to set up a `TimePicker` object and set the value from JS using a `Date` object:
+
+			<StackPanel>
+				<JavaScript>
+					var Observable = require("FuseJS/Observable");
+
+					var someTime = Observable(new Date(Date.parse("2007-02-14T12:34:56.000Z")));
+
+					someTime.onValueChanged(module, function(date) {
+						console.log("someTime changed: " + JSON.stringify(date));
+					});
+
+					module.exports = {
+						someTime: someTime,
+
+						timeToGetCracking: function() {
+							someTime.value = new Date(Date.parse("1970-01-01T13:37:00.000Z"));
+						}
+					};
+				</JavaScript>
+
+				<NativeViewHost>
+					<TimePicker Value="{someTime}" Is24HourTimeView="true" />
+				</NativeViewHost>
+
+				<Button Text="Time to get cracking!" Clicked="{timeToGetCracking}" Margin="5" />
+			</StackPanel>
+
+	*/
+	public partial class TimePicker
+	{
+	}
+}
--- a/Source/Fuse.Controls.TimePicker/TimePicker.uno
+++ b/Source/Fuse.Controls.TimePicker/TimePicker.uno
@@ -1,0 +1,126 @@
+using Uno;
+using Uno.UX;
+using Uno.Compiler.ExportTargetInterop;
+
+using Fuse;
+using Fuse.Controls.Native;
+using Fuse.Scripting;
+
+namespace Fuse.Controls
+{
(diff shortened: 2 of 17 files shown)


Source/Fuse.Controls.TimePicker/TimePicker.Docs.uno
```diff
@@ -0,0 +1,67 @@
...
+		components, and their interpretation depends on a given time zone, which can cause a great deal of confusion. To
+		simplify all of this and ensure consistent behavior accross different time zones and locales, `TimePicker` will assume
+		incoming values are relative to UTC, and truncate the date component to the Unix epoch (1 Jan 1970), effectively
+		ignoring the date component altogether. Similarly, values read from `TimePicker` properties will only consist of a time
+		component at on 1 Jan 1970. This makes values going to/from the `TimePicker` control easy to create and interpret
+		consistently, but also means that if a value with a date component other than the unix epoch is written to TimePicker`'s
+		`Value` property, subsequent values read from the property may not match the written value, as the date component will
+		have been truncated.
+
+		Note that this control should not be used to deal with historic dates, as calendar/date and timestamp translation is
```
mortoray: I thought this just selected time, not date?

Source/Fuse.Controls.TimePicker/TimePicker.Docs.uno
```diff
@@ -0,0 +1,67 @@
...
+		inconsistent between different locales due to when and where different calendar systems were adopted. However, the
+		behavior is consistent for all dates since at least 1900 including all representable future dates.
+
+		Certain Android versions (prior to Android 7) which use the clock view by default appear to have some issues with
+		setting the hour/minute of the `TimePicker` programmatically while interacting with the picker, so some care
```
mortoray: I think the wording should be changed to say something like, "You should avoid modifying the TimePicker values programmatically while the control has focus as this is know to have issues on some Android devices."  It just explicitly says what not to do.

Source/Fuse.Controls.TimePicker/TimePicker.ScriptClass.uno
```diff
@@ -0,0 +1,91 @@
...
+					_tp._inIs24HourView = value;
+				UpdateManager.PostAction(_tp.UpdateIs24HourView);
+			}
+			public Is24HourViewProperty(TimePickerBase timePicker) : base(TimePickerBase._is24HourViewName) { _tp = timePicker; }
+		}
+
+		static TimePickerBase()
+		{
+			ScriptClass.Register(typeof(TimePickerBase),
+				new ScriptProperty<TimePickerBase, DateTime>("value", GetValueProperty),
```
mortoray: Why do we need this JS interface?  Should not a binding on `Value="{jsVar}"` accomplish the same thing?

Source/Fuse.Controls.TimePicker/TimePicker.ScriptClass.uno
```diff
@@ -0,0 +1,91 @@
...
+				new ScriptProperty<TimePickerBase, bool>("is24HourView", GetIs24HourViewProperty));
```
mortoray: Same with this, we usually don't expose config variables on the JS interface as there is no need. The user can do `<TimePicker Is24HourView="{jsBool}">` if they want.

Source/Fuse.Controls.TimePicker/TimePicker.uno
```diff
@@ -0,0 +1,155 @@
...
+
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+			_outValue = Value;
+			_inValue = Value;
+			_outIs24HourView = Is24HourView;
+			_inIs24HourView = Is24HourView;
+
+			UpdateManager.AddAction(PollViewValue);
```
mortoray: Do the underlying controls not emit an event? Polling isn't so nice.

yupferris: The native events are extremely finicky and esp. with the newer (now-default) material appearances, it's quite often they don't fire at all. This was also a problem with the date pickers. Note the comment here: https://github.com/fusetools/fuselibs-public/pull/529/commits/12505751811384f36358f7579f79a3bff9a29715#diff-3927c03f10105bd8d64ff0c6418ea586R26

Source/Fuse.Controls.TimePicker/TimePicker.ScriptClass.uno
yupferris: It's possible I misunderstood the purpose of having these scriptclass impl's; I assumed due to the existing scriptclass modules that we would provide an alternative JS-only API in addition to the uno one. @nekronos second opinion?

Source/Fuse.Controls.TimePicker/TimePicker.Docs.uno
yupferris: It only selects time, however it does so by using `DateTime` values, which inherently have date components as well. When implementing the date pickers this was a big problem, as different locales would have completely different interpretations of dates from a given timestamp due to when different calendars were adopted.

This may not actually be a problem with the time pickers specifically, but I figured we could be conservative here and copy this comment over as the same problems could affect the interpretation of time values due to DST or similar. But perhaps that's misguided; it's possible I could test more to see if it's a problem, although it's a bit difficult to test each possible scenario without buying some expensive plane tickets :)

Source/Fuse.Controls.TimePicker/TimePicker.Docs.uno
yupferris: Yeah, that's certainly more clear. I can change that.

Source/Fuse.Controls.TimePicker/TimePicker.ScriptClass.uno
nekronos: If `Value="{jsVar}` in UX is possible I would argue that we dont need a second entrypoint in the scriptclass for the same thing.

mortoray: What Vegard said. :)

Generally we've only added JS interfaces for things that can't be done via bindings, and we've never done it for "Value" fields before -- those bindings must work.

Source/Fuse.Controls.TimePicker/TimePicker.uno
mortoray: Shouldn't this be resolved on the Native picker itself then, not in the higher level class? The high level control should assume the implementation works, and workarounds are at that lower-level implementation.

This would also give the native control wrapper better control over when precisely it needs to poll for values (such as only while it's active or has focus).

Source/Fuse.Controls.TimePicker/TimePicker.ScriptClass.uno
yupferris: I'll go ahead and remove the scriptclass then :)

Source/Fuse.Controls.TimePicker/TimePicker.uno
yupferris: Sure, I can move the impl.

yupferris:
@mortoray I think I've addressed all of your concerns now; is there anything I missed?

PR #532, opened 2017-10-04
Fixed a typo in ImageTools getImageFromBuffer Docs

AlexeyBuzdin:
The declaration in uno code is as following:

`AddMember(new NativePromise<Image, Scripting.Object>("getImageFromBuffer", ImageFromBufferInterface, Image.Converter));`

So documentation is incorrect on the method name. Plus method returned string is wrapped in `ImagePromiseCallback` so result of the method is an `Image`


kusma:
@AlexeyBuzdin: Please don't send PRs from the main repo, but instead from your own fork. Yes, github has some helpers that makes it really easy to do the wrong thing, which isn't awesome.

PR #538, opened 2017-10-06
Update devtools

knatten:
New devtools. No user facing changes, so no docs update.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
@knatten: In the future, please don't send PRs directly from the fuselibs-public repo, but instead from a fork. Doing it from the repo incurs 2x the CI cost (and latency), and creates a situation where people forking the repo while the branch exists creates a duplicate of the branch, which gets rather messy over time.

knatten:
Thanks @kusma, I'll remember that in the future.

PR #544, opened 2017-10-09
Feature image use exif

kristianhasselknippe:
depends on #547

This PR adds exif support to Image. It is a work in progress, but i wanted to open this PR to get some initial feedback on the approach.

In short, i've added an ImageOrientation property to ImageSource. Its value is set when the various image loaders receive image data.

The value of this ImageOrientation is then used to calculate a transformation matrix which is applied to the texture coordinates when drawing the image.

The imageloader stuff will get cleaned / removed since Kusma has already done some work there.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [x] Tests


Source/Fuse.Controls.Primitives/Image.Visual.uno
```diff
@@ -140,20 +179,22 @@ namespace Fuse.Controls
...
 				apply Fuse.Drawing.Planar.Image;
+
 				DrawContext: dc;
 				Visual: element;
 				Size: size;
 				Position: offset;
 				Texture: tex;
 				SamplerState SamplerState: GetSamplerState(resampleMode);
-				TexCoord: VertexData * uvSize + uvPosition;
+				TexCoord: Vector.TransformCoordinate(VertexData * uvSize + uvPosition, imageTransform);
```
kusma: I think I would do:
```uno
TexCoord: VertexData * uvSize + uvPosition;
TexCoord: Vector.TransformCoordinate(prev, imageTransform);
```
...just to untangle these two steps.

Source/Fuse.Controls.Primitives/Image.Visual.uno
```diff
@@ -5,16 +5,32 @@ using Fuse.Drawing;
...
+			var pixelSize = Source.PixelSize;
+			if (!IgnoreExif
+				&& (Source.Orientation.HasFlag(ImageOrientation.Rotate90)
+					|| Source.Orientation.HasFlag(ImageOrientation.Rotate270)))
+			{
+				size = float2(Source.Size.Y, Source.Size.X);
+				pixelSize = int2(Source.PixelSize.Y, Source.PixelSize.X);
+			}
+			var ret = Container.Sizing.CalcContentSize( size, pixelSize );
+			debug_log("ImageSize: " + ret);
```
kusma: Please drop the debug_log...

Source/Fuse.Controls.Primitives/Image.Visual.uno
```diff
@@ -5,16 +5,32 @@ using Fuse.Drawing;
...
 using Fuse.Elements;
 using Fuse.Nodes;
+using Fuse.Resources;
+using Fuse.Resources.Exif;
 
 namespace Fuse.Controls
 {
 	public partial class Image
 	{
+		public bool IgnoreExif { get; set; }
```
kusma: What's the purpose of this?

Source/Fuse.Controls.Primitives/Image.Visual.uno
```diff
@@ -5,16 +5,32 @@ using Fuse.Drawing;
...
-				return Container.Sizing.CalcContentSize( Source.Size, Source.PixelSize );
-			return float2(0);
+			if (Source == null)
+				return float2(0);
+
+			var size = Source.Size;
+			var pixelSize = Source.PixelSize;
+			if (!IgnoreExif
+				&& (Source.Orientation.HasFlag(ImageOrientation.Rotate90)
+					|| Source.Orientation.HasFlag(ImageOrientation.Rotate270)))
```
kusma: This is indented strangely...

Source/Fuse.Controls.Primitives/Image.Visual.uno
```diff
@@ -83,16 +99,39 @@ namespace Fuse.Controls
...
+					else if ((Source.Orientation & (int)0x03) == ImageOrientation.Rotate90)
+						rotation = Matrix.RotationZ(-Math.PIf / 2.0f);
+					else if ((Source.Orientation & (int)0x03) == ImageOrientation.Rotate180)
+						rotation = Matrix.RotationZ(Math.PIf);
+
+				}
+
+				var translateToCenter = Matrix.Translation(0.5f,0.5f,0.0f);
+				var translateBack = Matrix.Translation(-0.5f,-0.5f,0.0f);
+				var imageTransform = Matrix.Mul(translateBack, flip, rotation, translateToCenter);
```
kusma: I think we should make a helper-method to compute the image-transform from the orientation...

Source/Fuse.Controls.Primitives/Image.Visual.uno
kusma: Hmm, I kind of expected this to get rid of Fuse.Drawing.Planar.Rectangle.Invert...

kristianhasselknippe: Do you know when/why Invert is currently used?

Source/Fuse.Controls.Primitives/Image.Visual.uno
kristianhasselknippe: for testing. haven't done a proper cleanup yet.

Source/Fuse.Controls.Primitives/Image.Visual.uno
kusma: @kristianhasselknippe: Only when using textures that come from images, AFAIK. It's basically a hard-coding of one of the orientations.

kusma: Or, the combination of image-orientation and the fact that we're using a confused OpenGL coordinate-system for images.

kristianhasselknippe: @kusma:
So it turns out Invert is used in a bunch of places like 
Controls.Panels\Panel.Freeze.uno(227) 
Drawing.Surface\Surface.uno(143)
Controls\Viewport.uno(273)
Effects\Mask.uno(104,122,140)
iOS.TextRenderer\TextRenderer.uno(355)

and all of the effect classes. 

Should we perhaps look into getting rid of Invert as a separate issue?

Source/Fuse.Elements/Resources/HttpImageSource.uno
```diff
@@ -143,7 +152,13 @@ namespace Fuse.Resources
 			else
 				_contentType = ct;
 
-			new BackgroundLoad(data, _contentType, SuccessCallback, FailureCallback);
+			try
+			{
+			_orientation = ExifData.FromByteArray(data).Orientation;
+			}
+			//If Exif is not supported we just stick with the default (ImageOrientation.Identity)
+			catch (Exception e) { }
```
kusma: The indentation here seems wonky. And we'll leak the orientation of the previously loaded image here, if the image got loaded, switched on the server to a differently oriented version, and then reloaded (there's a few ways this can happen). So as-is, this isn't robust. We need to assign _orientation in all cases.

Also, since this is needed in multiple places, how about making a helper for it?

Something like:
```uno
ImageOrientation GetImageOrientationFromByteArrayOrDefault()
{
	try
	{
		return ExifData.FromByteArray(data).Orientation;
	}
	catch (Exception e)
	{
		return ImageOrientation.Identity;
	}
}
```

But thinking about it a bit more, it feels pretty wasteful to even raise an exception if all we're going to do is ignore the exception. So perhaps we should have a non-throwing API here instead? Something like `static bool TryFromByteArray(byte[] data, out ExifData exifData);`? That would make the helper 

```uno
ImageOrientation GetImageOrientationFromByteArrayOrDefault()
{
	ImageOrientation orientation;
	if (!ExifData.TryFromByteArray(data, out orientation))
		return ImageOrientation.Identity;

	return orientation;
}
```

kristianhasselknippe: Hmm, i realize that ImageOrientation could just default to Identity, so do we really need the bool? I guess it will be more relevant if we add support for more exif tags.

kusma:
@kristianhasselknippe: I'd prefer if we could do the Buffer vs byte[] as a preparatory step. But since cleaning that up is a bunch of work, I went ahead and [did it for you](https://github.com/kusma/fuselibs-public/tree/experimental-http-cleanups). Do you mind if inject these patches into your series?

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,196 @@
...
+		{
+			if defined(Android)
+				data = ExifAndroidImpl.FromByteArray(buffer);
+			if defined(iOS)
+				data = ExifIOSImpl.FromByteArray(buffer);
+			if defined(DotNet)
+				data = ExifDotNetImpl.FromByteArray(buffer);
+			else
+			{
+				data = new ExifData(0);
```
kusma: `data = default(ExifData);` instead?

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,196 @@
...
+				return false;
+			}
+			return true;
+		}
+
+		internal ExifData(int orientation)
+		{
+			switch (orientation)
+			{
+				case 0: Orientation = ImageOrientation.Identity; break; //Exif orientation is undefined
```
kusma: Have you ever seen this value in a file? If you do the suggestion above, I doubt we'd need this case...

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,196 @@
...
+				case 1: Orientation = ImageOrientation.Identity; break;
+				case 2: Orientation = ImageOrientation.FlipVertical | ImageOrientation.Rotate180; break;
+				case 3: Orientation = ImageOrientation.Rotate180; break;
+				case 4: Orientation = ImageOrientation.FlipVertical; break;
+				case 5: Orientation = ImageOrientation.FlipVertical | ImageOrientation.Rotate270; break;
+				case 6: Orientation = ImageOrientation.Rotate90; break;
+				case 7: Orientation = ImageOrientation.FlipVertical | ImageOrientation.Rotate90; break;
+				case 8: Orientation = ImageOrientation.Rotate270; break;
```
kusma:
```uno
default:
	throw new ArgumentOutOfRangeException(nameof(orientation));
```

CHANGELOG.md
```diff
@@ -23,6 +23,8 @@
  * The `Fuse.Reactive.IWriteable` interface has changed (breaking!). The method signature is now `bool TrySetExclusive(object)` instead of `void SetExclusive(object)`. Unlikely to affect your code.
  * `IObservable` and `IObservableArray` no longer push their initial value on `Subscribe`.
 
+## Image
+- Image will now respect exif orientation.
```
kusma: Micro-nit: The official name is "Exif", with an upper-case E.

Source/Fuse.Elements/Resources/Exif/Exif.uno
kusma: Oh. These changes are already merged...

kristianhasselknippe:
@kusma : About the other cases like ImageFill and effects that use images. Could those be handled by a separate PR? The changelog only claims that `Image` now respects Exif. What do you think?

kusma:
@kristianhasselknippe: I think that would be OK, as long as we don't end up breaking anything. And I don't think we currently do!

kusma:
Oh, wait. We might actually break something. Non-square ImageFills with non-identity Exif orientation will probably get squished by this, because they will report a Exif-corrected size, but does not take it into account while rendering...

I mean, it was already rendering incorrectly, so I'm not sure if having a "different incorrect rendering" is the worst thing imaginable...

Source/Fuse.Elements/Resources/Exif/Exif.uno
kristianhasselknippe: i'll do this and the one below as a separate PR

kristianhasselknippe:
The Size correction currently happens in Image.Visual, so it shouldn't affect anything related to ImageFill.

kusma:
@kristianhasselknippe: Right. In the long term, I think it's better if ImageSource does this bit, because it's used by more parts of our system. But for now, I think that's fine.

kristianhasselknippe:
not sure why router tests fail :S

kusma:
@kristianhasselknippe: Probably a spurious failure, we should really file a bug and disable tests like that... But it'll get a new run now.

kusma:
@kristianhasselknippe: Whups. Would you mind squashing that last one into 1098625f541197a660b3692aff35fc8a35d59519?

Source/Fuse.Elements/Resources/MultiDensityImageSource.uno
```diff
@@ -93,6 +94,18 @@ namespace Fuse.Resources
...
 
+		public override ImageOrientation Orientation
+		{
+			get
+			{
+				if (_active != null)
+				{
+					return _active.Orientation;
+				}
+				return ImageOrientation.Identity;
```
kusma: Why isn't this just `return _proxy.Orientation;`, similar to `Size` and `PixelSize`?

kristianhasselknippe: Just because I didn't think it through well enough I think. I'll change it to match the way Size and PixelSize works.

PR #545, opened 2017-10-09
Remove DatePicker scriptclass

yupferris:
Since we can databind, we don't need an alternative JS API for this functionality.


mortoray:
These weren't released yet, were they?

yupferris:
nope

PR #546, opened 2017-10-09
fix a regression with not updating items

mortoray:
A change in Each.OnItemsChanged breaks an expected behaviour in Charting. The fix appears to be providing a more complete dervied API to resolve a rooting issue. Premiumlibs will need to use this new `SetItemsDerivedRooting` function.

This PR contains:
- [ ] ~Changelog~
- [ ] ~Documentation~
- [x] Tests

### DIFF
--- a/Source/Fuse.Reactive.Bindings/Each.uno
+++ b/Source/Fuse.Reactive.Bindings/Each.uno
@@ -158,14 +158,11 @@ namespace Fuse.Reactive
 		*/
 		public object Items
 		{
-			get { return _items; }
+			get { return GetItems(); }
 			set
 			{
-				if (_items != value)
-				{
-					_items = value;
-					OnItemsChanged();
-				}
+				if (GetItems() != value)
+					SetItems( value );
 			}
 		}
 
--- a/Source/Fuse.Reactive.Bindings/Instance.API.uno
+++ b/Source/Fuse.Reactive.Bindings/Instance.API.uno
@@ -296,11 +296,28 @@ namespace Fuse.Reactive
 		
 		internal bool HasLimit { get { return _hasLimit; } }
 
-		protected internal object _items;
+		object _items;
+		/** @hide */
+		protected object GetItems() { return _items; }
+		/** @hide */
+		protected void SetItems( object value )
+		{
+			_items = value;
+			if (!IsRootingCompleted) return;
+			RefreshItems();
+		}
+		/** Call to set the items during the OnRooted override (post base.OnRooted call)
+			@hide */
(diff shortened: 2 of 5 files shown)


Source/Fuse.Reactive.Bindings/Instance.API.uno
```diff
@@ -296,11 +296,28 @@ namespace Fuse.Reactive
...
+		/** @hide */
+		protected void SetItems( object value )
+		{
+			_items = value;
+			if (!IsRootingCompleted) return;
+			RefreshItems();
+		}
+		/** Call to set the items during the OnRooted override (post base.OnRooted call)
+			@hide */
+		protected void SetItemsDerivedRooting( object value )
```
kusma: Typo, should be `SetItemsDuringRooting`, no?

kusma: Typo, should be `SetItemsDuringRooting`, no?

Source/Fuse.Reactive.Bindings/Each.uno
```diff
@@ -158,14 +158,11 @@ namespace Fuse.Reactive
 		*/
 		public object Items
 		{
-			get { return _items; }
+			get { return GetItems(); }
```
kusma: Is this getter really an improvement? It just seems like churn to me...

Source/Fuse.Reactive.Bindings/Instance.API.uno
mortoray: I actually used that before, but then decided the empahsis here is more on the "Derived" part, since other things can actually change the items during rooting, but only the derived class should call this.

Source/Fuse.Reactive.Bindings/Each.uno
mortoray: It's because I intended on removing `protected` and `internal` from `_items` to prevent direct access. Somehow that change got lost. I'll that back in -- the error is essentially caused by incorrect access to `_items`, that's why I'm hiding it now.

mortoray: Wait, wrong code, yes, I did make that change in the base class already.

mortoray: So no, it's not code churn, `return _items` can't work anymore.

mortoray:
The AppVeyor failure seems random an unrelated (but I can't run the test again there)
Recorded in https://github.com/fusetools/fuselibs-public/issues/554

kusma:
@mortoray: Thanks for clarifying :)

PR #547, opened 2017-10-09
Feature exif

kristianhasselknippe:
This PR adds exif orientation from byte[]. It is implemented for android, ios and dotnet. The android implementation depends on a third party libarary https://drewnoakes.com/code/exif/

There is a bug in Uno which causes the java code in this pr to be compiled even though it should have been stripped. In this case, the third party library dependencies haven't been satisfied, which causes compiler errors. This happens while running other tests than the one added in this PR.

This PR therefore depends on Uno issue https://github.com/fusetools/uno/issues/1368 to get fixed.

Since the commits in this PR were merged and then reverted, I had trouble rebasing onto master. I fixed this by renaming the first commit, which i guess gave all the other commits new shas as well.? Was this an OK approach?

Update https://docs.google.com/spreadsheets/d/1mYt9jve20DH2Zaxa03l8GSqZuMcqC-cUZoUKyvegxWM/edit#gid=0 when this is merged

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests


kristianhasselknippe:
So, i've been trying to test this locally for android by running `uno test -tandroid -run-local`. This caused problems because the file paths become too long, so i created a temp directory in C:\ and ran the following:
`uno test -tandroid --run-local --output-dir=C:\testDir Source\Fuse.Elements`. This time the first test run fine, but then the second test complains with something like the following:

```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':app:externalNativeBuildDebug'.
> Build command failed.
Error while executing 'C:\Users\krist\AppData\Local\Android\sdk\cmake\3.6.4111459\bin\cmake.exe' with arguments {--build C:\testDir\app\.externalNativeBuild\cmake\debug\armeabi-v7a --target ExifOrientationTest}
[1/1] Re-running CMake...
-- Configuring done
-- Generating done
-- Build files have been written to: C:/testDir/app/.externalNativeBuild/cmake/debug/armeabi-v7a
ninja: error: unknown target 'ExifOrientationTest'
```
.

However if i run this test directly:
`uno test -tandroid --run-local --output-dir=C:\testDir Source\Fuse.Elements`
the test builds fine (but i do get some tests that fail).

Could this be related to my use of `--output-dir=C:\testDir`, and possible something not getting cleaned up correctly?

Thoughts on how to approach this?

kristianhasselknippe:
☝️ turned out not to be an actual issue. Should be ready to merge now

Source/Fuse.Elements/Resources/Exif/Exif.uno
```diff
@@ -0,0 +1,192 @@
...
+		public static ExifData FromByteArray(byte[] buffer)
+		{
+			if defined(Android)
+				return ExifAndroidImpl.FromByteArray(buffer);
+			if defined(iOS)
+				return ExifIOSImpl.FromByteArray(buffer);
+			if defined(DotNet)
+				return ExifDotNetImpl.FromByteArray(buffer);
+			else
+				throw new Exception("Exif is not supported by this platform");
```
kusma: From reviewing the call-site that uses this, I'm starting to think that this shouldn't throw, but rather return `bool` and have an out-argument, so we don't need to deal with exceptions just to be able to find out there's no EXIF information...

PR #548, opened 2017-10-10
Fix ManualTestApp/issues/371

kristianhasselknippe:
https://github.com/fusetools/fuselibs-public/pull/506 changed the behavior of video slightly to go directly to the "paused" state when the video was ready to play. The expected behavior was neither "playing" nor "stopped", so I added an Undefined option to cover the last case. Should this perhaps be renamed to "Ready" or "Initialized"? This is simply to represent the state where the user hasn't done anything yet.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Video/VideoVisual.uno
+++ b/Source/Fuse.Controls.Video/VideoVisual.uno
@@ -57,12 +57,13 @@ namespace Fuse.Controls.VideoImpl
 
 		enum PlaybackTarget
 		{
+			Undefined,
 			Playing,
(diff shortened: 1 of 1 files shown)


kusma:
@kristianhasselknippe: What do you think?

kristianhasselknippe:
> On second thought, I think we should set _playbackTarget back to PlaybackTarget.Undefined after we've applied it.
@kusma yeah that makes sense to me.

PR #555, opened 2017-10-10
Fix min/max date for Android DatePicker

yupferris:
Previously, we had missed a crucial detail: local time zone offset is of course relative to a specific timestamp, as offset changes over time. Luckily `java.util.TimeZone` gives us a way to get the default timezone's offset easily. All of this is necessary as `android.widget.setMinDate/setMaxDate` take time in the device's default timezone.

Fixes https://github.com/fusetools/ManualTestApp/issues/373.


yupferris:
forced push a small typo fix ("subtracing" -> "subtracting")

PR #557, opened 2017-10-10
Javascript fixes

kusma:
This fixes two JS issues. The first, and by far most important one, is that we currently call JavaScript code from the UI-thread.

This also happens on release-1.3 (and at least also on release-1.2), so it's not a recent regression.

The second issue is ES6-usage in the tests.

Both issues were triggered by running tests on JavaScriptCore.

This PR contains:
- [ ] Changelog
- [ ] ~Documentation~ Bugfix only
- [ ] Tests


kusma:
This seems to be worse than I thought, as for instance `Context.NewArray` gets called by Context.Unwrap, which we use from the UI-thread all the time :(

Source/Fuse.Reactive.JavaScript/ModuleInstance.uno
```diff
@@ -116,4 +122,4 @@ namespace Fuse.Reactive
 			}
 		}
 	}
-}
```
yupferris: Nit: unnecessary whitespace change

Source/Fuse.Reactive.JavaScript/Tests/UX/DataBinding.BehaviorDataContext.ux
```diff
@@ -45,4 +45,4 @@
 		</PageControl>
 	</DockPanel>
 
-</Panel>
```
yupferris: Nit: unnecessary whitespace change

Source/Fuse.Reactive.JavaScript/Tests/UX/DataBinding.BindingDirections.ux
```diff
@@ -35,4 +40,4 @@
 	<Text ux:Name="mirror_c" Value="{c}" />
 
 	<FuseTest.Invoke Handler="{changeThingsUp}" ux:Name="changeThingsUp"/>
-</Panel>
```
yupferris: Nit: unnecessary whitespace change

Source/Fuse.Reactive.JavaScript/Tests/UX/Observable.InnerDetach.ux
```diff
@@ -46,4 +46,4 @@
 		<FuseTest.Invoke Handler="{step1}" ux:Name="Step1"/>
 		<Text Value="{d}" ux:Name="D"/>
 	</Panel>
-</Panel>
```
yupferris: Nit: unnecessary whitespace change

Source/Fuse.Reactive.JavaScript/Observable.uno
```diff
@@ -85,19 +119,31 @@ namespace Fuse.Reactive
 					Fuse.Diagnostics.InternalError( "Unexpected null object", this );
 					return;
 				}
-				
-				try
+
+				var op = new SetExclusiveOperation(_om._worker, _om.Object, newValue, _origin);
+				if (_om._worker.CanEvaluate)
```
yupferris: I like this pattern, but it seems to be repeated in a few places - any way we can get some reuse/abstraction here perhaps? Seems like this would be easy to break in the future.

Source/Fuse.Reactive.JavaScript/ModuleInstance.uno
kusma: It just makes sure the file has a trailing new-line, which is required by the unix-standard to be considered a text-file.

I usually just fix these as I encounter them, but I can separate them out into a separate commit if that helps?

Source/Fuse.Reactive.JavaScript/Observable.uno
kusma: It's not quite clear to me what you mean. If you think there's too much repetition here in checking CanEvaluate before dispatching, I kinda agree, but there's a kicker: According to @sebbert, we don't really need all three, as ReplaceAllExclusive() can accurately express all three cases anyway. But changing that is a separate change that I'd rather not have to depend on in order to restore thread-sanity. So while I agree, I'm not sure it's worth the effort, because I expect a lot of this will go away.

What I really dislike here, is not having any safety-net to catch cases like this. I kind of want to refactor things a bit to make it API-wise impossible to call JS from the UI thread, but again that's a bigger change.

yupferris: That is indeed what I meant, that it's repetitive and I'm worried that it will be easy to break one of them in the future. But I think dealing with that in a separate change makes perfect sense 👍

Source/Fuse.Reactive.JavaScript/ModuleInstance.uno
yupferris: Right, ok. I think it's fine as-is, but it looks like you already added them as separate commits, which I do think is a bit cleaner :)

Source/Fuse.Reactive.JavaScript/Observable.uno
mortoray: I'm a bit surprised that this is ever called from the JavaScript thread. These are meant as Uno APIs to update JS. They could probably just always dispatch to the other thread, since I don't know that they have another appropriate use-case.

mortoray:
I wonder if these errors could be responsible for https://github.com/fusetools/fuselibs-public/issues/565

kusma:
@mortoray: That doesn't sound entirely unlikely, yeah...

Issue #560, opened 2017-10-11
NullReferenceException during Reify with ScrollViewPager


mortoray:
```
Build completed in 13.50 seconds
[Viewport]: System.NullReferenceException: Object reference not set to an instance of an object
  at Fuse.Controls.ScrollViewPager.CheckSizing () <0x120eed920 + 0x00040> in <filename unknown>:0 
  at Fuse.UpdateListener.Invoke () <0x11e1d54c0 + 0x0006b> in <filename unknown>:0 
  at Fuse.UpdateManager.ProcessOnces (Fuse.Stage stage, System.Collections.Generic.List`1& _exceptions) <0x11e1d4bb0 + 0x00189> in <filename unknown>:0 
[Viewport]: System.NullReferenceException: Object reference not set to an instance of an object
  at Fuse.Controls.ScrollViewPager.CheckSizing () <0x120eed920 + 0x00040> in <filename unknown>:0 
  at Fuse.UpdateListener.Invoke () <0x11e1d54c0 + 0x0006b> in <filename unknown>:0 
  at Fuse.UpdateManager.ProcessOnces (Fuse.Stage stage, System.Collections.Generic.List`1& _exceptions) <0x11e1d4bb0 + 0x00189> in <filename unknown>:0 
```

PR #561, opened 2017-10-11
fix NullReferenceError in ScrollViewPager

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/560

This PR contains:
- [x] Changelog
- [ ] ~Documentation~ no API changes
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## ScrollViewPager
+- Fixed a NullReferenceError that could happen while using ScrollViewPager in preview
+
 ## DatePicker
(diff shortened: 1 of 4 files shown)


mortoray:
If the ScrollViewPager was released in 1.3 then it is a problem there as well. It's not likely used often enough yet though to warrant a back-port.

kusma:
@mortoray: Yeah, ScrollViewPAge is part of release-1.3. So perhaps we should consider back-porting?

mortoray:
The patch is simple enough that it'd be easy to backport -- I can't imagine any conflicts.

PR #567, opened 2017-10-11
Fix DatePicker bogus scroll

yupferris:
On older versions of Android, layout changes can cause the calendar part of the DatePicker's
view to scroll to a bogus month, which both makes the view look like it has the wrong value,
and also makes it hard to pick a relevant value in many cases. To get around this, we need
to somehow get the control to scroll to show the currently set value whenever layout changes.
To achieve this, we'll reset a counter to zero on layout change, count that up a couple times
in order to wait at least 1 frame (we need more than just one iteration to avoid potential
update manager ordering issues). Then, we'll grab the picker's current value and write it
back, tricking the picker to invalidate and scroll where we want it. The downside of this
approach is that _any_ change in layout that causes this view to change size will scroll
the picker view to the currently selected value, but this is a decent tradeoff.

Also refactor some DatePicker interface code to move view polling and
other knowledge into the Android-specific implementation and expose a
root/unroot interface, which is much more general and could be useful
for other things in the future.

Fixes https://github.com/fusetools/ManualTestApp/issues/372.

### DIFF
--- a/Source/Fuse.Controls.DatePicker/Android/DatePicker.uno
+++ b/Source/Fuse.Controls.DatePicker/Android/DatePicker.uno
@@ -53,13 +53,66 @@ namespace Fuse.Controls.Native.Android
 			_pollValueCache = Value;
 		}
 
-		public void PollViewValue()
+		void PollViewValue()
 		{
 			if (Value != _pollValueCache)
 			{
 				OnValueChanged();
 				UpdatePollValueCache();
 			}
+		}
+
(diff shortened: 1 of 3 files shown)


yupferris:
One issue we may have here is that this could probably only be done if the api level is lower than 24 (which is when the clock appearance was introduced and methods like `setCalendarViewShown` was deprecated). @nekronos @kusma @cbaggers do we have a good way to query the API level without calling into java? I suppose I could query that at root time and cache the value, for example.

PR #569, opened 2017-10-12
[Feature] Add navigation methods to WebView for use in FuseJS

DerMitch:
This patch exports four methods of WebViews to allow them to be controlled by FuseJS code. This allows the creation of UI elements or logic to build a mini-webbrowser, which can be sometimes useful in apps.

This PR contains:
- [X] Changelog
- [ ] Documentation
Are the code comments enough?
- [ ] Tests
If needed, some guidance would be appreciated as I'm pretty new to Fuse/Uno development.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,6 +40,9 @@
  * `IObservable` and `IObservableArray` no longer push their initial value on `Subscribe`.
 
 
+### WebView
+- Exported the methods goBack, goForward, reload and stop for use in FuseJS
+
 # 1.3
 
 ## 1.3.0
@@ -84,8 +87,6 @@
 
 ### WebView
 Fixed issue where custom URI schemes were matched too greedily in URLs, making for erroneously intercepted URL requests.
-
-- Exported the methods goBack, goForward, reload and stop for use in FuseJS
 
 ### Delay Push Notification Registration on iOS
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 - Fixed `RangeControl.RelativeValue` to properly update when bound in UX
 - Allowed `Minimum` to be less than `Maximum` on `RangeControl` making it easier to do left-to-right `100..0` ranges.
 
+### WebView
+- Exported the methods goBack, goForward, reload and stop for use in FuseJS
+
+# 1.2
 ## ScrollViewPager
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -80,6 +80,8 @@
 ### WebView
 Fixed issue where custom URI schemes were matched too greedily in URLs, making for erroneously intercepted URL requests.
 
+- Exported the methods goBack, goForward, reload and stop for use in FuseJS
+
```
kusma: This is in the wrong section of the changelog. This feature is not part of 1.3, which has been feature-frozen for a while.

Source/Fuse.Controls.WebView/WebView.ScriptClass.uno
```diff
@@ -47,6 +51,47 @@ namespace Fuse.Controls
...
+
+		/**
+			Stop loading the page.
+
+			@scriptmethod stop()
+		*/
+		static void stop(Context c, WebView view, object[] args)
+		{
+			view.Stop();
+		}
```
kusma: It seems to me like the other script-methods verify their arguments and raise a UserError if they're not the expected lenght. Any reason why these methods don't?

CHANGELOG.md
kusma: Just to be clear, this should be put in the "Unreleased"-section of the changelog. I like that you looked for a suitable heading, but sadly there isn't one that we can reuse for this :/

DerMitch:
Hey @kusma, thank you for the quick response.

I moved the section in the CHANGELOG.md and added argument validation. As the methods don't take any arguments, I didn't saw a reason at first to add further checks.

As 1.3 is in feature freeze, I assume this might be released in 1.4 then?

CHANGELOG.md
```diff
@@ -5,6 +5,10 @@
 - Fixed `RangeControl.RelativeValue` to properly update when bound in UX
 - Allowed `Minimum` to be less than `Maximum` on `RangeControl` making it easier to do left-to-right `100..0` ranges.
 
+### WebView
+- Exported the methods goBack, goForward, reload and stop for use in FuseJS
+
+# 1.2
```
kusma: I doubt you meant to add this heading...

DerMitch:
Oh sorry, I think something went wrong when rebasing from upstream master. 😞 

Patch should be clean now.

PR #579, opened 2017-10-13
hard-code FUSELIBS_NO_TOASTS on

kusma:
FUSELIBS_NO_TOASTS is a blast from the past, when we needed the different
versions of the tool with different error-handling using the same fuselibs
version.

But that need is long gone, and the tool always enables this switch. So
let's get rid of the disabled case, as it just complicates things.

### DIFF
--- a/Source/Fuse.Reactive.JavaScript/Tests/UXExpressionsTest.uno
+++ b/Source/Fuse.Reactive.JavaScript/Tests/UXExpressionsTest.uno
@@ -82,37 +82,6 @@ namespace Fuse.Reactive.Test
 				root.StepFrameJS();
 
 				Assert.AreEqual("Hello foo : FOO!", e.t1.Value);
-
-				if defined(FUSELIBS_NO_TOASTS)
-				{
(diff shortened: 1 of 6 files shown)


Source/Fuse.Reactive.JavaScript/Tests/UXExpressionsTest.uno
```diff
@@ -82,37 +82,6 @@ namespace Fuse.Reactive.Test
 				root.StepFrameJS();
 
 				Assert.AreEqual("Hello foo : FOO!", e.t1.Value);
-
-				if defined(FUSELIBS_NO_TOASTS)
```
mortoray: Why has this bit been removed? Shouldn't this always run now since the define would be true?

kusma: @mortoray: Read the commit message. This test with this define enabled has been broken since before 1.0. Since https://github.com/fusetools/fuselibs/commit/d2aa412e0bbb77bc961ffc79a98c9cd3787dabc7, to be accurate.

PR #589, opened 2017-10-16
Fuse.Common: fix FontListParser for Android O

kusma:
In Android O, FontVariationAxis was introduced to the fontlist format
in a way that wasn't backwards compatible with the old FontListParser.

Unfortunately, our copy has diverged quite a bit, and can't be replaced
with the new one, because the new one doesn't keep compatibility with
the old format.

So the best thing we can do for now, is to patch up the meat to match
the code better.

Sicne we don't really care about the FontVariationAxis-data yet, we
just ignore it to avoid depending on later SDK versions. We can revisit
this in the future.

This was not helped by #440, which just painted over the initial
problem instead of fixing it.

The change was adopted from the upstream, available [here](https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/master/graphics/java/android/graphics/FontListParser.java)

### DIFF
--- a/Source/Fuse.Common/Internal/FontListParser.java
+++ b/Source/Fuse.Common/Internal/FontListParser.java
@@ -29,6 +29,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.regex.Pattern;
 
 /**
  * Parser for font config files.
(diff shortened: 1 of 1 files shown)


Source/Fuse.Common/Internal/FontListParser.java
```diff
@@ -224,6 +219,29 @@ private static Family readFamily(XmlPullParser parser)
...
+    /** Matches leading and trailing XML whitespace. */
+    private static final Pattern FILENAME_WHITESPACE_PATTERN =
+            Pattern.compile("^[ \\n\\r\\t]+|[ \\n\\r\\t]+$");
+
+    private static Font readFont(XmlPullParser parser)
+            throws XmlPullParserException, IOException {
+        String indexStr = parser.getAttributeValue(null, "index");
+        int index = indexStr == null ? 0 : Integer.parseInt(indexStr);
+        String weightStr = parser.getAttributeValue(null, "weight");
+        int weight = weightStr == null ? 400 : Integer.parseInt(weightStr);
```
cbaggers: Could this be NormalWeight? or is it just coincidentally the same value?

kusma: Dunno. I just took this verbatim from [upstream](https://raw.githubusercontent.com/aosp-mirror/platform_frameworks_base/master/graphics/java/android/graphics/FontListParser.java), and I'd prefer to keep it as similar as that to avoid needless divergence...

cbaggers: nice then time to approve :)

Issue #452, opened 2017-09-14
UpdateManager Onces don't get processed if there is no frame invalidation


mortoray:
UpdateManager.Update is not called if nothing is invalidated (my guess). This can happen with events that get processed the next frame, in particualr `PerformNextFrame`.

Seen on the manual test app page for User Events. clicked "Contaacts" doesn't go back to the previous page. It doesn't "trigger" the change if you change pages in the higher level navigation though. Like something is stalled somehow.

This may affect more than just the Onces, at least NextFrame, and possibly deferred actions if they go into the next frame as well.

mortoray:
Hmm, in isolation this defect doesn't happen!

mortoray:
The issue isn't navigation, or even the trigger. Something is just stalling entirely. Thus I'm giving this critical since it's very unusual.

mortoray:
This is caused by `Update` not being called on DesktopApp. This is because there is nothing which is dirty:

```
		public override bool NeedsRedraw 
		{ 
			get { return DesktopRootViewport.IsDirty; } 
}
```

This would appear to be because `Update` is called by the `UnoGLControl.Render` function.  I'm not clear on what the correct fix is. The quick fix would be to simply have the above function return `true` if there are pending items in the UpdateManager.

mortoray:
This affects basically any page with a ScrollView as well.

kusma:
This seems to behave correctly in `uno build` (from code-inspection), so I assume the problem is with the tool.

mortoray:
This fails on `uno build` with CMAke backend on OSX.
The linked issue indicates it is likely a problem in Preview as well.

kusma:
Just for the sake of a paper-trail, I believe this issue was introduced by ed20a04e66ca57a60c86c384b8bf406cfa5448df.

PR #595, opened 2017-10-17
fix frameindex issue when not drawing

mortoray:
fix https://github.com/fusetools/fuselibs-public/issues/452

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [ ] ~Tests~ ManualTestApp on Desktop will catch this in at least one location

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## DesktopApp Updates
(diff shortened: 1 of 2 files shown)


kusma:
This has been like this since 1.2, so I think this needs a changelog-entry.

I'd also like to see the comment mention that this is there to match the IncreaseFrameIndex-call in Update, for the cases where it's not being called.

PR #600, opened 2017-10-17
Feature Tree Observable

nekronos:
Fixed history

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
So, if I understand the situation correctly, this is the same version of the PR that got merged, but broke integration-tests with the tool, and hung tests at startup for the MSVC-target.

nekronos:
@kusma yes, still need to resolve those problems. Had to fix some broken git history with https://github.com/fusetools/fuselibs-public/pull/601. This PR and https://github.com/fusetools/fuselibs-public/pull/601 is now properly in-sync

nekronos:
@kusma this is now building on TC https://tc.outracks.com/viewLog.html?buildId=698600&tab=queuedBuildOverviewTab
Should we try merge again? :)

kusma:
@nekronos: Yeah, but let's wait until 1.4 has been branched out. That's happening today.

PR #601, opened 2017-10-17
Feature Models

nekronos:
This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests


Source/Fuse.Scripting.JavaScript/TreeObservable/TreeObservable.uno
```diff
@@ -125,7 +125,7 @@ namespace Fuse.Reactive
 			void PerformStart()
 			{
 				Perform(TreeObservable, 0);
-			}
+			}
```
kusma: Please drop this

Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
```diff
@@ -34,8 +34,8 @@ namespace Fuse.Controls
 			
 			if (_pageHistory == null)
 				return;
-			
-			var obs = _pageHistory as IObservable;
+
```
kusma: Adding this tab seems pointless

Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
```diff
@@ -64,6 +64,21 @@ namespace Fuse.Controls
 			Replace = 1 << 2,
 		}
 
+		protected string GetObjectPath( object data )
+		{
+			string path = null;
```
kusma: This variable is unused.

Source/Fuse.Scripting.JavaScript/FuseJS/Timer.uno
```diff
@@ -135,14 +135,13 @@ namespace Fuse.Reactive.FuseJS
...
 		}
-		
-		internal void UpdateModule(Scripting.Context context)
+
+		internal bool UpdateModule(Scripting.Context context)
 		{
 			// NOTE: Don't use UpdateManager for this, for things to run smoothly, this needs to be a JS only thread thing.
-			if(_tm != null)
-				_tm.Tick(context);
+			return _tm != null ? _tm.Tick(context) : false;
```
kusma: In this case, I think
```uno
if (_tm != null)
	 return _tm.Tick(context);
return false;
```
(e.g the old code + return) would be clearer...

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+		class ModelData
+		{
+			public IExpression Model;
+			public NameTable NameTable;
+		}
+		
+		//Requires a NameTable that will be set first.
+        [UXAttachedPropertySetter("JavaScript.Model"), UXNameScope, UXAuxNameTable("ModelNameTable")]
+        public static void SetModel(Visual v, IExpression model)
+        {
```
kusma: Broken indentation...

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+				
+			var parsed = ModelJavaScript.ParseModelExpression(md.Model, md.NameTable);
+			v.Children.Add( new ModelJavaScript(parsed, md.NameTable, null) );
+        }
+
+        static PropertyHandle _modelHandle = Properties.CreateHandle();
+        
+        [UXAttachedPropertySetterAttribute("ModelNameTable")]
+        public static void SetModelNameTable(Visual v, NameTable nt)
+        {
```
kusma: Broken indentation. Whoever wrote this, fix your config. This keeps happening in this source-file. Fix all of it.

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+			}
+			else
+			{
+				md.NameTable = nt;
+			}
+			
+			Complete( md, v );
+        }
+
+        //TODO: This should probably be JavaScript.Model, Preview would need to be adjusted as well
```
kusma: What's up with this TODO? What implications does leaving this as-is have?

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+			public string ModuleName;
+			public string ClassName;
+			public List<string> Args = new List<string>();
+			public List<Dependency> Dependencies = new List<Dependency>();
+			
+			public string ArgString
+			{
+				get
+				{
+					var str = "";
```
kusma: `var str = string.Empty;`

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+					for (int i=0; i < Args.Count; ++i)
+						str += ", " + Args[i];
+					return str;
```
kusma: Actually, this should probably use `StringBuilder` instead.

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+					
+				return o.ModuleName == ModuleName &&
+					o.ClassName == ClassName;
+			}
+		}
+		
+		internal static ParsedModelExpression ParseModelExpression(IExpression exp, NameTable nt)
+        {
+            if (exp is Data) {
+                var className = ((Data)exp).Key;
```
kusma: First of all, please no Egyptian braces.

Second, please drop the double-cast, instead do:
```Uno
var data = exp as Data;
if (data !=null)
{
	var className = data.Key;
	return [...]
}

var divide = exp as Divide;
if (divide != null)
[...]
```

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+                return result;
+            }
+            else throw new Exception("Invalid Model path expression: " + exp);
+        }
+
+        void SetupModel()
+        {
+			if (_model == null)
+			{
+				Code = "";
```
kusma: `Code = string.Empty;`

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+		}
+		
+		ParsedModelExpression _model;
+		internal ModelJavaScript(ParsedModelExpression model, NameTable nt,
+			string previewStateId)
+			: base(nt)
+		{
+			_previewStateModelId = previewStateId;
+			_model = model;
+			FileName=  "(model-script)";
```
kusma: The spacing around the `=` is off.

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -40,16 +41,98 @@ namespace Fuse.Scripting.JavaScript
...
+				_value = value;
+			}
+
+			// UI thread
+			public void Perform()
+			{
+				// HACK: This should really check if the property has ever been set.
+				// However, this requires some bigger changes to the UX compiler,
+				// so we check for the default value of the UX Property's type instead.
+				// Note that this will fail when the initial value is explicitly set to the default value of the type.
```
kusma: How about we file a ticket for this, and leave a reference to that ticket here?

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
```diff
@@ -0,0 +1,581 @@
...
+				e.callIncr.Perform();
+				root.StepFrameJS();
+				Assert.AreEqual( 116, e.v.Value );
+				Assert.AreEqual( 16, e.r.Value );
+				Assert.AreEqual( 232, e.d.Value );
+			}
+		}
+
+		[Test]
+		//TODO: working on this this one still (found the Each one)
```
kusma: Sounds like this TODO should be finished up...

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
```diff
@@ -0,0 +1,581 @@
...
+
+		[Test]
+		public void ListOrder()
+		{
+			var e = new UX.Model.ListOrder();
+			using (var root = TestRootPanel.CreateWithChild(e))
+			{
+				root.StepFrameJS();
+				Assert.AreEqual( "2,3", e.oc.JoinValues() );
+				e.oc.Log.Clear(); //TODO: it's undecided what the bootstrapping messages will be still
```
kusma: Sounds like there's some more work to be done here. Either fix it, or file a ticket (and put a reference to the ticket here)...

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
```diff
@@ -0,0 +1,581 @@
...
+				e.resolveNow.Perform();
+				root.StepFrameJS();
+				Assert.AreEqual("hoho!", e.t.Value);
+			}
+		}
+
+		[Test]
+		public void EmptyList() {
+			var e = new UX.Model.EmptyList();
+			using (var root = TestRootPanel.CreateWithChild(e)) {
```
kusma: Don't do Egyptian. This applies elsewhere in this file as well.

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
```diff
@@ -0,0 +1,581 @@
...
+				Assert.OfType<UX.Model.MainPage>(e.navigator.Active);
+			}
+		}
+
+		List<T> ChildrenOfType<T>(Visual n) where T : Node
+		{
+			var l = new List<T>();
+			for (int i=0; i < n.Children.Count; ++i)
+			{
+				var m = n.Children[i] as T;
```
kusma: Please do:
```uno
foreach (var child in n.Children)
{
	var m = child as T;
	[...]
```

This way, we don't have to bend over backwards to support indexing, which isn't needed in this case.

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
```diff
@@ -0,0 +1,581 @@
...
+				if (m != null)
+					l.Add(m);
+			}
+			return l;
+		}
+	}
+}
+
+namespace Uno.Testing {
+	public static partial class Assert {
```
kusma: No, you can't add a partial definition to a class defined in a different package. If that compiles, that's a compiler-bug, not a feature.

This is a useful assert, so please send a PR for it in Uno, and have a temporary helper in the mean-time.

Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -133,12 +133,11 @@ namespace Fuse.Scripting.JavaScript
...
 					}
 				}
-				else
-					_idle.Set();
 
 				try
 				{
-					_fuseJS.UpdateModules(_context);
+					var activity = _fuseJS.UpdateModules(_context);
+					didAnything ||= activity;
```
kusma: [C# doesn't support `||=`](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/), so if Uno supports this, I'd consider that a compiler-bug. Please don't use it.

Source/Fuse.Scripting.JavaScript/FuseJS/Builtins.uno
```diff
@@ -73,10 +73,12 @@ namespace Fuse.Reactive.FuseJS
 			}
 		}
 
-		internal void UpdateModules(Fuse.Scripting.Context context)
+		internal bool UpdateModules(Fuse.Scripting.Context context)
 		{
+			var activity = false;
 			if(_timer != null)
-				_timer.UpdateModule(context);
+				activity = _timer.UpdateModule(context) || activity;
```
nekronos: why `|| activity`? `activity` is always false there

nekronos:
```uno
if (_timer != null)
    return _timer.UpdateModule(context);
return false;
```

nekronos:
@sebbert this test is failing: https://github.com/nekronos/fuselibs-public/blob/792c7edeef0e77393ef944164c3aa25a6734aaa4/Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno#L265

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
mortoray: I wrote it. It means the property should probably be called "JavaScript.Model" for consistency, but I wasn't involved in the naming to know why that wasn't chosen (maybe something to do with preview).

The typical use-case would work either way `<App Model="blah()"/>` but without the name change this wouldn't work `<App JavaScript.model="blah()"/>` though `<Panel ux:Class="myPanel" JavaScript.Model="blah()"/>` does work.

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
mortoray: I think the TODO just needs to be removed now.

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
mortoray: I think we should just change it and fix Preview if it doesn't work there.

Source/Fuse.Nodes/AppBase.uno
```diff
@@ -11,6 +11,11 @@ using Fuse.Resources;
 
 namespace Fuse
 {
+	public interface IApp
```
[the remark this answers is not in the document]
sebbert: The `IApp` interface

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+
+        void SetupModel()
+        {
+			if (_model == null)
+			{
+				Code = "";
+				return;
+			}
+
+            //TODO: this should not be necessary. It's done because there's a UX processor error, we get
```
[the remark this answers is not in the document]
mortoray: The problem here goes beyond the TODO. It shouldn't be getting an expression at all, it's rebuilding a scoped name from an expression (a hack). If that were fixed then the TODO wouldn't be an issue -- nothing should be looking at the structure of an IExpression!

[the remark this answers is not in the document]
yupferris: > nothing should be looking at the structure of an IExpression!

Absolutely, we need to be able to change how a given expression is compiled/evaluated, even if when evaluated it will provide the same result (which is of course a bit orthogonal to this use case).

Source/Fuse.Scripting.JavaScript/FuseJS/Tests/tests/diTests.js
```diff
@@ -0,0 +1,36 @@
+
+var GlobalDI = require("../../DI");
+var assert = require("assert");
+
+describe("FuseJS/DI", function() {
+    var DI = null;
+    beforeEach(function() {
```
kristianhasselknippe: indentation seems off

Source/Fuse.Scripting.JavaScript/FuseJS/Timer.uno
```diff
@@ -135,12 +135,11 @@ namespace Fuse.Reactive.FuseJS
 
 			return null;
 		}
-
```
kristianhasselknippe: extra whitespace added

Source/Fuse.Nodes/AppBase.uno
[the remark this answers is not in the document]
nekronos: The naming was unfortunate, written in frustration. Lets rename to `IRootVisualProvider` or something similar

Source/Fuse.Scripting.JavaScript/JavaScript.Model.uno
```diff
@@ -0,0 +1,254 @@
...
+					for (int i=0; i < Args.Count; ++i)
+						str += ", " + Args[i];
+					return str;
+				}
+			}
+			
+			public bool CompatibleTo( ParsedModelExpression o )
+			{
+				//there is no way to migrate these now as they might refer to tree objects, thus reject entirely
+				if (Args.Count != 0 || o.Args.Count != 0 ||
```
kristianhasselknippe: whitespace at the end of this line

Source/Fuse.Scripting.JavaScript/Tests/Model/Model.Test.uno
nekronos: Also, lets not use egyptian braces 🇪🇬 Reminds me of Java 😨

Source/Fuse.Scripting.JavaScript/FuseJS/Timer.uno
sebbert: Added where?

Source/Fuse.Scripting.JavaScript/JavaScript.uno
```diff
@@ -54,6 +59,7 @@ namespace Fuse.Reactive
 
 			if(--_javaScriptCounter <= 0)
 			{
+				//TODO: probably have to block this a _moduleInstance was preserved somewhere...?
```
yupferris: uhh, what? :)

Source/Fuse.Scripting.JavaScript/JavaScript.uno
```diff
@@ -72,7 +78,7 @@ namespace Fuse.Reactive
 
 		object _currentDc;
 		Uno.IDisposable _sub;
-
```
yupferris: seems pointless

Source/Fuse.Scripting.JavaScript/JavaScript.uno
```diff
@@ -81,7 +87,7 @@ namespace Fuse.Reactive
 			_currentDc = newDc;
 
 			var obs = newDc as IObservable;
-			if (obs != null)
```
yupferris: pointless whitespace change

Source/Fuse.Scripting.JavaScript/RootableScriptModule.uno
```diff
@@ -42,7 +45,7 @@ namespace Fuse.Scripting.JavaScript
 		{
 			var argsString = base.GenerateArgs(c, result, args);
 
-			foreach (var dep in Dependencies)
```
yupferris: pointless whitespace change

Source/Fuse.Scripting.JavaScript/JavaScript.uno
mortoray: I think there's a missing "if", "...block this if a _moduleInstance..."

I think it was my question about contexts and initialization. So far I've not seen problems in preview, but if something does some up this might be a place to look. On the short-term though, perhasp the comment can just be removed for now.

Source/Fuse.Scripting.JavaScript/ModuleInstance.uno
```diff
@@ -67,16 +83,6 @@ namespace Fuse.Scripting.JavaScript
...
 
-		object EvaluateExports(Scripting.Context context)
-		{
-			EvaluateModule(context);
-
-			if (_moduleResult != null)
-				return _moduleResult.Object["exports"];
-
-			return null;
-		}
```
kusma: This rewrite here seems rather pointless...

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -23,11 +24,11 @@ namespace Fuse.Scripting.JavaScript
...
 		This prevents leakage of strong references to observables on the JS side.
 	*/
-	class ObservableProperty: IObserver, IPropertyListener
+	class ObservableProperty: Fuse.Reactive.IObserver, IPropertyListener
 	{
 		protected readonly ThreadWorker _worker;
-		Uno.UX.Property _property;
-		Scripting.Object _obj;
+		readonly Uno.UX.Property _property;
+		readonly Scripting.Object _obj;
```
kusma: readonly is a cleanup that could come as a separate PR...

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -23,11 +24,11 @@ namespace Fuse.Scripting.JavaScript
 
 		This prevents leakage of strong references to observables on the JS side.
 	*/
-	class ObservableProperty: IObserver, IPropertyListener
+	class ObservableProperty: Fuse.Reactive.IObserver, IPropertyListener
```
kusma: Why do we need to fully qualify this all of a sudden?

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -40,16 +41,95 @@ namespace Fuse.Scripting.JavaScript
...
+					|| IsDefault<int2>(value, t)
+					|| IsDefault<int3>(value, t)
+					|| IsDefault<int4>(value, t)
+					|| IsDefault<byte2>(value, t)
+					|| IsDefault<byte4>(value, t)
+					|| IsDefault<float2>(value, t)
+					|| IsDefault<float3>(value, t)
+					|| IsDefault<float4>(value, t)
+					|| IsDefault<Size>(value, t)
+					|| IsDefault<Size2>(value, t);
```
kusma: Yikes! This seems like a maintenance nightmare. When do we need to add types here?!

Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -86,7 +86,7 @@ namespace Fuse.Scripting.JavaScript
 
 			try
 			{
-				context.FuseJS.UpdateModules(context);
+				didAnything = context.FuseJS.UpdateModules(context) || didAnything;
```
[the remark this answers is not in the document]
kusma: Also, we should have a test that detects if this breaks in the future.

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
nekronos: It collides with `Uno.IObserver`

kusma:
@sebbert: Seems like we need a rebase on master...

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
kusma: How much worse would it be to drop `using Uno;`? There's a *lot* of qualifying going on here, and it looks to me like this is to be able to write `Uno.Action` and `Uno.Type` four times at the cost of having to add "Fuse.Reactive." 9 times...

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
nekronos: @sebbert

Source/Fuse.Scripting.JavaScript/JavaScript.uno
kusma: Soo, I guess it should be removed, then?

Source/Fuse.Models/FuseJS/Internal/Model.js
```diff
@@ -0,0 +1,489 @@
...
+	var idToMeta = new Map();
+	var evaluatingDerivedProps = 0;
+	var idEnumerator = 0;
+	var store = this;
+
+	instrument(null, this, initialState, stateInitializer)
+
+	function instrument(parentMeta, node, state, stateInitializer)
+	{
+		if(stateInitializer instanceof Function) {
```
kusma: The rest of the code does a space between `if` and `(` ...

kusma: Actually, seems like we're pretty insonsistent... We should pick one and stick with it it (bonus points for adding it to 
[Documentation/CodingStyle.md](https://github.com/fusetools/fuselibs-public/blob/master/Documentation/CodingStyle.md) as well)

Source/Fuse.Models/Fuse.Models.unoproj
```diff
@@ -0,0 +1,33 @@
+{
+  "Copyright": "Copyright (c) Fusetools AS 2017",
+  "Description": "Native Location",
+  "Publisher": "Fusetools AS",
+  "Version": "1.4.0-rc4",
```
kusma: This version number is incorrect. You can use 0.1.0 as a placeholder until we do a real release.

Source/Fuse.Models/Fuse.Models.unoproj
```diff
@@ -0,0 +1,33 @@
+{
+  "Copyright": "Copyright (c) Fusetools AS 2017",
+  "Description": "Native Location",
```
kusma: This description is wrong.

Source/Fuse.Models/Fuse.Models.unoproj
```diff
@@ -0,0 +1,33 @@
...
+  "Publisher": "Fusetools AS",
+  "Version": "1.4.0-rc4",
+  "Package": {
+    "ProjectUrl": "https://fusetools.com"
+  },
+  "Packages": [
+    "Uno.Collections",
+    "Uno.Threading",
+    "Uno.Permissions",
```
kusma: I doubt that models requires Uno.Permissions... Also, are you sure it needs Uno.Threading?

Source/Fuse.Models/Fuse.Models.unoproj
```diff
@@ -0,0 +1,33 @@
...
+  "Projects": [
+    "../Fuse.Common/Fuse.Common.unoproj",
+    "../Fuse.Marshal/Fuse.Marshal.unoproj",
+    "../Fuse.Scripting/Fuse.Scripting.unoproj",
+    "../Fuse.Scripting.JavaScript/Fuse.Scripting.JavaScript.unoproj",
+    "../Fuse.Reactive/Fuse.Reactive.unoproj",
+    "../Fuse.Reactive.Expressions/Fuse.Reactive.Expressions.unoproj",
+    "../Fuse.Platform/Fuse.Platform.unoproj",
+    "../Fuse.Common/Fuse.Common.unoproj",
+    "../Fuse.Nodes/Fuse.Nodes.unoproj",
```
kusma: I doubt that models need Fuse.Platforms, Fuse.Nodes... And it definitely don't need Fuse.Common twice.

Source/Fuse.Models/FuseJS/Internal/Model.js
```diff
@@ -0,0 +1,489 @@
+var TreeObservable = require("FuseJS/TreeObservable")
+
+require("Polyfills/Window");
+require("./zone.min")
+require("./ZonePatches");
+
+var rootZone = Zone.current;
+
+function shouldEmitProperty(key) {
+	return !key.startsWith("__fuse_");
```
kusma: Perhaps it would be a good idea to say that any property starting with double underscores are omitted? That way, the user can intentionally hide implementation details as well, which seems to me like a good idea.

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -40,16 +41,95 @@ namespace Fuse.Scripting.JavaScript
...
+				_property = op._property;
+				_subscription = op._subscription;
+				_pushValueOnJSThread = op.PushValueOnJSThread;
+				_value = value;
+			}
+
+			// UI thread
+			public void Perform()
+			{
+				// HACK: https://github.com/fusetools/fuselibs-public/issues/541#issuecomment-335101235
```
kusma: This seems like something that should be fixed before we merge, no?

nekronos: This is the stuff that we want to delay fixing after release as discussed in the daily last friday

kusma: Yeah, I figured after seeing this. So I guess this part is OK for now.

Source/Fuse.Models/JavaScript.Model.uno
```diff
@@ -0,0 +1,267 @@
...
+		{
+			if (_model == null)
+			{
+				Code = string.Empty;
+				return;
+			}
+
+			//TODO: this should not be necessary. It's done because there's a UX processor error, we get
+			//the IExpression prior to it being complete
+			var module = ParseModelExpression( _model.Source, _nameTable );
```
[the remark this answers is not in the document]
nekronos: @sebbert ?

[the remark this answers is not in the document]
sebbert: The comment isn't incomplete, just oddly worded. The issue here is that, for properties that accept an `IExpression`, the UX compiler calls the property setter before the `IExpression` has been fully constructed. @mortoray probably knows more about this.

Source/Fuse.Models/JavaScript.Model.uno
```diff
@@ -0,0 +1,267 @@
...
+			//var module = _model;
```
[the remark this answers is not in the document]
sebbert: It's probably been overlooked. Will fix.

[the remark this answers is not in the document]
sebbert: Fixed.

kusma:
> Source\Fuse.Scripting.JavaScript\ObservableProperty.uno(115.52): E3114: There is no identifier named 'Type' accessible in this scope. Did you mean 'Type' (as in Fuse.Scripting.ScriptClass.Type), 'Type' (as in Fuse.Animations.Attractor`1.Type) or 'Type' (as in Fuse.Animations.Attractor<T>.Type), or one of 19 other similar candidates?

Source/Fuse.Models/JavaScript.Model.uno
```diff
@@ -0,0 +1,266 @@
...
+using Uno.Compiler;
+using Uno.IO;
+using Uno.Text;
+using Fuse;
+using Fuse.Reactive;
+using Fuse.Scripting;
+
+namespace Fuse.Models
+{
+	public class ModelJavaScript : JavaScript, IPreviewStateSaver
```
kusma: Please name the file here ModelJavaScript.uno rather than JavaScript.Model.uno. This is no longer a partial class.

Source/Fuse.Models/JavaScript.Model.uno
```diff
@@ -0,0 +1,266 @@
...
+			var divide = exp as Divide;
+			if (divide != null)
+			{
+				var left = ParseModelExpression(divide.Left, nt);
+				var right = ParseModelExpression(divide.Right, nt);
+				
+				if (left.Args.Count > 0 || left.Dependencies.Count > 0)
+					throw new Exception( "Invalid Model path expression: " + exp);
+				
+				right.ModuleName = left.ModuleName + "/" + right.ModuleName;
```
kusma: I've [said this before](https://github.com/fusetools/fuselibs-public/pull/601#discussion_r151416319), but this here looks like a pretty nasty abuse of the expression-system to me... I'm rather worried that needs of the expression system and model-paths might diverge in the future, rendering a solution like this unfeasible. Don't we have some other options here, like `[UXVerbatim]`?

Source/Fuse.Navigation/RouterPage.uno
```diff
@@ -167,58 +166,58 @@ namespace Fuse.Navigation
...
 		}
-		
+
 		public RouterPageRoute Append( RouterPageRoute subRoute )
 		{
 			var sub = SubRoute == null ? subRoute : SubRoute.Append(subRoute);
 			return new RouterPageRoute(RouterPage, sub);
 		}
-		
+
```
kusma: Looks like the changes in this file is just white-space cruft now, can we please drop it?

Source/Fuse.Models/JavaScript.Model.uno
```diff
@@ -0,0 +1,266 @@
...
+		{
+			_previewStateModelId = previewStateId;
+			_model = model;
+			FileName = "(model-script)";
+		}
+		
+		protected override void OnRooted()
+		{
+			//TODO: prior to base.OnRooted is questionable... The TODO: in SetupModel though is part of the
+			//reason we can't easily fix this now. Ideally the constructor would just set the needed values
```
kusma: Should this be investigated? If not, please add a ticket to do so, and link it here...

Source/Fuse.Scripting.JavaScript/JavaScript.Dependencies.uno
```diff
@@ -133,8 +133,12 @@ namespace Fuse.Reactive
 			_moduleInstance = new Fuse.Scripting.JavaScript.ModuleInstance(Worker, this);
 		}
 
-		void DisposeModuleInstance()
+		protected bool _preserveModuleInstance = false;
+		protected void DisposeModuleInstance()
 		{
+			if (_preserveModuleInstance)
+				return;
+
```
kusma: This here isn't exactly a beautiful work of art. Calling DisposeModuleInstance() when we really don't want to do it seems counter-intuitive.

I have a feeling that the right thing would have been to avoid calling DisposeModuleInstance() in this case 
in `JavaScript.OnUnrooted()`. I doubt `JavaScript.DispatchEvaluate()` wants this to be preserved anyway...

Source/Fuse.Models/JavaScript.Model.uno
kusma: This still applies

Source/Fuse.Scripting.JavaScript/JavaScript.Dependencies.uno
kusma: By the way, I don't think the above is a deal-breaker, just a notice. It'd be nice to clean up things like this.

Source/Fuse.Models/Tests/Model.Test.uno
```diff
@@ -0,0 +1,585 @@
...
+			if(!(obj is T))
+			{
+				var expected = "object of type '" + typeof(T).FullName + "'";
+				var actual = (obj == null) ? "null" : "object of type '" + obj.GetType().FullName + "'";
+
+				// Temporary: `Assert.ReportFailure` is private
+				var message = "Expected " + expected + ", got " + actual;
+				Assert.Fail(message, filePath, lineNumber, memberName);
+			}
+		}
```
kusma: cleanup-notice: I *think* this one has made it to uno and up by now, so it can probably be removed.

Source/Fuse.Scripting.JavaScript/FuseJS/Builtins.uno
```diff
@@ -75,10 +75,10 @@ namespace Fuse.Reactive.FuseJS
 
 		internal bool UpdateModules(Fuse.Scripting.Context context)
 		{
-			var activity = false;
 			if(_timer != null)
-				activity = _timer.UpdateModule(context) || activity;
-			return activity;
+				return _timer.UpdateModule(context);
+
+			return false;
```
kusma: This cleanup could have been done separately, as it doesn't change any behavior.

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -40,16 +40,104 @@ namespace Fuse.Scripting.JavaScript
...
+				_property = op._property;
+				_subscription = op._subscription;
+				_pushValueOnJSThread = op.PushValueOnJSThread;
+				_value = value;
+			}
+
+			// UI thread
+			public void Perform()
+			{
+				if(IsDefaultValueForType(_property.GetAsObject(), _property.PropertyType))
```
kusma: space between "if" and '('

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
```diff
@@ -144,6 +236,8 @@ namespace Fuse.Scripting.JavaScript
 
 		void PushValue(Scripting.Context context, object val)
 		{
+			if (_subscription == null) return;
+
```
kusma: This looks like a general bug-fix, and could have been done separately on master.

Source/Fuse.Scripting.JavaScript/TreeObservable/TreeArray.uno
```diff
@@ -81,7 +81,7 @@ namespace Fuse.Scripting.JavaScript
 
 					var replaceAllFn = (Scripting.Function) context.Evaluate("replaceAll",
 						"(function(array, values) {" +
-							"if ('__fuse_replaceAll' in array) array.__fuse_replaceAll(values);" +
+							"if('__fuse_replaceAll' in array) array.__fuse_replaceAll(values);" +
```
kusma: This change seems pointless...

Source/Fuse.Scripting.JavaScript/FuseJS/Builtins.uno
kusma: I've separate this into #811

Source/Fuse.Scripting.JavaScript/ObservableProperty.uno
kusma: I've separate this into #811

Source/Fuse.Scripting.JavaScript/JavaScript.Dependencies.uno
mortoray: Hmm, I think I agree. Only the one call-site should preserve it, since the other one happily overwrites it anyway (but I don't think that's called if the is a preserved instance, thus this broken code doesn't fail).

Source/Fuse.Models/Tests/Model.Test.uno
kusma: No, that doesn't seem to be true.

Source/Fuse.Models/ModelJavaScript.uno
```diff
@@ -0,0 +1,165 @@
...
+			else
+			{
+				md.NameTable = nt;
+			}
+			
+			Complete( md, v );
+		}
+
+		//TODO: This should probably be JavaScript.Model, Preview would need to be adjusted as well
+		[UXAttachedPropertySetter("Model")]
```
kusma: Ticket and link, please?

Source/Fuse.Scripting.JavaScript/ModuleInstance.uno
```diff
@@ -103,8 +109,6 @@ namespace Fuse.Scripting.JavaScript
 					// Don't report chain-errors of already reported errors
 					if (!se.Message.Contains(ScriptModule.ModuleContainsAnErrorMessage))
 						_diagnostic.SetDiagnostic(se);
-
-					newModuleResult.Dispose();
```
kusma: Please don't! This was a recent resource-leak fix, and just removing it seems wrong.

kusma:
I've pointed this out before, but 94ab2e7ce50b6151a7d6507662e664d73ee524cc doesn't build. This needs to be fixed, to keep the ability to bisect.

Source/Fuse.Scripting.JavaScript/ModuleInstance.uno
```diff
@@ -67,16 +83,6 @@ namespace Fuse.Scripting.JavaScript
...
-		object EvaluateExports(Scripting.Context context)
-		{
-			EvaluateModule(context);
-
-			if (_moduleResult != null)
-				return _moduleResult.GetExports(context);
-
-			return null;
-		}
-
```
kusma: It's not clear to me what all the refactoring in here is good for. I can't see any functional change, just moved code. Care to explain?

Source/Fuse.Scripting.JavaScript/RootableScriptModule.uno
```diff
@@ -27,8 +27,11 @@ namespace Fuse.Scripting.JavaScript
 
 		void EnsureClassInstanceRooted(Scripting.Context c)
 		{
-			if (_classInstance == null) _classInstance = ((JSContext)c).GetClassInstance(_names);
-			_classInstance.EnsureRooted(c);
+			if (_names != null)
+			{
+				if (_classInstance == null) _classInstance = ((JSContext)c).GetClassInstance(_names);
+				_classInstance.EnsureRooted(c);
+			}
```
kusma: Are we sure _names == null is OK here? Why would we get a null NameTable passed to the constructor? It seems nonsensical to me...

Source/Fuse.Models/Tests/Model.Test.uno
sebbert: Sounds like you're talking about different methods (OfType vs. ReportFailure). OfType has now landed on uno master, will switch to that

Source/Fuse.Scripting.JavaScript/RootableScriptModule.uno
```diff
@@ -53,7 +56,10 @@ namespace Fuse.Scripting.JavaScript
 
 		protected override void CallModuleFunc(Context context, Function moduleFunc, object[] args)
 		{
-			_classInstance.CallMethod(context, moduleFunc, args);
+			if (_classInstance != null)
+				_classInstance.CallMethod(context, moduleFunc, args);
+			else
+				moduleFunc.Call(context, args);
```
kusma: This also seems bogus to me. This would only trigger if `_names` is null or `CallModuleFunc` was called before `EnsureClassInstanceRooted`, neither seems very reasonable to me.

Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
```diff
@@ -34,8 +34,7 @@ namespace Fuse.Controls
 			
 			if (_pageHistory == null)
 				return;
-			
-			var obs = _pageHistory as IObservable;
+			var obs = _pageHistory as IObservableArray;
```
kusma: It's not quite clear to me why change is needed. it is done in 01fd5efe68d018b3c2a95771b0cf95a8dd5e3531 while adding a test, which seems a bit odd... Is this a bug-fix? If so, I think we need a better explanation in the commit message...

sebbert: That seems to be a squash of these two commits here: https://github.com/sebbert/fuselibs-public/compare/36742cb3c2070b63dbcf8f62975ba71de5960f75~2...36742cb3c2070b63dbcf8f62975ba71de5960f75

Source/Fuse.Models/FuseJS/Internal/Model.js
mortoray: There's supposed to be an "underscore" convention to omit variables already (single underscore).  If that's not a feature then we need to add (can be done later). Otherwise our state will include way too much stuff.

Source/Fuse.Scripting.JavaScript/ModuleInstance.uno
mortoray: The changes account for a simplification/split in the JavaScript tag into the implied one for model and the one the user explicitly does. In particular, the implied one can be persisted (moduleInstance) in Preview. These are all minor changes to accomodate that change.

Source/Fuse.Scripting.JavaScript/RootableScriptModule.uno
mortoray: I don't think _names is allowed to be null anymore, but I think it was in the past. I suspect it'd be a good improvement to check it in the constructor(s).

Source/Fuse.Controls.Navigation/NavigationControl.Pages.uno
mortoray: Many places that used `IObservable` before only actually need an `IObservableArray` now (which is a base class to `IObservable`. This is part of the type cleanup in the bindings.

kusma: Then this seems to be squashed into the wrong commit.

Source/Fuse.Scripting.JavaScript/RootableScriptModule.uno
kusma: Yeah, that's indeed what we should do.

Source/Fuse.Scripting.JavaScript/FuseJS/Builtins.uno
kusma: We should rebase on master, to get rid of this.

Source/Fuse.Scripting.JavaScript/ModuleInstance.uno
kusma: Ah, so `ReflectExports()` gets called by `ModelJavaScript`? That makes a lot of sense, thanks for the clarification :)

Source/Fuse.Models/ModelJavaScript.uno
sebbert: @mortoray I've tried changing this to `ModelJavaScript.Model`, but the UX compiler complains that the `Model` attribute could not be found. I suppose it's illegal to have multiple attached properties with the same qualified name, even if they can be discriminated by the type of their first argument? In any case, I assume the comment about preview is outdated (went away with the introduction of IRootVisualProvider)

mortoray: I thought that Preview needed to know about `Model`. But if not then we don't have to worry about that part.

I'd still call it `JavaScript.Model` though, since `ModelJavaScript` is an internal detail.  I don't know why the UX compiler would give a conflict though.

sebbert: @mortoray Unless you know of any more places, [this is the only place I know of where preview used to have knowledge of the model feature](https://github.com/fusetools/Fuse/commit/03067734d22ecf54330297a0e9988b3e142581d5#diff-7eae56ceee21977d27ede24e1ae9569dL61). The rest has been abstracted away via `IRootVisualProvider` and your preview state system IIRC.

mortoray: Nope. That was it, thus it should be safe to change.

Source/Fuse.Scripting.JavaScript/RootableScriptModule.uno
sebbert: @mortoray In `ModelJavaScript`, we're passing in `null` as the NameTable for the JavaScript instance (which then gets passed to RootableScriptModule):

```
//app-level model does not have a nametable otherwise migration would not be possible
var js = new ModelJavaScript(null, modulePath, previewStateId);
```

Any idea what to do there instead then?

mortoray: It'd probably be cleaner to pass in an empty NameTable. Then the code wouldn't need to deal with the null. It's at the root as well, thus it won't cost any unintended overhead. I don't thinks it's an essential change prior to merge though.

kusma:
39c2357f536b12d495b470dbeaac05ecf10b3ebe still fails to compile:
```
Source\Fuse.Scripting.JavaScript\ModuleInstance.uno(44.4): E3128: Call to 'Fuse.Scripting.JavaScript.ModuleInstance.EvaluateModule(Fuse.Scripting.Context)' has some invalid arguments ()
Source\Fuse.Scripting.JavaScript\ModuleInstance.uno(45.4): E3128: Call to 'Fuse.Scripting.JavaScript.ModuleInstance.ReflectExportsJS(Fuse.Scripting.Context)' has some invalid arguments ()
Source\Fuse.Scripting.JavaScript\JavaScript.uno(37.33): E3107: Fuse.Scripting.ScriptModule does not contain a member called 'AddMagicPath'. Could you be missing a package reference?
```

This needs to be fixed before we merge, so we don't break bisectability...

nekronos:
Fixed `15c9dbf4192341fe61314e5a3a04786a2737a9db`

Duckers:
🎉  great work!!

Issue #594, opened 2017-10-17
Clean up example template


knatten:
We got an issue from manual testing in https://github.com/fusetools/Fuse/issues/4325, and after some discussion figured out the issue was with the [example template](https://github.com/fusetools/fuselibs-public/tree/master/ProjectTemplates/Projects/Example) itself.

See this comment by @mortoray  in particular: https://github.com/fusetools/Fuse/issues/4325#issuecomment-337167447
> File an issue on whever "examples" issues go. The example should be cleaned up, no border on the image, better layout on that page.

kristianhasselknippe:
I've attempted a fix here https://github.com/fusetools/fuselibs-public/pull/606 , but it seems to me like the bug we experience in the video is less due to the borders around the image and more because its using the default stretch mode: since the height is set to 500x500, which is much taller than the image is.

PR #606, opened 2017-10-18
Using Path instead of Image to display fuse logo in the example template

kristianhasselknippe:
An attempt at fixing https://github.com/fusetools/fuselibs-public/issues/594

This PR contains:
- ~~[ ] Changelog~~
- ~~[ ] Documentation~~
- ~~[ ] Tests~~


kristianhasselknippe:
So, there was a small bug with the path rendering on dotnet, which i put in a small hack to circumvent. Explained here https://github.com/fusetools/fuselibs-public/pull/606/commits/f0e6f91288fb079d495faa21901c26443fd1b868

kusma:
@kristianhasselknippe: Nice. But would you mind also filing a ticket for the inaccurate clipping, so we have a chance of getting it fixed eventually? :)

Issue #593, opened 2017-10-17
A lerp expression with binding that silently does nothing


mortoray:
This code doesn't emit any diagnostics. It should since "ch" is not a valid color.

```
<App>
	<JavaScript>
		exports.color = "ch"
	</JavaScript>
	<StackPanel>
		<Panel Height="20" Color="lerp({color},#FFF,0.5)"/>
		<Panel Height="20" Color="#F80"/>
	</StackPanel>
</App>
```

PR #607, opened 2017-10-18
Emit user warning when lerp gets passed invalid arguments

kristianhasselknippe:
An attempt at fixing https://github.com/fusetools/fuselibs-public/issues/593

I'm not entirely sure whether it is correct sending `this` to the UserWarning function as it ends up pointing to the lerp uno implementation when displaying to the user (at least when i ran it locally).

Thoughts @mortoray ?

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Reactive.Expressions/MathFunctions.uno
+++ b/Source/Fuse.Reactive.Expressions/MathFunctions.uno
@@ -425,10 +425,22 @@ namespace Fuse.Reactive
 			float4 av = float4(0), bv = float4(0);
 			int asize = 0, bsize = 0;
 			float tv = 0;
-			if (!Marshal.TryToZeroFloat4(a, out av, out asize) ||	
-				!Marshal.TryToZeroFloat4(b, out bv, out bsize) ||
-				!Marshal.TryToType<float>(t, out tv))
+
+			var aValid = Marshal.TryToZeroFloat4(a, out av, out asize);
+			var bValid = Marshal.TryToZeroFloat4(b, out bv, out bsize);
+			var tValid = Marshal.TryToType<float>(t, out tv);
+			if (!aValid)
+				Fuse.Diagnostics.UserWarning("The first argument of lerp is not supported.", this);
+			if (!bValid)
+				Fuse.Diagnostics.UserWarning("The second argument of lerp is not supported.", this);
+			if (!tValid)
+				Fuse.Diagnostics.UserWarning("The third argument of lerp is not supported.", this);
+
+			if (!aValid ||
+				!bValid	||
+				!tValid)
 				return null;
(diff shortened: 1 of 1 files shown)


kusma:
Actually, seems like this triggers some test-failures... Because of unexpected diagnostics. You should update the tests to record the diagnostics and assert their count and message.

kristianhasselknippe:
@kusma : not sure how to do that.. could you point me in the right direction?

kusma:
@kristianhasselknippe: Check out uses of `RecordDiagnosticGuard`, it's what you want to do...

Source/Fuse.Reactive.Expressions/MathFunctions.uno
```diff
@@ -425,10 +425,22 @@ namespace Fuse.Reactive
...
+
+			var aValid = Marshal.TryToZeroFloat4(a, out av, out asize);
+			var bValid = Marshal.TryToZeroFloat4(b, out bv, out bsize);
+			var tValid = Marshal.TryToType<float>(t, out tv);
+			if (!aValid)
+				Fuse.Diagnostics.UserWarning("The first argument of lerp is not supported.", this);
+			if (!bValid)
+				Fuse.Diagnostics.UserWarning("The second argument of lerp is not supported.", this);
+			if (!tValid)
+				Fuse.Diagnostics.UserWarning("The third argument of lerp is not supported.", this);
```
kusma: Actually, I suspect that this should use `ReportTemporalUserWarning` instead, and I think those are temporarily ignored...

mortoray: I think `UserWarning` is fine, since this function won't be called until all values are available. It's not a "temporal" warning like the "doesn't exist" warnings are. Though I would argue it's a `UserError` here, not a warning. They're provdiing invalid values to lerp.

mortoray:
You might need to not emit if you get a `null`. There's a lack of clarity still as to whether null is a non-value propagation or an actual propagation.

Source/Fuse.Reactive.Expressions/MathFunctions.uno
kusma: OK, if that's the case, I agree.

mortoray:
@kristianhasselknippe This need not be blocked anymore. I finished my work where a value lost signal is sent when values are lost. You can now assume that any value which doesn't convert is an error, even null.

mortoray:
After my changes to the operators the fix for this will look different. I'll make the change and submit a new PR (closing this one when I'm done).

mortoray:
Generic error handling replaces this.

PR #619, opened 2017-10-23
Moved ByteBufferInputStream to Fuse.Common

kristianhasselknippe:
Otherwise, Exif would not compile for android when running the
Fuse.Scripting.JavaScript tests. Exif is compiled at all in this case
because it doesn't get properly stripped, most likely due to uno/issues/1368.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
@kristianhasselknippe: This is needed against release-1.4, not master.

kusma:
Confirmed to fix the issue for me!

PR #625, opened 2017-10-24
Mobile test fixes

kusma:
Here's the first batch of test-fixes for mobile. This makes a bunch more tests pass on Android and iOS.

The first patch is about making the tests actually compile on iOS, due to a missing require.

### DIFF
--- a/Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
+++ b/Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
@@ -51,8 +51,10 @@ namespace Fuse.Controls
 		{
 			if defined(iOS || Android)
 			{
-				var x = SystemUI.TopFrame.Size / AppBase.Current.PixelsPerPoint;
-				return x;
+				var pixelsPerPoint = 1.0f;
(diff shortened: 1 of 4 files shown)


Source/Fuse.Controls.Panels/Backgrounds/TopFrameBackground.uno
```diff
@@ -51,8 +51,10 @@ namespace Fuse.Controls
 		{
 			if defined(iOS || Android)
 			{
-				var x = SystemUI.TopFrame.Size / AppBase.Current.PixelsPerPoint;
-				return x;
+				var pixelsPerPoint = 1.0f;
```
mortoray: Hmm, this is a `visual` so it should actually be using the `Viewport` to get access to this information (and it should actually be using `PixelsPerOSPoint`).  But I guess that's not the purpose of this fix.

kusma: Yeah, that feels out-of-scope for this fix. I would prefer to *just* fix the tests for now. We need a deep-dive into the TopFrameBackground-stuff soon anyway to deal with Android 8 issues.

PR #629, opened 2017-10-25
Fuse.Controls.Panels: respect Panel.Opacity when rendering

kusma:
This was discovered during code-inspection.

This PR contains:
- [x] Changelog
- [ ] ~Documentation~ No new feature
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.Panels/Panel.Freeze.uno
+++ b/Source/Fuse.Controls.Panels/Panel.Freeze.uno
@@ -198,7 +198,7 @@ namespace Fuse.Controls
 				return;
 			}
 
-			FreezeDrawable.Singleton.Draw(dc, this, 1, Scale, _frozenRenderBounds, _frozenBuffer);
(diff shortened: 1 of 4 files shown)


Source/Fuse.Controls.Panels/Panel.Freeze.uno
```diff
@@ -226,7 +226,7 @@ namespace Fuse.Controls
 				Texture: buffer.ColorBuffer;
 				Invert: true;
 				
-				PixelColor: float4( prev.XYZ, prev.W * Opacity );
+				PixelColor: prev * Opacity;
```
mortoray: This converts to premultiplied output, doesn't it? Was that intended?

kusma: It doesn't convert, it multiplies with the alpha according to pre-multiplied rules. The image is rendered on the GPU, and the draw-block uses `Fuse.Drawing.PreMultipliedAlphaCompositing`, we must keep the alpha fully premultiplied.

Issue #615, opened 2017-10-23
Block interaction during navigation


mortoray:
With default transitions  for `Navigator` and `Pagecontrol` the interaction with the page should be disabled until navigation completes. This blocking must occur from the moment of navigations start -- at the prepare time, to avoid allowing multiple clicks that invoked the navigation.

`HitTestMode="None"` is probably what should block interaction (`IsEnabled="false"` would change the appearance).

PR #644, opened 2017-10-26
Block interaction while navigating

mortoray:
Fixes https://github.com/fusetools/fuselibs-public/issues/615

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.Navigation/NavigationControl.BlockInput.uno
+++ b/Source/Fuse.Controls.Navigation/NavigationControl.BlockInput.uno
@@ -1,0 +1,64 @@
+using Fuse.Triggers;
+
+namespace Fuse.Controls
+{
+	public enum NavigationControlBlockInput
+	{
+		/** Input is not blocked */
+		Never,
+		/** Input is blocked during navigation -- when `WhileNavigating` would be active */
+		WhileNavigating,
+	}
+	
+	public abstract partial class NavigationControl
+	{
+		NavigationControlBlockInput _blockInput = NavigationControlBlockInput.WhileNavigating;
+		/**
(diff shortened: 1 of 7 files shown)


Source/Fuse.Controls.Navigation/NavigationControl.BlockInput.uno
```diff
@@ -0,0 +1,64 @@
...
+		
+		
+		void BlockInputUnrooted()
+		{
+			UpdateBlockInput(true);
+		}
+		
+		void UpdateBlockInput(bool forceOff = false)
+		{
+			if (forceOff || _blockInput == NavigationControlBlockInput.Never)
```
kusma: I think I would separate this a bit, like so:

```Uno
void UpdateBlockInput()
{
	if (_blockInput == NavigationControlBlockInput.Never)
		DisableBlockInput();
	else
		EnableBlockInput();
}

void EnableBlockInput()
{
	if (_blockInputTrigger == null)
	{
		_blockInputTrigger = new NavigationInternal.BlockInputWhileNavigating(this);
		Children.Add(_blockInputTrigger);
	}
}

void DisableBlockInput()
{
	if (_blockInputTrigger != null)
	{
		Children.Remove(_blockInputTrigger);
		_blockInputTrigger = null;
	}
}
```

This gets us rid of the `forceOff`-argument, and we can just call `DisableBlockInput()` from `BlockInputUnrooted()` instead.

mortoray: Yeah, I'm really not sure why I used the flag.

Issue #614, opened 2017-10-20
`PageControl.ActiveIndex` not updating after a `seekToPath`


mortoray:
https://www.fusetools.com/community/forums/bug_reports/pagecontrol_seektopath_issues_after_fuse_130

fusebuild:
The forum thread [pageControl seekToPath issues after fuse 1.3.0](https://www.fusetools.com/community/forums/bug_reports/pagecontrol_seektopath_issues_after_fuse_130) posted by mircea@code11.com was linked to this issue.

PR #645, opened 2017-10-26
fixes PageControl.ActiveIndex

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/614

@kusma This fixes a regression in 1.3, so at the very least it should be in 1.4, or if there is planned 1.3 release then there.

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 # 1.3
 
+## 1.3.1
+
+### Navigation
+- Fixed an issue where `PageControl.ActiveIndex` would not update if navigation done with JavaScript `seekToPath` or `Router` interfaces.
+
 ## 1.3.0
 
 ### Native UI:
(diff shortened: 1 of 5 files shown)


kusma:
@mortoray: OK, thanks for the heads up. This means this should at least be pulled against 1.4, and we should also consider doing a 1.3.1 IMO. But the latter is of course a bigger question that more people needs to be involved in.

mortoray:
I don't think this alone would justify a 1.3.1. The defect appears as part of a feature (Pagecontrol with dynamic pages) that was previously littered with defects and problems -- limiting it's use, thus limiting the impact of the defect. The rewrite of this code in 1.3 fixed a lot of those issues (but obviously introduced this one).

kusma:
Well, the common wisdom is that for every user who reports a bug, there's a bunch who doesn't. I think if we know that we break something in an actively maintained release, we should fix it. It's not a lot of work.

bentstamnes:
> Well, the common wisdom is that for every user who reports a bug, there's a bunch who doesn't.

I believe the ratio is 1:26 (for every single report, there are 26 that that experience the same issue but don't know how or don't bother to report it)

mortoray:
Yes, I know that's the normal case, I just mean because the feature-set was flaky before, the rate of use was likely more limited than a functioning feature.

kusma:
@mortoray: Yikes: https://ci.appveyor.com/project/fusetools/fuselibs-public/build/1689#L438

PR #646, opened 2017-10-27
Optimize image transform

kusma:
Using trigenometry and matrix-multiplictaions is a pretty slow way
of calculating these matrices. Let's instead hand-roll a version of
this that manually expands the known-zero parts of the matrices for
each variation.

As a nice benefit, we get a more accurate result :)

This PR contains:
- [x] Tests


kristianhasselknippe:
LGTM :)

PR #656, opened 2017-10-29
transition to new Native build target (fusetools/uno#1382)

mortend:
* We now use `<Require SharedLibrary>` to get a DLL for the right architecture copied to the right place.
* Remaining `TARGET_MSVC`-things are no longer used by the new target and can soon be removed.
* Includes a compile fix after the `char16_t`-refactoring.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Text/Implementation/ICU.cpp.uxl
+++ b/Source/Fuse.Text/Implementation/ICU.cpp.uxl
@@ -1,31 +1,60 @@
 <Extensions Backend="CPlusPlus">
 	<Define USE_ICU="!Android" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icudt55.dll" TargetName="bin/Win32/Release/icudt55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icuin55.dll" TargetName="bin/Win32/Release/icuin55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icuio55.dll" TargetName="bin/Win32/Release/icuio55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icule55.dll" TargetName="bin/Win32/Release/icule55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/iculx55.dll" TargetName="bin/Win32/Release/ixulx55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icutu55.dll" TargetName="bin/Win32/Release/icutu55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icuuc55.dll" TargetName="bin/Win32/Release/icuuc55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icudt55.dll" TargetName="bin/Win32/Debug/icudt55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icuin55.dll" TargetName="bin/Win32/Debug/icuin55.dll" />
-	<CopyFile Condition="MSVC" Name="../icu/x86/icu/bin/icuio55.dll" TargetName="bin/Win32/Debug/icuio55.dll" />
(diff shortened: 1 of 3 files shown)


Source/Fuse.Text/Implementation/ICU.cpp.uxl
```diff
@@ -1,31 +1,60 @@
...
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icuuc55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icudt55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icuin55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icuio55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icule55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/iculx55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icutu55.dll':Path)" />
+	<Require Condition="WIN32 && X64" SharedLibrary="@('../icu/x64/icu/bin64/icuuc55.dll':Path)" />
+	<!-- Legacy -->
+	<CopyFile Condition="TARGET_MSVC" Name="../icu/x86/icu/bin/icudt55.dll" TargetName="bin/Win32/Release/icudt55.dll" />
```
kusma: What's the point of keeping these?

mortend: We still need these because the MSVC target isn't actually removed yet (https://github.com/fusetools/uno/pull/1382). We can remove these later when the new stuff has landed.

kusma: Why is this different from V8Simple, then?

mortend: It's not different - V8Simple already had the CopyFile TARGET_MSVC elements, so wasn't much to do there...

mortend: To clarify:
- `MSVC` is still defined by the new native target.
- `TARGET_MSVC` is only defined by the legacy MSVC target.
- We don't want to copy these files when using the new target, so we now use `TARGET_MSVC` to avoid breaking the old target until it's been completely removed.

kusma: I still don't understand why V8Simple doesn't need a lecacy CopyFile-line. You deleted that in this commit, and added a `<Require ... SharedLibrary="..." />`-line. For ICU, you duplicated this stuff instead.

mortend: Here are V8Simples CopyFile lines
```
	<CopyFile Condition="TARGET_MSVC" Name="lib/Windows/x86/V8Simple.dll" TargetName="bin/Win32/Debug/V8Simple.dll" />
	<CopyFile Condition="TARGET_MSVC" Name="lib/Windows/x86/V8Simple.dll" TargetName="bin/Win32/Release/V8Simple.dll" />
	<CopyFile Condition="TARGET_MSVC" Name="lib/Windows/x64/V8Simple.dll" TargetName="bin/x64/Debug/V8Simple.dll" />
	<CopyFile Condition="TARGET_MSVC" Name="lib/Windows/x64/V8Simple.dll" TargetName="bin/x64/Release/V8Simple.dll" />
```
https://github.com/mortend/fuselibs-public/blob/dc23e81de07bdfd22f3efe40ebd123bf4abaabc8/Source/Fuse.Scripting.JavaScript/V8/V8Simple.cpp.uxl

They were already there so I didn't have to change anything. For ICU I had to change `MSVC` -> `TARGET_MSVC`.

kusma: Oh, right. I get it now. Sorry for the confusion :P

PR #657, opened 2017-10-30
Fix ManualTestApp/issues/375

kristianhasselknippe:
Fixes fusetools/ManualTestApp#375

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Tests/ManualTests/ManualTestingApp/Singles/VideoPage.ux
+++ b/Tests/ManualTests/ManualTestingApp/Singles/VideoPage.ux
@@ -3,50 +3,6 @@
 		<p>This page contains a video-player, that should show the video "Big Buck Bunny".</p>
 	</InfoStack>
 	<DockPanel>
-		<Panel>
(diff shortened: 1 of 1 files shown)


Tests/ManualTests/ManualTestingApp/Singles/VideoPage.ux
```diff
@@ -3,47 +3,49 @@
 		<p>This page contains a video-player, that should show the video "Big Buck Bunny".</p>
 	</InfoStack>
 	<DockPanel>
-		<Panel>
-			<Panel ux:Name="_loadingMessage" Background="#000" Opacity="0.5" Visibility="Hidden">
-				<Text TextColor="#fff" Alignment="Center">Loading...</Text>
-			</Panel>
+		<DockPanel>
```
kusma: Do we really need two nested dockpanels? Can't we just dock to top and bottom of the same one instead?

kristianhasselknippe: ok, got rid of the extra dockpanel now :)

PR #662, opened 2017-10-31
Fuse.Scripting.JavaScriptCore: do not throw exceptions through scriping-engine

kusma:
JavaScriptCore is written in C, and cannot handle C++ exception-unwinding. So let's
avoid throwing exceptions through the scripting-engine, and do like we do on V8;
rethrow the exception when we've exited the VM.

This fixes ScriptingTest.CatchingUnoExceptions on iOS.

This PR contains:
- [x] Changelog
- [ ] ~Tests~ This already has a test that failed on iOS

### DIFF
--- a/Source/Fuse.Scripting.JavaScript/JavaScriptCore/Array.uno
+++ b/Source/Fuse.Scripting.JavaScript/JavaScriptCore/Array.uno
@@ -26,17 +26,23 @@ namespace Fuse.Scripting.JavaScriptCore
 		{
 			get
 			{
-				return _context.Wrap(_value.GetPropertyAtIndex(
-					_context._context,
-					index,
-					_context._onError));
+				object result = null;
+				using (var vm = new Context.EnterVM(_context))
+					result = _context.Wrap(_value.GetPropertyAtIndex(
+						_context._context,
+						index,
+						_context._onError));
+				_context.ThrowPendingException();
+				return result;
 			}
 			set
 			{
-				_value.SetPropertyAtIndex(_context._context,
-					index,
-					_context.Unwrap(value),
-					_context._onError);
+				using (var vm = new Context.EnterVM(_context))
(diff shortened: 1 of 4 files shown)


kusma:
It's also worth noting that this was recently [experienced in the wild](https://www.fusetools.com/community/forums/general/terminating_with_uncaught_exception_of_type_uthrow), but [this uno-fix](fusetools/uno#1408) makes the problem go away. This really only triggers in cases where things already have gone south.

ollef:
Looks good!

Some questions, as I've been out of the loop for a while:

* Is the test that failed on iOS something added recently, or did that start failing due to some other change? As far as I know all tests were passing on iOS last I ran them.
* What cleanups to V8 are you referring to (perhaps something removed from the PR)?

kusma:
@ollef: I don't know, but it's hard for me to imagine that it ever worked given the implementation. But perhaps some iOS version have compiled their JSC-build with `-fexceptions` or something?

Ah, yes. I backed out of the V8-cleanups, but forgot to update the cover-letter! Will fix!

ollef:
Cool, thanks for the answers!

ollef:
Btw, this is only somewhat related (and affects the old V8 code as well), but should we be using some InnerException trick or something to get the right stack traces? I'm guessing you're the one that knows how that stuff works. :)

kusma:
@ollef: Yeah, I used InnerException here. Cleaning that up for V8 as well would be nice, but I decided to leave cleanups like that for later ;)

ollef:
I don't see any InnerExceptions here, but that's good to hear. 👍

kusma:
@olle: It's right here: https://github.com/fusetools/fuselibs-public/pull/662/files#diff-558f1496654f59f9d990cb200a3fe759R106

ollef:
Ah, there it is! :)

PR #663, opened 2017-10-31
Fix notification regression

cbaggers:
If you implement the notification methods on the `AppDelegate` then apple
detect this and require you to declare your usage of the
notification's feature, *even if* you aren't actually using it,
otherwise your app does not get approved.

Uno now wraps the AppDelegate methods in `#if` so we use declare to
include those methods when necessary. To use the changes in uno we
pull in newer uno stuff files.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
@cbaggers: When did this break? Has it been reported by users? Do we have a ticket?

cbaggers:
Related to this forum thread https://www.fusetools.com/community/forums/general/app_store_message?page=1&highlight=be30f004-7ec0-4a02-9d49-f50c7f972071#post-be30f004-7ec0-4a02-9d49-f50c7f972071

cbaggers:
IIRC booster has also seen this issue but I think that report was in the community slack, so it's gone now

kusma:
Superseded by #676

PR #664, opened 2017-10-31
Fix threading bugs in TreeObservable

### LINKED ISSUE #639: It is possible for TreeObservable to execute JS on the UI thread
## Reasoning that Reflect should run on the JS thread

- ThreadWorker Implements IMirror

- Implementing IMirror means implementing Reflect

- Reflect can call CreateMirror

- CreateMirror can create an Observable

- Observable's constructor calls Scripting.Context's CallbackToFunction method which eagerly executes some javascript (so it has to be on the JS thread)

- Therefore Reflect should take a Scripting.Context to guarentee it is on the correct thread.

## However

- TreeObservable has `Operation`s

- Inside some of the `Operation`'s `Perform` method there is a Call to the `Set` method of the parent type (`TreeObject`)

- `TreeObject`'s `Set` method calls `Reflect`

- `Perform` is run on the UI thread. (using `UpdateManager.PostAction(..)`)

- Therefore `Reflect` is called on the UI thread, which as we have seen can eagerly execute JS.

sebbert:
This fixes a couple of cases where JS would be called on the UI thread and vice versa.

- Parsing of arguments passed to TreeObservable operators now happens on the JS thread, and only the call to `Perform(object dc)` implemented by each `Operation` is dispatched onto the UI thread. This has the side-effect of no longer calling JS code from the UI thread with a cached `Scripting.Context` (fixes #639)
- Replacing the contents of a TreeObject using `Set(Scripting.Context, IMirror, Scripting.Object)` now updates UI state and notifies subscribers on the UI thread (as it should)
   - The same goes for TreeArray write-back operations

This PR contains:
- [ ] Changelog (N/A)
- [ ] Documentation (N/A)
- [ ] Tests (not really sure how one would go about writing meaningful tests for this)

### DIFF
--- a/Source/Fuse.Scripting.JavaScript/TreeObservable/TreeArray.uno
+++ b/Source/Fuse.Scripting.JavaScript/TreeObservable/TreeArray.uno
@@ -97,11 +97,10 @@ namespace Fuse.Scripting.JavaScript
 			{
 				var ta = SubscriptionSubject as TreeArray;
 
-				var worker = Fuse.Reactive.JavaScript.Worker;
 				var replaceAll = new ReplaceAllOperation((Scripting.Array)ta.Raw, values);
 				replaceAll.Perform(context);
 
-				ta.ReplaceAll(values, this);
+				UpdateManager.PostAction(new ReplaceAllOnUIThreadClosure(ta, values, this).Perform);
 			}
 
 			public void ReplaceAllExclusive(IArray values)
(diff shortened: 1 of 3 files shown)


sebbert:
`Operation` implements the protocol used from JavaScript to notify Uno that some state has changed. Every `TreeObservable` operator has some common arguments, which are handled by `Operation`.

> RemoveAtOperation, SetOperation, AddOperation & InsertAtOperation only run on JS thread (they are instantiated and Perform is immediately called)

Yes, the _argument parsing_ happens on the JS thread. [The actual implementation of each operator is dispatched on the UI thread via PerformClosure](https://github.com/fusetools/fuselibs-public/pull/664/files#diff-21137eae73b28ca7c2fc9fd5b3eaf310R161).

That being said, this stuff is definitely due for a refactoring. I'll see if I can come up with a pattern that makes this more readable, .

> Neither JSThreadSet or ReplaceAllOperation subclass Operation

Indeed. They are two different things.
`JSThreadSet` and `ReplaceAllOperation` handle writing back to JavaScript from Uno. They are simply closures. That being said, the names have kind of been neglected here :) Something like `SetClosure` and `ReplaceAllClosure` might be more fitting?

> Given this I'm thinking maybe they can be refactored out to helper methods which would avoid the whole 'make observable support context' thing

I don't quite understand what you mean. What does this have to do with `Observable`?

cbaggers:
Yup I missed that Perform (with context) can dispatch Perform (without context) down to the UI thread. That was the missing piece for me. I agree this could do with a refactor but also that this may not be the right commit for it.

PR #665, opened 2017-11-01
General-purpose blitter

kusma:
We blit a lot of images in Fuselibs. And draw-statements have made that very easy to open-code. However, the future of draw-statements in uno is a bit uncertain. So let's prepare a bit for moving away from them.

This PR introduce a general-purpose blitter, that has an interface that is trivial to implement with plain OpenGL code if needed.

The net-result is a reduction of 9 draw-statements, while keeping all features(*) intact.

### DIFF
--- a/Source/Fuse.Elements/Caching/Cache.uno
+++ b/Source/Fuse.Elements/Caching/Cache.uno
@@ -3,6 +3,7 @@ using Uno.Graphics;
 using Uno.UX;
 using Uno.Collections;
 
+using Fuse.Drawing;
 using Fuse.Nodes;
 
 namespace Fuse.Elements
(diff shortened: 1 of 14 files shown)


Source/Fuse.Elements/Caching/Cache.uno
```diff
@@ -215,31 +216,9 @@ namespace Fuse.Elements
...
-
-					float4 p: Vector.Transform(float4(Coord * Size, 0, 1), tile._compositMatrix);
-
-					ClipPosition: Vector.Transform(p, dc.Viewport.ViewProjectionTransform );
-
-					PixelColor: sample(tile.Texture, TexCoord, SamplerState.LinearClamp);
-
-					PixelColor: float4(prev.XYZ*opacity, prev.W*opacity);
-
-					CullFace: dc.CullFace;
```
kusma: So, this subtly breaks the Viewport.CullFace feature. However, it seems this is pretty heavily broken already; the vast majority of call-sites already ignore it. But I can't really understand why we would need a feature like that in the first place.

PR #669, opened 2017-11-01
JsErrorTest fixes

kusma:
These are a few incorrect assumptions I noticed while running the tests using JavaScriptCore instead of V8.


kusma:
@sebbert: That was a part of #557.

PR #676, opened 2017-11-02
Fix notification regression

cbaggers:
If you implement the notification methods on the AppDelegate then apple
detect this and require you to declare your usage of the
notification's feature, *even if* you arent actually using it,
otherwise your app does not get approved.

Uno now wraps the AppDelegate methods in `#if` so we use declare to
include those methods when neccessary

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@
 
 ## 1.4.0
 
+### Notifications
+- Fix regression causing iOS apps not to be accepted to the Store apparently due to use of push-notifications even though they are not used in the project.
+
 ### TextInput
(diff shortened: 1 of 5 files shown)


kusma:
Why was this filed as a new PR instead of just patching up #663? No need to change it, I just wonder why...

cbaggers:
because #663 was based on master and this was a hotfix for 1.4

kusma:
@cbaggers: You can change the base of a branch, and you can rebase a branch. So there wasn't any need for a new PR. I guess I'll just close #663, then.

cbaggers:
ah of course, I didnt think about how quickly fuselibs release branches are merged back into master.thanks

PR #681, opened 2017-11-03
Fuse.Nodes: do not try to perform layout every single frame

kusma:
There's no point in polling the invalidation-state every frame. So let's instead insert an event on every invalidate instead.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Nodes/RootViewport.uno
+++ b/Source/Fuse.Nodes/RootViewport.uno
@@ -54,7 +54,7 @@ namespace Fuse
 			RootInternal(null);
 
 			// Hook up global layout
-			UpdateManager.AddAction(PerformLayout, UpdateStage.Layout);
+			OnInvalidateLayout();
 		}
 
 		extern(Android || iOS)
@@ -82,12 +82,16 @@ namespace Fuse
 			RootInternal(null);
 
 			// Hook up global layout
-			UpdateManager.AddAction(PerformLayout, UpdateStage.Layout);
+			OnInvalidateLayout();
+		}
+
+		protected override void OnInvalidateLayout()
+		{
+			UpdateManager.AddOnceAction(PerformLayout, UpdateStage.Layout);
 		}
 		
 		void IDisposable.Dispose()
 		{
-			UpdateManager.RemoveAction(PerformLayout, UpdateStage.Layout);
 			Children.Clear();
 		}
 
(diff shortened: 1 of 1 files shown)


kusma:
A tiny bit of backstory:

This fixes a bunch of tests on Android for me, that fails like this:
```
FAILED:  AllTests.Fuse.Elements.Test.LayoutAnimationTest.TransitionLayout
          Exception was thrown: Uno.Exception: no Action found to remove
   at Fuse.UpdateManager.RemoveAction(Uno.Action,[Fuse.UpdateStage])
   at Fuse.RootViewport.Uno.IDisposable.Dispose()
   at FuseTest.TestRootPanel.Uno.IDisposable.Dispose()
   at Fuse.Elements.Test.LayoutAnimationTest.TransitionLayout()
   at MainClass.Invoke___Fuse_Elements_Test_LayoutAnimationTest_TransitionLayout()
   at Uno.Testing.AbstractRunner.RunTest(Uno.Testing.NamedTestMethod)
   at Uno.Testing.AbstractRunner.Update()
   at Uno.Testing.TestSetup.OnTick(object,Uno.Platform.TimerEventArgs)
   at Uno.Platform.Display.OnTick(Uno.Platform.TimerEventArgs)
   at Uno.Platform.AndroidDisplay.OnFrameCallback(double,double)
```

This doesn't really identify nor solve the underlying problem, just changes this code to be a bit less error-prone. I should probably try to find out why these tests fail in the first place also.

mortoray:
The reason this remained previously was for window size changes and device rotation -- we didn't accurately report them via events that would invalidate the layout.

I've tested OSX window resize and it still works. 

On iOS I have the problem that the ManualTestApp doesn't rotate at all now! I'll file an issue for this (as I don't think it's actually related to this PR, but alas, I can't properly test this PR then).

kusma:
@mortoray: To keep device-rotation working was the reason I changed `OnResized` also. I believe this should be how we ensure that this works. It seems to work fine on Android.

kusma:
So, I can no longer reproduce the crash that motivated me to do this change, so it's hard to figure out why the problem happened in the first place. Perhaps I had a bad build of some sort. In either case, I think this would be a good move, if we can prove that iOS rotation still works (after it's been fixed on master, that is)

kusma:
@mortoray: I've just verified that together with #700, device rotation on iOS works fine. Do I have your approval? :)

PR #687, opened 2017-11-06
Remove redundant _context checks during ThreadWorker creation

cbaggers:
- `RunInner` is private and can only be run from `Run`
- `Run` is private and is only every invoked from the constructor
- _context has no default value at construction so must be null at the start of `Run` & `RunInner`

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
+++ b/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
@@ -83,17 +83,15 @@ namespace Fuse.Scripting.JavaScript
 		{
 			try
 			{
+				_context = Fuse.Scripting.JavaScript.JSContext.Create();
+
 				if (_context == null)
 				{
-					_context = Fuse.Scripting.JavaScript.JSContext.Create();
-					if (_context == null)
-					{
-						throw new Exception("Could not create script context");
-					}
-					UpdateManager.AddAction(CheckAndThrow);
-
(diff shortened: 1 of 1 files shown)


Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -74,26 +74,21 @@ namespace Fuse.Scripting.JavaScript
 				Fuse.Diagnostics.UnknownException( "ThreadWorked failed", e, this );
 				_exceptionQueue.Enqueue(e);
 			}
-
-			if (_context != null)
-				_context.Dispose();
```
kusma: This seems wrong. Shouldn't we unconditionally Dispose here instead? Otherwise we'll *never* dispose, which seems like a bad choice...

Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -83,17 +83,15 @@ namespace Fuse.Scripting.JavaScript
 		{
 			try
 			{
+				_context = Fuse.Scripting.JavaScript.JSContext.Create();
+
 				if (_context == null)
 				{
```
kusma: While you're at it; I don't think this function can fail, so we shouldn't actually need this if-guard either...

PR #699, opened 2017-11-06
Fuse.Drawing: reduce textureTransform to float3x3

kusma:
A float3x3 is sufficient to express all two-dimentional linear
transformations, using 4x4 is just wasteful. Let's take advantage
of this, and save some computations.


kristianhasselknippe:
looks good to me

kristianhasselknippe:
that was quick :P

kusma:
image

PR #704, opened 2017-11-07
ThreadWorker's public api is safe so we can remove the _ready event

cbaggers:
Previously there were a number of public fields that would be
initialized during RunInner and so we couldnt trust other code to use
the ThreadWorker without potentially causing issues.

Now the public api revolves around Invoke with is a `ConcurrentQueue`,
thus I think we are in a good place to loosen the construction time
blocking we had previously

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
+++ b/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
@@ -18,7 +18,6 @@ namespace Fuse.Scripting.JavaScript
 
 		readonly Thread _thread;
 
-		readonly ManualResetEvent _ready = new ManualResetEvent(false);
 		readonly ManualResetEvent _idle = new ManualResetEvent(true);
 		readonly ManualResetEvent _terminate = new ManualResetEvent(false);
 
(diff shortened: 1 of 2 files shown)


Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -76,15 +73,8 @@ namespace Fuse.Scripting.JavaScript
...
-			{
-				_context = Fuse.Scripting.JavaScript.JSContext.Create();
-				UpdateManager.AddAction(CheckAndThrow);
-			}
-			finally
-			{
-				_ready.Set();
-			}
+			_context = Fuse.Scripting.JavaScript.JSContext.Create();
+			UpdateManager.AddAction(CheckAndThrow);
```
mortoray: This call to `AddAction` looks like an error. `RunInner` is being called from the thread, right?  The only safe UpdateManager call form a non-UI thread is `PostAction`.

kusma: @mortoray: This was already the case, though.

kusma: I have some changes in this area, I'll fix the problem as part of that.

cbaggers: Interesting, I'll open an issue for this

PR #706, opened 2017-11-07
ThreadWorker fixes

kusma:
Here's a couple of cleanups and bugfixes made possible by @cbaggers' awesome JSContext-work!

### DIFF
--- a/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
+++ b/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
@@ -73,7 +73,7 @@ namespace Fuse.Scripting.JavaScript
 
 		void RunInner()
 		{
-			_context = Fuse.Scripting.JavaScript.JSContext.Create();
+			_context = JSContext.Create();
 			UpdateManager.AddAction(CheckAndThrow);
 
 			double t = Uno.Diagnostics.Clock.GetSeconds();
--- a/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
+++ b/Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
@@ -14,8 +14,6 @@ namespace Fuse.Scripting.JavaScript
 
 	class ThreadWorker: IDisposable, IThreadWorker
 	{
-		JSContext _context;
-
 		readonly Thread _thread;
 
 		readonly ManualResetEvent _idle = new ManualResetEvent(true);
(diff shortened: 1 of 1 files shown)


Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -140,21 +133,24 @@ namespace Fuse.Scripting.JavaScript
...
+		void DispatchException(Exception e)
+		{
+			_exceptionQueue.Enqueue(e);
+			UpdateManager.PostAction(CheckAndThrow);
+		}
+
+		/*
+			Throws an exception that was generated on the thread.
 		*/
 		public void CheckAndThrow()
```
cbaggers: I think this could be private now. Otherwise LGTM :)

kusma: Yeah, good call!

kusma: @cbaggers: Fixed!

kusma:
Indeed!

Source/Fuse.Scripting.JavaScript/ThreadWorker.uno
```diff
@@ -140,21 +133,24 @@ namespace Fuse.Scripting.JavaScript
...
+		void DispatchException(Exception e)
 		{
-			Exception next = null, prev = null;
-			while (_exceptionQueue.TryDequeue(out next))
-			{
-				if (prev != null)
-					Fuse.Diagnostics.UnknownException("Skipped Exception", prev, this);
-				prev = next;
-			}
+			_exceptionQueue.Enqueue(e);
```
mortoray: This could probably be simplified. Instead of using a queue we could make a closure on the exception and post an action on that directly. It's a more straight forward 1:1 relationship and avoids the queue.

kusma: Yeah, I considered that, but it introduces new allocations. I don't really mind either way, though, as this only happens in error-cases. I can change if you think that'd be better... I don't really care.

mortoray: I think it might be better for a future reader of this code. The queue use is a bit odd now.

If you derive the closure from WrapException you could avoid any new allocation, but for this error path that's like the least of concerns I think.

kusma: OK, fix incoming ;)

kusma: @mortoray: Done :)

PR #710, opened 2017-11-07
Context renames

kusma:
I find it rather annoying to have several Context classes with the same name. The JSContext was already named, but the source-file wasn't yet. We can also rename the different implementations to ease source-file browsing.


kusma:
Nah, needs more work.

PR #720, opened 2017-11-09
Adds object based routing to the router/NavigationControl

mortoray:
This exposes some of the internal ability to use objects in the routing.

This PR contains:
- [x] Changelog
- [x] Documentation - but only `modifyPath`. Most of this feature is meant to be used in conjunction with Models, so there's no need to draw attention to it yet.
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.Navigation/NavigationControl.ScriptClass.uno
+++ b/Source/Fuse.Controls.Navigation/NavigationControl.ScriptClass.uno
@@ -12,7 +12,8 @@ namespace Fuse.Controls
 		{
 			ScriptClass.Register(typeof(NavigationControl),
 				new ScriptMethod<NavigationControl>("gotoPath", gotoPath, ExecutionThread.MainThread),
-				new ScriptMethod<NavigationControl>("seekToPath", seekToPath, ExecutionThread.MainThread));
+				new ScriptMethod<NavigationControl>("seekToPath", seekToPath, ExecutionThread.MainThread),
+				new ScriptMethod<NavigationControl>("modifyPath", modifyPath, ExecutionThread.MainThread));
 		}
 		
 		/**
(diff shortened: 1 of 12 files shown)


Source/Fuse.Controls.Navigation/NavigationControl.ScriptClass.uno
```diff
@@ -67,5 +68,138 @@ namespace Fuse.Controls
...
+				Fuse.Diagnostics.UserError( "`modifyPath` should be passed an object", nav );
+				return;
+			}
+
+			//reusing RouterRequest to ensure same meaning and defaults on supported arguments
+			var rr = new RouterRequest(RouterRequest.Flags.FlatRoute);
+			var keys = obj.Keys;
+			for (int i=0; i < keys.Length; ++i)
+			{
+				var key = keys[i];
```
kusma: This accesses JS objects from the UI thread, which is not legal.

I'm working on a threading-improved ScriptMethod API that discourages things like this, and I've already ported this class over to that. I'd prefer if we didn't land more users of the old API until that has landed, especially not when they clearly do threading-violations.

kusma: An easy way to see how this is broken, is to trigger a long-running task on the UI-thread (for instance layout), call modifyPath with an object, and immediately modify that in the calling JS code. This method will be queued up, but the argument will have changed before we get a chance to inspect it.

mortoray: I agree this should be fixed, but I don't think it should hold up this PR. This PR isn't keeping with what was previously done, so it's no worse at least, plus at the moment it has no atlernative way to achieve this. Blocking this PR though will in turn block work on navigation and the models release. The features in this PR themselves also won't change when your branch is ready, just their implementation; we wouldn't be accidentally committing to something we can't support.

kusma: It's worse, because there's one more tricky case to port. I do think this should hold this up, now that we're aware of the problem. This *does* introduce more of it. So no, I do think that this takes a step in the wrong direction.

To get this ready sooner, you could do what the new API will require: Run things on the JS-thread, extract the arguments needed and use a closure to dispatch to the other thread.

mortoray: When will your PR be ready?

kusma: In theory, I could send it now. I haven't migrated all code yet, but I'm very close. The old code still works (at least as well as it used to), but produce warnings.

However, the API isn't really the big deal, it's that we shouldn't access the arguments from the ui thread. That's just not going to work robustly. You can already fix that in this PR. If so, I don't see a problem with moving forward.

mortoray: I'm not sure doing parameter extraction on the JavaScript thread would be enough, I think the underlying objects must become mirrors.  Consider that the `object` types are used directly as the `Context` for data in navigation pages. There is no way those can be preprocessed before given to the recipient side.  

That is, I don't see a way this API could be rewritten to not assume it can access these objects safely from the UI thread. The magic has to happen behind the scenes.

The same thing happens in Model, where the JS objects (their Uno IObject), are passed around and accessed freely throughout the code. This was all done on the assumption that the data is being properly mirrored.

I don't want to change the way this code is written since there is no way to address this "context" object this way.

kusma: @mortoray: So to be clear, here's my current plan:

There will be two principal ways of getting called from `ScriptMethod`:
- `Func<Context, T, object[], object>`: This always runs on the JS-thread, and any other desired thread-dispatching is up to the user. This requires a return-value. Most call-sites just returns null.
- `Action<T>`: This always runs on the UI-thread, and is a simple way of doing hazard-free pulse-events, like Video.pause etc. This does not support any return-values.

These two comfortably covers a big portion of the API usage, but it does lead to some more explicit closures being created to dispatch things to the UI-thread.

In addition, I'm thinking that we can make a couple of variations on top of these two, that forwards the arguments to the UI-thread, by calling `JSContext.Reflect()` on each argument. This would be a pure convenience-thing, and can be omitted at first, to get things moving.

Anything advanced will have to bridge manually, and I think that should be OK. This PR is typically the kind of code that would have to do so.

You can already do the effective change: call `JSContext.Reflect()` to get mirrors of the arguments you want, and pass it to the UI-thread through a closure.

If you'd rather wait, I'm fine with that.

mortoray: I'm not sure I consider out corrent design of mirroring all arguments (the intended approach) to be bad. I actually consider our current API to be preferable to the one you are proposing, for a couple of reasons:

- Virtually all arguments to the functions need to be passed to the UI-thread at some point anyway.
- Having to manually do the dispatch on all of the functions is a wasted overhead, since it'll be the same pattern, same repeated code, at each function interface. I'm opposed to this since it's highly redundant and error-prone.
- Most of these functions do not use return values, and usually can't provide anything meaningful since it would involve locking the UI-thread to get those values. Thus supporting return values does not provide any significant value.
- Most functions take only simple arguments, there's no much lost in just mirroring those to the UI-thread always. The functions that take large arguments, like full object trees, are those used in navigation, and they still need the entire object mirrored as they use it for a context. It's reasonable then to just always mirror all of the arguments.

I'm opposed to your design because it'd increase the burden of writing script interfaces and not provide any gains. I agree there might be a problem with data races, but it can be solved without changing the script interface API (by mirroring in the arguments automatically). 

The entire code on top of this API, the script interfaces, expressions, and bindings, is all assuming these objects are mirrored and ready to be used on the UI-Thread.

kusma: It seems like you're misunderstanding things a bit:
- Our current design is *not* to mirror all arguments. It's to pass them verbatim. It wasn't the intentional design; mirroring wasn't even a concept when the current interface was introduced. Starting to mirroring them would be a breakage, because it would change the data passed to existing users.
- We're currently also pass `Scripting.Context` to users who aren't running on the UI-thread. This is also a threading-hazard.

The only change I propose is to fix these two things in a graceful way, by deprecating the old constructors to `ScriptMethod<T>`, and adding some new ones that doesn't pass the users a loaded gun pointing at them. Because we only want to pass `Scripting.Context` only to the non UI-thread, these will have slightly different interfaces, and `enum ExecutionThread` will no longer be needed.

I think this is a much less impactful change than you seem to suspect it is.

In addition:
- Manually dispatching will no longer be necessary for the vast majority of users (hopefully all) once I implement the auto-mirroring I talked about above. This is pretty easy to implement, but callers will have to be ported one by one to make sure they don't expect things like `Scripting.Object` and `Scripting.Array` in their arguments. We have a few methods that do that, I wouldn't be surprised if users also do.
- Even if I *don't* end up doing what I propose above (not saying I won't, easier porting is a big plus); yesterday I ported ~70 users of this API in a couple of hours *while* fiddling around with the interface. It's just not a big deal to port code like this. But I *will* try to make it easier than it currently is.
- The vast majority of the users on the UI-thread don't care about arguments; only about 20 of the users ended in my porting above ended up having to create closures.
- Sadly, it seems we accidentally broke compatibility in the `ScriptMethod<T>`-interface in 1.4.0 (when `Scripting.Context` was added to the signatures). I will try to make a patch for 1.4.1 to remedy this, so old code will still work.

mortoray: If you have a solution that doesn't involve significantly changing how the script interface functions actually work (signature adjustments aside), then I'm okay with that. But I'm just not seeing how we'll avoid the redundant dispatching and conversion of arguments. I need to see this before I can agree with it.

That said, I'd really like to has this PR merged before we make that decision. It doesn't save us any time by not merging now. Either way I'll have to migrate this code. Merging at least provides another set of tests that can be used to verify the new system works. If the new API is done soon we'll have this migrated prior to release, if the new API is not done soon then not merging this will hold up further work on nav and models.  That is, regardless of what we do with the API I don't see any downside to merging this now.

kusma:
OK, I've more or less finished up my branch, but I guess we can land this as-is, and fix it up together on my branch instead.

PR #721, opened 2017-11-09
ScriptException: Truncate long source lines

sebbert:
As this is kind of an unsolicited PR, feel free to discuss :)
In particular, the number `300` is chosen fairly arbitrarily, so we might want to be more conservative there.

My reasoning is that, if a single line of source code extends beyond 300 characters, it's either:

- A minified JavaScript library, where all the code is on a single line
- Very poorly written code

In any case, it will only clutter up the log, making the actual source of the error harder to find.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Scripting/ScriptException.uno
+++ b/Source/Fuse.Scripting/ScriptException.uno
@@ -10,6 +10,9 @@ namespace Fuse.Scripting
 		public string SourceLine { get; private set;}
 		public string JSStackTrace { get; private set;}
 
+		const int MaxSourceLineLength = 300;
+		const string SourceLineTooLongMessage = " ... source line truncated for readability ...";
+
 		public ScriptException(
 			string name,
 			string errorMessage,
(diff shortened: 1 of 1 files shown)


kusma:
So... `ScriptException.SourceLine` is a bit strange, for two reasons:
1. It only works on V8
2. It can be deduced from FileName + LineNumber (or maybe even `ScriptModule.Code`)

Basically, it's non-robust, and not all that helpful, really. Perhaps we just get rid of the property all together instead?

sebbert:
^ Agreed

kusma:
@sebbert: Here's the alternative: #749 

I don't *really* care which one we land. For instance, mine could be done as a clean-up on top of yours if you prefer that.

kusma:
This change has the benefit of touching *only* `ScriptException`, while solving the problem.

PR #728, opened 2017-11-13
ScriptClass thread-safety

kusma:
Our current ScriptMethod API is handing a loaded gun to the user:
- `ScriptMethod(..., ExecutionThread.MainThread)` pass Context and args down that aren't thread-safe to do anything with, because the user-code does not execute on the JavaScript-thread.
- `ScriptMethod(..., ExecutionThread.Any)` is quite pointless; running a method on a thread takes time away from that thread. We should always care about what thread we do this to. So instead, methods who don't care about what thread they're on should simply create their own thread. With blackjack and... yeah.

The end result here is that we no longer peek on Fuse.Scripting.Object / Fuse.Scripting.Array on the UI-thread. This means a bunch of smaller threading-issues in our script-classes are fixed now.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
+++ b/Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
@@ -10,8 +10,8 @@ namespace Fuse.Controls
 		static EdgeNavigator()
 		{
 			ScriptClass.Register(typeof(EdgeNavigator),
-				new ScriptMethod<EdgeNavigator>("dismiss", dismiss, ExecutionThread.MainThread),
-				new ScriptMethod<EdgeNavigator>("open", open, ExecutionThread.MainThread));
+				new ScriptMethod<EdgeNavigator>("dismiss", dismiss),
+				new ScriptMethod<EdgeNavigator>("open", open));
 		}
 
 		/**
@@ -19,7 +19,7 @@ namespace Fuse.Controls
 			
 			@scriptmethod dismiss()
 		*/
-		static void dismiss(Context c, EdgeNavigator e, object[] args)
+		static void dismiss(EdgeNavigator e, object[] args)
 		{
 			if (args.Length != 0)
 			{
@@ -36,7 +36,7 @@ namespace Fuse.Controls
 			@scriptmethod open(edge)
 			@param edge The enum name of the edge to open @Fuse.Navigation.NavigationEdge
 		*/
-		static void open(Context c, EdgeNavigator e, object[] args)
+		static void open(EdgeNavigator e, object[] args)
 		{
 			if (args.Length != 1)
 			{
--- a/Source/Fuse.Navigation/Router.ScriptClass.uno
+++ b/Source/Fuse.Navigation/Router.ScriptClass.uno
@@ -192,7 +192,7 @@ namespace Fuse.Navigation
 				return;
 			}
 			
-			var request = new ScriptRouterRequest();
+			var request = new RouterRequest(RouterRequest.Flags.FlatRoute);
 			
 			var keys = obj.Keys;
 			for (int i=0; i < keys.Length; ++i)
@@ -203,18 +203,6 @@ namespace Fuse.Navigation
 			}
 
 			request.MakeRequest(r);
-		}
-
-		class ScriptRouterRequest : RouterRequest
-		{
-			public ScriptRouterRequest() : base(Flags.FlatRoute)
-			{
-			}
-			
-			protected override Node ParseNode(object value)
-			{
-				return value as Node;
-			}
 		}
 
 		/**
--- a/Source/Fuse.Navigation/Router.ScriptClass.uno
+++ b/Source/Fuse.Navigation/Router.ScriptClass.uno
@@ -8,7 +8,7 @@ namespace Fuse.Navigation
 		{
 			ScriptClass.Register(typeof(Router),
 				new ScriptMethod<Router>("bookmark", Bookmark),
-				new ScriptMethod<Router>("getRoute", GetRoute, ExecutionThread.MainThread),
+				new ScriptMethod<Router>("getRoute", GetRoute),
 				new ScriptMethod<Router>("goBack", GoBack),
 				new ScriptMethod<Router>("goto", Goto),
 				new ScriptMethod<Router>("gotoRelative", GotoRelative),
@@ -325,41 +325,52 @@ namespace Fuse.Navigation
 					// and so on
 				})
 		*/
-		static void GetRoute(Context c, Router r, object[] args)
+		static object GetRoute(Context c, Router r, object[] args)
 		{
 			if (args.Length != 1) 
 			{
 				Diagnostics.UserError("Router.getRoute(): must provide exactly 1 argument.", r);
-				return;
+				return null;
 			}
 			var callback = args[0] as Function;
 			if (callback == null) 
 			{
 				Diagnostics.UserError("Router.getRoute(): argument must be a function.", r);
-				return;
+				return null;
 			}
 
-			var route = r.GetCurrentRoute();
-			c.Invoke(new GetRouteCallback(route, callback).Run);
+			UpdateManager.PostAction(new GetRouteCallback(c.ThreadWorker, r, callback).RunUI);
+			return null;
 		}
+
 		class GetRouteCallback
 		{
-			readonly Route _route;
+			readonly IThreadWorker _threadWorker;
+			readonly Router _router;
 			readonly Function _callback;
-			public GetRouteCallback(Route route, Function callback)
+
+			Route _route;
+
+			public GetRouteCallback(IThreadWorker threadWorker, Router router, Function callback)
 			{
-				_route = route;
+				_threadWorker = threadWorker;
+				_router = router;
 				_callback = callback;
 			}
 
-			public void Run(Scripting.Context context)
+			public void RunUI()
 			{
-				_callback.Call(context, ToArray(context));
+				_route = _router.GetCurrentRoute();
+				_threadWorker.Invoke(RunJS);
 			}
 
-			Array ToArray(Scripting.Context context)
+			public void RunJS(Scripting.Context context)
 			{
-				var route = _route;
+				_callback.Call(context, ToArray(context, _route));
+			}
+
+			static Array ToArray(Scripting.Context context, Route route)
+			{
 				var len = route.Length;
 				var arr = context.NewArray(len*2);
 				for (int i = 0; i < len; i++)
--- a/Source/Fuse.Navigation/Router.ScriptClass.uno
+++ b/Source/Fuse.Navigation/Router.ScriptClass.uno
@@ -7,14 +7,14 @@ namespace Fuse.Navigation
 		static Router()
 		{
 			ScriptClass.Register(typeof(Router),
-				new ScriptMethod<Router>("bookmark", Bookmark, ExecutionThread.MainThread),
+				new ScriptMethod<Router>("bookmark", Bookmark),
 				new ScriptMethod<Router>("getRoute", GetRoute, ExecutionThread.MainThread),
-				new ScriptMethod<Router>("goBack", GoBack, ExecutionThread.MainThread),
-				new ScriptMethod<Router>("goto", Goto, ExecutionThread.MainThread),
-				new ScriptMethod<Router>("gotoRelative", GotoRelative, ExecutionThread.MainThread),
-				new ScriptMethod<Router>("modify", Modify, ExecutionThread.MainThread),
-				new ScriptMethod<Router>("push", Push, ExecutionThread.MainThread),
-				new ScriptMethod<Router>("pushRelative", PushRelative, ExecutionThread.MainThread));
+				new ScriptMethod<Router>("goBack", GoBack),
+				new ScriptMethod<Router>("goto", Goto),
+				new ScriptMethod<Router>("gotoRelative", GotoRelative),
+				new ScriptMethod<Router>("modify", Modify),
+				new ScriptMethod<Router>("push", Push),
+				new ScriptMethod<Router>("pushRelative", PushRelative));
 		}
 
 		/**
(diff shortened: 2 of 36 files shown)


Source/Fuse.Navigation/Router.ScriptClass.uno
```diff
@@ -82,22 +82,22 @@ namespace Fuse.Navigation
...
+		static RouterPageRoute GetRelative(Router r, object[] args)
 		{
 			if (args.Length < 1)
 			{
 				Diagnostics.UserError( "Router.gotoRelative(): requires 1+ parameters", r );
 				return null;
 			}
 			
-			var node = c.Wrap(args[0]) as Node;
+			var node = args[0] as Node;
```
mortoray: Is this `Wrap` taken care of implicitly now?

Source/Fuse.Navigation/Router.ScriptClass.uno
```diff
@@ -218,6 +222,27 @@ namespace Fuse.Navigation
...
 				return _context.Wrap(value) as Node;
 			}
+
+			public void DiscardContext()
+			{
+				_context = null;
+			}
+		}
+
+		class MakeRequestClosure
```
mortoray: I wonder if RouterRequest could also just have an optional  `Router` and then use `MakeRequest` as the action. It'd save an allocation and remove the need for this class.

Source/Fuse.Navigation/Router.ScriptClass.uno
```diff
@@ -175,21 +175,21 @@ namespace Fuse.Navigation
 				- `relative`: An optional node that indicates the path is relative to this router outlet. The path is specified like in `gotoRelative`
 				- `style`: The style of the operation, which can be used as a matching criteria in transitions.
 		*/
-		static void Modify(Context c, Router r, object[] args)
+		static object Modify(Context c, Router r, object[] args)
```
mortoray: What is the significance of the `object` return? Do signatures with `Context` still run on the JS thread?

Source/Fuse.Navigation/Tests/UX/RouteParamValidation.ux
```diff
@@ -32,6 +32,15 @@
 			router.goto("page5", big_object);
 		};
 
+		module.exports.goto_test6 = function () {
+			var x = {
+				hello:"w3rld",
+				number:1338,
```
mortoray: So 1337 with objects? 1337++

Source/Fuse.Scripting/ScriptClass.uno
```diff
@@ -105,25 +112,107 @@ namespace Fuse.Scripting
...
+
+			@param name Name of method
+			@param method The native implementation of the method
+		*/
+		public ScriptMethod(string name, Action<T, object[]> method): base(name, ExecutionThread.JavaScript)
+		{
+			if (method == null)
+				throw new ArgumentNullException(nameof(method));
+
+			_method = new ArgumentMirrorClosure<T>(method).Run;
```
mortoray: Does this mean that all ScriptMethod functions will have their arguments automatically "mirrored" to the UI-thread?

Source/Fuse.Scripting/ScriptClass.uno
```diff
@@ -105,25 +112,107 @@ namespace Fuse.Scripting
...
+		{
+			readonly Action<T, object[]> _action;
+			public ArgumentMirrorClosure(Action<T, object[]> action)
+			{
+				_action = action;
+			}
+
+			public object Run(Context c, T obj, object[] args)
+			{
+				var marahalledArguments = new List<object>();
```
mortoray: Is the marahall where nightmares (mara) live? ;P

Source/Fuse.Scripting/ScriptClass.uno
```diff
@@ -105,25 +112,107 @@ namespace Fuse.Scripting
...
+				foreach (var arg in args)
+					marahalledArguments.Add(c.Reflect(arg));
+
+				UpdateManager.PostAction(new CallWithArgumentsClosure(_action, obj, marahalledArguments.ToArray()).Run);
```
mortoray: I wonder if we should allow PostAction to also take an IAction object? Any place using closure classes could then implement it and we'd save an allocation of the action.

Source/Fuse.Triggers/Timeline.ScriptClass.uno
```diff
@@ -27,7 +27,7 @@ namespace Fuse.Triggers
 			
 			@scriptmethod pulse()
 		*/
-		static void pulse(Context c, Timeline n, object[] args)
+		static void pulse(Timeline n)
```
mortoray: What happens in the JS thread if the user calls these functions without arguments? They should get an error (as functions that take arguments can check and emit errors).

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: Yes, at least in this case. Wrap is about bridging a set of known types between the JS and UI thread. We're nod doing mirroring, which also bridge objects, but with slightly different semantics. It seems the biggest difference is the handling of `Scripting.External` , which I'm not even sure what is... There's some minor details about float, int and uint as well, but in those cases, simple casts in the user-code seems more appropriate.

However, it's worth noting that in this particular case, this change was already a nop. Context.Wrap for "unknown" types is just a nop, and that's what happened here.

Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
```diff
@@ -10,16 +10,16 @@ namespace Fuse.Controls
...
 
 		/**
 			Closes any open edge panels.
 			
 			@scriptmethod dismiss()
 		*/
-		static void dismiss(Context c, EdgeNavigator e, object[] args)
+		static void dismiss(EdgeNavigator e, object[] args)
 		{
 			if (args.Length != 0)
```
mortoray: This function is actually a no-argument function and should lose the `args` parameter -- assuming we still report an error on calling it.

mortoray: There are a few more such functions, anything doing `if (args.length != 0)`.

Source/Fuse.Navigation/Router.ScriptClass.uno
```diff
@@ -175,21 +175,21 @@ namespace Fuse.Navigation
...
-			var obj = args[0] as Fuse.Scripting.Object;
+			var obj = args[0] as IObject;
 			if (obj == null)
 			{
 				Fuse.Diagnostics.UserError( "`Router.modify` should be passed an object", r );
-				return;
+				return null;
 			}
 			
 			var request = new ScriptRouterRequest(c);
```
mortoray: I'm assuming this function is currently running in the JS thread, meaning it's arguments haven't be reflected into the UI thread yet. I don't see any changes in the `RouterRequest` that does the reflection. Am I missing something?

Source/Fuse.Navigation/Router.ScriptClass.uno
mortoray: Why does this function take a `Context` but `NavigationControl.modifyPath` not take a `Context`? Is it required here only to the `Wrap` call on the nodes?  Do we even need `Wrap` anymore, or does reflection take care of it.

Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
kusma: We don't report errors when dropping the arguments. I tried doing this, but a bunch of tests started failing due to unexpected diagnostics.

Thinking about this a bit more, I'm a bit puzzled why that happened in the first place. Repeating the experiment indicates that I did some error in doing so. So I can clean this up, it seems.

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: object is just a return-value, and is only available on things that run on the JS-thread (same as with Context).

The reason why this takes a Context is that ScriptRouterRequest takes a Context. Perhaps it wouldn't have to? If so, we could get rid of the closure.

Source/Fuse.Navigation/Tests/UX/RouteParamValidation.ux
kusma: I just wanted to modify the values from the previous test. 1337 is just as arbitrary.

Source/Fuse.Scripting/ScriptClass.uno
kusma: If they run on the UI-thread and take arguments, yes.

Source/Fuse.Scripting/ScriptClass.uno
kusma: whups! Thanks for noticing ;)

Source/Fuse.Scripting/ScriptClass.uno
kusma: I'm not sure what that would solve. I could also just stuff the arguments inside ArgumentMirrorClosure<T>, and add a separate method on the UI-thread to save the allocation...

kusma: I'm not sure I understand. We need an allocation per invocation as well. What allocation would be avoided?

Source/Fuse.Triggers/Timeline.ScriptClass.uno
kusma: Nothing, like before.

I can add a general error-case here if we really want to. So far I opted to try not to change any behavior in this PR.

Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
kusma: I would prefer if we could do behavioral changes like this as a follow-up, though.

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: Yeah, this method might have been incorrectly ported.

kusma: Yeah, I think I have a better version now. I'll fix up some other small things an push out...

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: That sounds like a nice cleanup. Can we do that after landing this, though?

kusma: I rewrote this method to no longer need MakeRequestClosure, which should void this...

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: Done.

Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
mortoray: Sure, if you create an issue to do it afterward. :)

Source/Fuse.Triggers/Timeline.ScriptClass.uno
mortoray: You can add this to the issue about converted the `args != 0` functions into zero-arg functions. I think it's quite important we start issuing errors on these. Allowing args hurts future compatibility.

Source/Fuse.Controls.Navigation/EdgeNavigator.ScriptClass.uno
kusma: https://github.com/fusetools/fuselibs-public/issues/733

Source/Fuse.Scripting/ScriptClass.uno
mortoray: The `.Run` bit creates a binding object implicitly. If we added a `IAction` interface we'd avoid that overhead.

In profiling the creation of these action bindings does appear a lot. However, in the JS->Uno functions there's already so much overhead it probably won't be noticed.

Source/Fuse.Navigation/Router.ScriptClass.uno
mortoray: It was using the context to call `Wrap` for the `Node`. But if you're saying that was a no-op anyway, chances are we don't need this context here, and we don't need this function to run on the JS thread at all.

Source/Fuse.Triggers/Timeline.ScriptClass.uno
kusma: Yeah, good point. I've added that detail to the ticket.

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: Yeah. I've updated this in this PR already, and that made the method a lot more pleasant to read.

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: Nevermind, I just re-read the code after eating some food, and `Scripting.External` is indeed handled in the mirroring also. So I think the semantics are the same already.

Source/Fuse.Navigation/Router.ScriptClass.uno
```diff
@@ -201,22 +201,19 @@ namespace Fuse.Navigation
...
-				base( Flags.FlatRoute )
+			public ScriptRouterRequest() : base(Flags.FlatRoute)
 			{
-				_context = context;
 			}
 			
 			protected override Node ParseNode(object value)
 			{
-				return _context.Wrap(value) as Node;
+				return value as Node;
```
mortoray: This function could just be replaced now with `as Node` in the base class. It's purpose was just to abstract access to the `Context`, which is now gone. Hmm, possibly the entire `ScriptRouterRequest` is redundant now, it just passes an argument to the base constructor.  Maybe it could be cleaned up later, but leaving as is might be confusing (what purpose does it serve?)

kusma: Right, yeah. I'll remove this.

Source/Fuse.Scripting/ScriptClass.uno
kusma: Ah, I see. Yeah, perhaps there's some use-cases where this would make sense, but I doubt this is that one.

Source/Fuse.Navigation/Router.ScriptClass.uno
kusma: Fixed!

kusma:
@mortoray: I believe all your feedback have been addressed. What do you think?

PR #729, opened 2017-11-13
Fix path case issue

yupferris:
This didn't cause any issues technically, but is confusing in the source
tree.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
@yupferris: It actually might have caused issues with people using case-sensitive HFS+ on macOS. However, that is a configuration that I believe is rather buggy anyway.

PR #736, opened 2017-11-13
No more `texture2d.Load`

kusma:
We don't really need texture2D.Load any more, so let's get rid of a few usages.


kusma:
Guess I forgot to add some ux-files, I'll fix!

PR #737, opened 2017-11-13
Fuse.Scripting: give error on unconsumed args

kusma:
In the argument-less case, we can make a generic error if arguments
were passed to a method that doesn't care about arguments.

This ensures calling code is forward-compatible in the case where we
want to add arguments to existing methods.


mortoray:
Wait, I think this needs a CHANGELOG entry. If somebody happened to pass an argument to one of those functions before, it will now produce an error and do nothing instead of being ignored.

kusma:
Good point, I'll add this!

PR #742, opened 2017-11-14
Fuse.Scripting.JavaScript: do not make mirror contain JS objects

kusma:
When we truncate while reflecting the JS state, we accidentally left
JS objects in the mirror, because we weren't able to tell the difference
between truncation and primitive-types. So let's rewrite the code a bit
to be able to tell the difference.

Because the mirroring gives up after 50 recursions, make the
RouterRequest validation give up after 49 instead, so we retain the
same errors even when truncating.

### DIFF
--- a/Source/Fuse.Navigation/RouterRequest.uno
+++ b/Source/Fuse.Navigation/RouterRequest.uno
@@ -274,7 +274,7 @@ namespace Fuse.Navigation
 		
 		static bool ValidateParameter(object arg, int depth = 0)
 		{
-			if (depth > 50)
+			if (depth > 49)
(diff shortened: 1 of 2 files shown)


kusma:
@mortoray: I'm not really in love with the way I keep the errors in `RouterRequest`, but this was the only thing I could come up with. Do you have a better idea?

mortoray:
It's probably okay, but I'd like to see the magic number disappear (otherwise somebody will change this and hte point of your fix will be lost).

kusma:
We'd at least have a failing test if it does... Still far from awesome, though.

PR #751, opened 2017-11-15
Reevaluate builtins

kusma:
This will force builtins to be reevaluated, which fixes some problems
with Polyfills/Window not being available when an app that uses
the JavaScript-tag but *not* FuseJS gets rooted before one that uses
both.

This fixes problems with tests using AllTests.unoproj on feature-Models,
but the problem should also be possible to trigger from preview or other
carefully crafted apps.


sebbert:
@kusma Two tests are failing on AppVeyor

kusma:
I fixed this problem in a different way on feature-Models. I might revisit this for testing-sanity in the future, but not as this PR.

PR #754, opened 2017-11-17
Feature/nulls

mortoray:
Adds a few useful UX functions: hasValue and isNull

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 # Unreleased
 
 ## Expressions
-- Added `hasValue` to check if a value is known in the context
+- Added `isDefined` to check if a value is known in the context
 - Added `isNull` to check if a value is null or doesn't exist
 
 ## Router
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Expressions
+- Added `hasValue` to check if a value is known in the context
+- Added `isNull` to check if a value is null or doesn't exist
+
 ## Router
 - Added object support to `Router` script functions, such as `goto`, `push`, `bookmark`, etc. This mirrors the upcoming Model ability to use objects as path elements.
 - Added object support to `Modify/Push/GotoRoute` actions.
--- a/Source/Fuse.Reactive.Bindings/Tests/UX/Expression.HasValue.ux
+++ b/Source/Fuse.Reactive.Bindings/Tests/UX/Expression.HasValue.ux
@@ -1,0 +1,4 @@
+<Panel ux:Class="UX.Expression.HasValue">
+	<FuseTest.Data ux:Name="a"/>
+	<FuseTest.DudElement BoolValue="hasValue({Property a.Value})" ux:Name="ha"/>
+</Panel>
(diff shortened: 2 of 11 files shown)


Source/Fuse.Reactive.Expressions/UnaryOperator.uno
```diff
@@ -63,21 +73,32 @@ namespace Fuse.Reactive
...
-			protected virtual void OnNewOperand(object value)
+			
+			void UpdateOperands()
 			{
 				ClearDiagnostic();
 
 				try
 				{
-					_uo.OnNewOperand(_listener, value);
+					if ( (_hasValue || _uo.IsOperandOptional) )
```
kusma: There's a needless set of parens here..

Source/Fuse.Reactive.Expressions/UnaryOperator.uno
```diff
@@ -63,21 +73,32 @@ namespace Fuse.Reactive
 
 			protected override void OnNewData(IExpression source, object value)
 			{
-				OnNewOperand(value);
+				if (source == _uo.Operand) { _hasValue = true; _value = value; }
```
kusma: please add some newlines here for readability...

Source/Fuse.Reactive.Expressions/UnaryOperator.uno
```diff
@@ -63,21 +73,32 @@ namespace Fuse.Reactive
...
+				UpdateOperands();
 			}
 			
 			protected override void OnLostData(IExpression source)
 			{
-				_listener.OnLostData( _uo );
+				if (source == _uo.Operand) { _hasValue = false; _value = null; }
```
kusma: Same, please add some newlines

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Expressions
+- Added `hasValue` to check if a value is known in the context
```
kusma: Are we sure this is the best name? How about `isDefined` or something instead? This would match the `null` vs `undefined`-distinction that JavaScript has, at least...

mortoray: Don't know, I asked in #apireviews and didn't get a response. Perhaps `isDefined` is better?!

Source/Fuse.Reactive.Expressions/UnaryOperator.uno
mortoray: Yeah, that'd be my preference to, but this matches the preexisting form in Binary/Ternanry/QuaternarnryOperator that I didn't do.  I was trying to keep it looking exactly the same in case somebdoy ever realized how to merge all four of them into one class.

Source/Fuse.Reactive.Expressions/HasValue.uno
```diff
@@ -0,0 +1,117 @@
...
+					_opSub.Dispose();
+					_opSub = null;
+				}
+				_listener = null;
+			}
+		}
+	}
+	
+	[UXFunction("isNull")]
+	/** Returns true if the value exists and is non-null, false otherwise.
```
Duckers: Is this documentation correct? Sounds inverted to me

mortoray: Ah, yeah, inverted. Changed the name forget the docs.

Source/Fuse.Reactive.Bindings/Tests/UX/Expression.HasValue.ux
```diff
@@ -1,4 +1,4 @@
 <Panel ux:Class="UX.Expression.HasValue">
 	<FuseTest.Data ux:Name="a"/>
-	<FuseTest.DudElement BoolValue="hasValue({Property a.Value})" ux:Name="ha"/>
+	<FuseTest.DudElement BoolValue="isDefined({Property a.Value})" ux:Name="ha"/>
```
kusma: This rename didn't make its way to the changelog, it seems...

PR #774, opened 2017-11-21
Fix TimePicker properties

yupferris:
Originally, they were erroneously stored on the view, which led to some invalid conditions where a property would be set before the view was ready and the value lost. Now, following other native controls like Slider, values are stored on the semantic control, and propagated to the view.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.TimePicker/TimePicker.uno
+++ b/Source/Fuse.Controls.TimePicker/TimePicker.uno
@@ -13,55 +13,59 @@ namespace Fuse.Controls
 
 	interface ITimePickerView
 	{
-		DateTime Value { get; set; }
-		bool Is24HourView { get; set; }
+		DateTime Value { set; }
+		bool Is24HourView { set; }
 
 		void OnRooted();
 		void OnUnrooted();
 	}
 
-	interface ITimePickerHost
-	{
-		void OnValueChanged();
-	}
-
-	public abstract partial class TimePickerBase : Panel, ITimePickerHost
+	public abstract partial class TimePickerBase : Panel
 	{
 		static Selector _valueName = "Value";
 
+		DateTime _value = DateTime.UtcNow;
 		[UXOriginSetter("SetValue")]
 		/**
 			Gets or sets the current time value selected by the `TimePicker`.
 		*/
 		public DateTime Value
 		{
-			get
-			{
-				var tpv = TimePickerView;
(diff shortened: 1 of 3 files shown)


Source/Fuse.Controls.TimePicker/TimePicker.uno
```diff
@@ -100,9 +96,19 @@ namespace Fuse.Controls
...
+			{
+				tpv.Value = Value;
+				tpv.Is24HourView = Is24HourView;
+			}
+		}
+
+		void ITimePickerHost.OnValueChanged(DateTime newValue)
 		{
-			OnValueChanged(this);
+			SetValue(newValue, null);
```
mortoray: The `null` looks wrong, it should be the view I think. There's a condition in `SetValue` that checks the origin to prevent sending the message back-there.

yupferris: Yeah I thought that a bit strange too, but I wrote it following this line in `RangeControl.uno`: https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Controls.Primitives/RangeControls/RangeControl.uno#L246 . Perhaps they should both be changed?

mortoray: Yeah, it looks like that is wrong as well. We still have lots of inconsistencies with value handling in our controls unfortunately.

yupferris: Gotcha. I'll go ahead and make a PR for that specific case too then.

yupferris: https://github.com/fusetools/fuselibs-public/pull/778

kusma:
> Source\Fuse.Controls.TimePicker\TimePicker.uno(111.4): E3128: Call to 'Fuse.Controls.TimePickerBase.SetValue(Uno.DateTime,Uno.UX.IPropertyListener)' has some invalid arguments (Uno.DateTime,Fuse.Controls.ITimePickerView)

... I guess this is one of the reasons why you need to deal with IPropertyListener?

yupferris:
> ... I guess this is one of the reasons why you need to deal with IPropertyListener?

Precisely, same with #778. Thanks for adding the `blocked` labels.

PR #778, opened 2017-11-21
Fix null sender in RangeControl

yupferris:
I don't know of any bugs this actually caused, but the logic is clearly a bit strange. This should use the NativeView as the sender when calling SetValue to avoid value propagation loops.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Primitives/RangeControls/RangeControl.uno
+++ b/Source/Fuse.Controls.Primitives/RangeControls/RangeControl.uno
@@ -243,7 +243,7 @@ namespace Fuse.Controls
 
 		void IRangeViewHost.OnProgressChanged(double newProgress)
 		{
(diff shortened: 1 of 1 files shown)


kusma:
> Source\Fuse.Controls.Primitives\RangeControls\RangeControl.uno(246.4): E3128: Call to 'Fuse.Controls.RangeControl.SetValue(double,Uno.UX.IPropertyListener)' has some invalid arguments (double,Fuse.Controls.Native.IRangeView)

... I guess this is one of the reasons why you need to deal with IPropertyListener?

yupferris:
This is actually not correct; we shouldn't be shoehorning extra functionality into the UX property system. The code here is still a bit smelly, but I think a more complete/thought-out approach is required to fix that, and since no users are screaming for this, I'm going to drop it for now.

PR #791, opened 2017-11-23
Fuse.Reactive.Expressions: do not crash on null input

kusma:
Fixes fusetools/uno#1454

This PR contains:
- [x] Changelog
- [ ] ~Documentation~ Does not contain new features
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,12 @@
 
 # 1.4
 
+## 1.4.1
+
+### Expressions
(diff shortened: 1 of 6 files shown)


kusma:
It just changes from "crash" to "propagate the bad input". I don't think changing the behavior on clearly bad input is a bad change, neither in this PR nor in the next release...

kusma:
Yeah, I'm totally fine with this as just a stop-gap solution.

PR #793, opened 2017-11-23
introduction of experimental Let feature

mortoray:
This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Unreleased
 
+## Let
+- Introduced an experimental `Let` feature that allows creating expression aliases and local variables in UX
+
 ## Timer
 - Fixed issue where creating a repeating `Timer` with 0 delay in JavaScript would not prevent the worker thread to become idle.
 
 ## RangeControl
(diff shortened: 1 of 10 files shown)


Source/Fuse.Reactive.Bindings/Let.uno
```diff
@@ -0,0 +1,279 @@
...
+		The value is now part of the data context:
+		
+			<Slider Value="{a}"/>
+		
+		It can also be accessed directly outside of the context:
+		
+			<Slider Value="{Property a.Value}"/>
+			
+		
+		If you are using an expression it's recommended now to use `Expression` instead of value:
```
kristianhasselknippe: `Expression` instead of value =>  `Expression` instead of `Value`

Source/Fuse.Reactive.Bindings/Let.uno
```diff
@@ -0,0 +1,279 @@
...
+		
+			<Let Expression="{pos} + 5" ux:Name="p"/>
```
kristianhasselknippe: should we have ux:Name as the first attribute in these examples? as per the normal convention

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Let
+- Introduced an experiment `Let` feature that allows creating expression aliases and local variables in UX
```
kristianhasselknippe: experiment => experimental (?)

PR #794, opened 2017-11-23
remove old preview-selection code

kusma:
This functionality hasn't been used by the tool since Fuse 0.29, which
is more than a year old. So let's remove this, as the tool does a much
better job of this without our help.


kusma:
Technically speaking, this removes public API. But it's under the Fuse.Preview namespace, which I'd be very surprised if any users even knew about. I also doubt anyone could really have done anything useful with it.

So I think just yanking this out directly should be OK-ish. Thoughts?

mortoray:
I think we should at least add a CHANGELOG entry saying it's been removed, and that Fuse.Preview was never intended to be a public API (we've done this when hiding other internals previously). Otherwise I'm okay with this.

kusma:
Yeah, good point.

Issue #673, opened 2017-11-01
RangeControl.RelativeValue does not play nice with data-bound properties


eksperts:
Fuse 1.4.0 build 14679, macOS.

Repro:
```
<App>

    <AttractorConfig ux:Global="rangeSnap" Unit="Normalized" Type="SmoothSnap" />

    <DockPanel ux:Class="MySlider">
        <float ux:Property="Min" />
        <float ux:Property="Max" />
        <object ux:Property="Val" />

        <RangeControl ux:Name="_range" Value="{Property Val}"
            Minimum="{ReadProperty Min}" Maximum="{ReadProperty Max}" UserStep="10"
            Height="48" Margin="24,0">

            <LinearRangeBehavior />

            <Circle ux:Name="handle" Width="48" Alignment="Left" Anchor="50%,50%" X="attract({Property _range.RelativeValue}, rangeSnap) * 100%">
                <Text Value="{= round({ReadProperty _range.Value})}" FontSize="13" Alignment="Center" Color="#565cff" />
                <Circle Margin="8" Color="#fff" />
            </Circle>

        </RangeControl>

        <Rectangle Layer="Background" CornerRadius="4" Color="#565cff" />

    </DockPanel>

    <JavaScript>
    var Observable = require("FuseJS/Observable");
    var val = Observable(70);
    var min = 0;
    var max = 200;
    module.exports = {
        val: val,
        min: min,
        max: max
    };
    </JavaScript>

    <ClientPanel>
        <StackPanel Alignment="VerticalCenter" ItemSpacing="24">

            <MySlider Margin="24,0" Min="0" Max="200" Val="{val}" />

            <MySlider Margin="24,0" Min="{min}" Max="{max}" Val="{val}" />

        </StackPanel>
    </ClientPanel>

</App>
```
Gives you this result:
<img width="411" alt="screen shot 2017-11-01 at 18 24 35" src="https://user-images.githubusercontent.com/7188878/32285107-f6b22fde-bf31-11e7-8994-efbe93bda4bb.png">

The problem is that `_range.RelativeValue` in the `X="attract({Property _range.RelativeValue}, rangeSnap) * 100%"` on the drag handle for some reason does not initialise right when we data-bind the `Min` and `Max` properties of the custom component.

mortoray:
[...]

PR #799, opened 2017-11-27
Fix/range control.relative value

mortoray:
Corrects some behavior with using RangeControl and bound values. Fixes a related numeric stability issue in `attract`.

RangeControl no longer clamps the Value, only the behaviours do.

Fixes https://github.com/fusetools/fuselibs-public/issues/673

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## RangeControl
+- `RangeControl.Value` and `RangeControl2D.Value` are not longer clamped to the `Range` of the control. This fixes issues where the `Value` was incorrectly modified when the range and value were both data bound. The user behaviors `LinearRangeBehavior` and `CircularRangeBehavior` will both however clamp to the range -- the user cannot select outside the range.
+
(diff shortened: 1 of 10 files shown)


kusma:
I think it would be better to keep track of the unclamped-value, and rather re-clamp when min/max is set. That way we keep the same API, while still handling the use-case listed above...

mortoray:
I thought about that approach at first. It doesn't address several kinds of feedback loops that might involve the `Value` of this control, since the `Value` would still momentarily be clamped, and that clamped value propagates (which is relevant to the orignal example I had using `attract`). Even a two-way data binding has this problem, since the clamped value is fed back to the JS code.

I think this new approach is fair, not having semantic controls enforce these constraints. Consider that `UserStep` on this control is treated the same way -- so the fix also resolves this insoncistency. In terms of logic it's simpler, and I find saner, to just not apply modifications when setting a read/write value (it's just unexpected behaviour even if the bindings weren't being used).

PR #800, opened 2017-11-27
Fix idleness in ThreadWorker when using setTimeout( func, 0)

nekronos:
This PR contains:
- [x] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Timer
+- Fixed issue where creating a repeating `Timer` with 0 delay in JavaScript would not prevent the worker thread to become idle.
+
 ## RangeControl
 - `RangeControl.Value` and `RangeControl2D.Value` are not longer clamped to the `Range` of the control. This fixes issues where the `Value` was incorrectly modified when the range and value were both data bound. The user behaviors `LinearRangeBehavior` and `CircularRangeBehavior` will both however clamp to the range -- the user cannot select outside the range.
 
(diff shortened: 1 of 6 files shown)


nekronos:
@mortoray maybe? This is part of https://github.com/fusetools/fuselibs-public/pull/601, but doing it as a separate PR

kusma:
I think 6bca8da3eec667cd89dd8670b2c6c2c6c5ec074b should be squashed in here. Also, we should make a test-case for this.

mortoray:
Ah, I didn't realize this was on that branch where I did it.

nekronos:
From a quick look at the code I think writing meaningful tests would involve larger changes to `ThreadWorker`/`Timer` to make the code testable

kusma:
@nekronos: How so? Can't we just use a reasonably large timeout, and wait a reasonable amount of time?

nekronos:
Again, I dont know this code. Just looked like it would require some refactoring to make stuff testable as units. The `Timer`, `ThreadWorker` looks pretty coupled to `JSContext`

mortoray:
You could test the Test functions in the ThreadWorker class. THere are some exposed to check if the thread is idle. You could set a large timeout in JS and then check that it isn't idle. Of course, I don't know how you'd check when it is again. A short timetout would also cause a race condition.

kusma:
@nekronos: Perhaps this should be mentioned in the changelog? Or do you think it's insignificant enough to "normal" users to omit?

nekronos:
yeah, lets add some changelog

PR #802, opened 2017-11-28
fixing Bundle.list example

eksperts:
fixing syntax in the example snippet

This PR contains:
- [ ] Changelog
- [x] Documentation
- [ ] Tests

### DIFF
--- a/Source/FuseJS/Bundle.uno
+++ b/Source/FuseJS/Bundle.uno
@@ -81,10 +81,9 @@ namespace FuseJS
 			```
 			var Bundle = require("FuseJS/Bundle");
 			
-			Bundle.list()
-				.then(function(list) {
-					//list is an array of paths, such as "assets/image.jpg"
-				})
+			Bundle.list().then(function(list) {
+				//list is an array of paths, such as "assets/image.jpg"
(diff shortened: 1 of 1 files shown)


kusma:
@eksperts: dashes is not the right way of crossing things off the list. A checkbox is either checked or not. If you want to remove it, you can either remove the checkbox, or just ~strike-out~ the the label.

eksperts:
@kusma fixed

Source/FuseJS/Bundle.uno
```diff
@@ -85,7 +85,6 @@ namespace FuseJS
 				.then(function(list) {
 					//list is an array of paths, such as "assets/image.jpg"
 				})
-			}
```
kusma: The whole indentation of the .then etc seems off to me. Wouldn't it be cleaner to do
```js
Bundle.list().then(function(list) {
	//list is an array of paths, such as "assets/image.jpg"
})
```
?

eksperts: I agree that would look better. Wanted my fix to be as minimal as possible, but I can change it to what you suggest.

PR #804, opened 2017-11-28
document Android.NotificationIcon support

eksperts:
Adding a paragraph about Android.NotificationIcon support in `.unoproj` files. This has been out there for a while, but we're lacking docs.

This PR contains:
- [ ] Changelog
- [x] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.PushNotifications/Docs/Guide.md
+++ b/Source/Fuse.PushNotifications/Docs/Guide.md
@@ -127,7 +127,23 @@ For Android we can use exactly the same `'aps'` entry or the alternatively the f
 
 The `notification` entry is the standard Google way of doing this but we felt that it wouldn't hurt to support the Apple way too.
 
-> The current implementation only guarantees the `title` and `body` entries will work. We also always use your app's icon as the notification icon. This is an area we will extend as Fuse matures. If you have specific requests, be sure to let us know!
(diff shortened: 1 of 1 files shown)


Source/Fuse.PushNotifications/Docs/Guide.md
```diff
@@ -127,7 +127,23 @@ For Android we can use exactly the same `'aps'` entry or the alternatively the f
...
+        "Android": {
+            "NotificationIcon": {
+                "LDPI": "ldpi.png",
+                "MDPI": "mdpi.png",
+                "HDPI": "hdpi.png",
+                "XHDPI": "xhdpi.png",
+                "XXHDPI": "xxhdpi.png",
+                "XXXHDPI": "xxxhdpi.png"
+        }
+    }
```
kusma: Hmm, we usually document stuff like this [here](https://www.fusetools.com/docs/basics/uno-projects)...

I'm not saying this is a better place, but it'd be good if we were consistent. Otherwise it's easy to end up with multiple sets of docs of the same thing...

eksperts: fair point, I'll start with a change to that article and close this PR for now, then make a new one when those changes are published.

PR #806, opened 2017-11-28
Conversion functions

mortoray:
Adds `float(expr)` and `string(expr)` to force conversion.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Conversions
 - Added `float()` expression to force conversion to float values
+- Added `string()` expression to force conversion to string values
 
 ## Let
 - Introduced an experimental `Let` feature that allows creating expression aliases and local variables in UX
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## Conversions
+- Added `float()` expression to force conversion to float values
+
 ## Let
 - Introduced an experimental `Let` feature that allows creating expression aliases and local variables in UX
 
--- a/Source/Fuse.Reactive.Expressions/BinaryOperator.uno
+++ b/Source/Fuse.Reactive.Expressions/BinaryOperator.uno
@@ -27,14 +27,18 @@ namespace Fuse.Reactive
 		protected virtual bool IsLeftOptional { get { return false; } }
 		protected virtual bool IsRightOptional { get { return false; } }
 
-		protected virtual object Compute(object left, object right)
+		protected virtual bool Compute(object left, object right, out object result)
 		{
 			throw new Exception(GetType().FullName + " does not implement the required methods");
 		}
 
 		protected virtual void OnNewOperands(IListener listener, object left, object right)
 		{
-			listener.OnNewData(this, Compute(left, right));
+			object result;
+			if (Compute(left, right, out result))
+				listener.OnNewData(this, result);
+			else
+				listener.OnLostData(this);
 		}
 		
 		protected virtual void OnLostOperands(IListener listener)
--- a/Source/Fuse.Reactive.Expressions/BinaryOperator.uno
+++ b/Source/Fuse.Reactive.Expressions/BinaryOperator.uno
@@ -29,9 +29,14 @@ namespace Fuse.Reactive
 
 		protected virtual bool Compute(object left, object right, out object result)
 		{
-			throw new Exception(GetType().FullName + " does not implement the required methods");
+			Fuse.Diagnostics.Deprecated( " No `Compute`, or a deprecated form, overriden. Migrate your code to override the one with `bool` return. ", this );
+			result = Compute(left, right);
+			return true;
 		}
-
+		
+		/** @deprecated Override the other `Compute` function. 2017-11-29 */
+		protected virtual object Compute(object left, object right) { return null; }
+		
 		protected virtual void OnNewOperands(IListener listener, object left, object right)
 		{
 			object result;
--- a/Source/Fuse.Reactive.Expressions/UnaryOperator.uno
+++ b/Source/Fuse.Reactive.Expressions/UnaryOperator.uno
@@ -20,14 +20,27 @@ namespace Fuse.Reactive
 
 		protected virtual bool IsOperandOptional { get { return false; } }
 		
-		protected virtual object Compute(object operand)
+		/**
+			@param result the result of the computation
+			@return true if the value could be computed, false otherwise
+		*/
+		protected virtual bool Compute(object operand, out object result)
 		{
 			throw new Exception(GetType().FullName + " does not implement the required methods");
 		}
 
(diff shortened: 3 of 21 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+## Conversions
+- Added `float()` expression to force conversion to float values
+- Added `string()` expression to force conversion to string values
```
kusma: I think these should be named toFloat() and toString(), as `float` and `string` seems not that unlikely to become keywords at some point.... It also communicates a bit better what they do IMO. Or are you thinking about these like constructors?

mortoray: It was the syntax duckers first propose to me. I think it makes sense as we won't likely introduce actual object construction. But in that sense, these can be well interpreted as conversion constructors.

There should never be a symbol overlap in this scope. It's controlled what is export to it -- names don't leak around.

kusma: Well, my point was that if "float" and "string" becomes keywords, this probably won't compile any more, and if it would, these functions would not be reached through them. Yeah, I'm aware that such a change would already be backwards-incompatible, but I don't really want us to paint ourselves further into a corner.

That being said, I don't really feel strongly about this. It was just my first reaction, that I expected these to be called something else.

mortoray: I think the idea was that these are essentially keywords. The just don't need to be programmed at the UX compiler level and work fine as functions.  That is, any introduction of a keyword would reasonably have the same behaviour, so I think we'd be fine. Not that I see any reason for those eventual keywords (this is just an expression language, not a full programming language).

Let me ask in #ux first. It's no big deal to change the name now.

kusma: I think you're right that introduction of this as keywords would be reasonable to expect to do the same, so my objection is probably moot.

yupferris: yeah I agree with @mortoray here; if we want `float` and `string` as fn's it essentially reserves the symbols anyways in the same way keywords would be, and if we ever did anything with the same symbols later we'd want to play nice with these kinds of features anyways. So I'm on board 👍

Source/Fuse.Reactive.Expressions/UnaryOperator.uno
```diff
@@ -20,14 +20,27 @@ namespace Fuse.Reactive
 
 		protected virtual bool IsOperandOptional { get { return false; } }
 		
-		protected virtual object Compute(object operand)
+		/**
+			@param result the result of the computation
+			@return true if the value could be computed, false otherwise
+		*/
+		protected virtual bool Compute(object operand, out object result)
```
kusma: This breaks public API :/

Source/Fuse.Reactive.Expressions/BinaryOperator.uno
```diff
@@ -27,14 +27,18 @@ namespace Fuse.Reactive
 		protected virtual bool IsLeftOptional { get { return false; } }
 		protected virtual bool IsRightOptional { get { return false; } }
 
-		protected virtual object Compute(object left, object right)
+		protected virtual bool Compute(object left, object right, out object result)
```
kusma: This also breaks public API :/

mortoray: I think I can add the old API and mark it deprecated (they have different overload paraemters so it shouldn't conflict).

kusma: If you can find a way that works, then that would be great. It's not clear to me how we'd know which version to call, though.

yupferris: mark the old one as deprecated?

kusma: @yupferris: If you're asking how we do that, the answer is that we use the `[Obsolete]`-attribute in uno.

mortoray: I've pushed the backwards compatible version. It can't issue a compile-time error, but will emit a runtime error about compatibility.

Source/Fuse.Reactive.Expressions/BinaryOperator.uno
```diff
@@ -27,14 +27,23 @@ namespace Fuse.Reactive
...
 		{
-			throw new Exception(GetType().FullName + " does not implement the required methods");
+			Fuse.Diagnostics.Deprecated( " No `Compute`, or a deprecated form, overriden. Migrate your code to override the one with `bool` return. ", this );
+			result = Compute(left, right);
```
kusma: I'm not sure this quite does the trick. What happens if a user overrides for instance BinaryFloatOperator, which already overloads this? Or NameValuePair? Or InfixOperator?

mortoray: Those already had their own set of virtual functions that shouldn't be affected by this I believe.

kusma: I guess we should mark the Compute method as `sealed` on these, so the user can't override them, then.

kusma: OK, seems like this is not as bad as I thought. We can do this to fix the theoretical problems in fuselibs:
```patch
diff --git a/Source/Fuse.Reactive.Expressions/MathFunctions.uno b/Source/Fuse.Reactive.Expressions/MathFunctions.uno
index 83c2067309..4dadcd52c2 100644
--- a/Source/Fuse.Reactive.Expressions/MathFunctions.uno
+++ b/Source/Fuse.Reactive.Expressions/MathFunctions.uno
@@ -152,7 +152,7 @@ namespace Fuse.Reactive
 			_name = name;
 			_op = op;
 		}
-		protected override object Compute(object operand)
+		protected sealed override object Compute(object operand)
 		{
 			float4 v;
 			int size;
@@ -193,7 +193,7 @@ namespace Fuse.Reactive
 			_name = name;
 			_op = op;
 		}
-		protected override object Compute(object left, object right)
+		protected sealed override object Compute(object left, object right)
 		{
 			double lv = 0;
 			double rv = 0;
diff --git a/Source/Fuse.Reactive.Expressions/NameValuePair.uno b/Source/Fuse.Reactive.Expressions/NameValuePair.uno
index 9460843397..30f5ed622f 100644
--- a/Source/Fuse.Reactive.Expressions/NameValuePair.uno
+++ b/Source/Fuse.Reactive.Expressions/NameValuePair.uno
@@ -3,7 +3,7 @@ using Uno.UX;
 namespace Fuse.Reactive
 {
 	/** Creates a `Fuse.NameValuePair` from a name and a value. */
-	public class NameValuePair: BinaryOperator
+	public sealed class NameValuePair: BinaryOperator
 	{
 		[UXConstructor]
 		public NameValuePair([UXParameter("Name")] Expression name, [UXParameter("Value")] Expression value) : base(name, value)
```

PR #811, opened 2017-11-29
Pre-models cleanups

kusma:
Here's a couple of small cleanups from the models-branch that are independently a good idea.

This PR contains:
- [ ] Changelog - *Not sure; the `_subscription == null`-case is the only one that might be worth logging, but I'm not sure it really can trigger*
- [ ] Tests - *Same thing as above, unsure if the functional change can actually trigger, or how to do so*

### DIFF
--- a/Source/Fuse.Scripting.JavaScript/FuseJS/Builtins.uno
+++ b/Source/Fuse.Scripting.JavaScript/FuseJS/Builtins.uno
@@ -75,10 +75,10 @@ namespace Fuse.Reactive.FuseJS
 
 		internal bool UpdateModules(Fuse.Scripting.Context context)
 		{
-			var activity = false;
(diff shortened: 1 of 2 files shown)


kusma:
@mortoray: Do you have any idea how to trigger the `_subscription = null`-case? I suspect that two-way binding to a non-existent data-binding or something might do the trick...

mortoray:
I believe that was coming from a race condition (or it may have been something similar). Something had unsubscribed while there was a value in flight (such as unrooting during animation). It may have been an intermittent problem due to JS/Uno races.

Some of the null checks were added for defensive reasons, since Dispose clears them, I just wanted to be safe. Some are reall do happen though at odd times (damn JS async during cleanup!!)

PR #813, opened 2017-11-29
Travis: install pip

nekronos:
Looks like pip is not part of xcode command line tools anymore, could not find any details in apples changelogs. But on my fresh high sierra install I was also missing pip

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ env:
     - XCODE_XCCONFIG_FILE=$TRAVIS_BUILD_DIR/.travis.xcconfig
 
 install:
+  - sudo easy_install pip
   - pip install six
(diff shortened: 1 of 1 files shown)


nekronos:
Looks like this is the cause: https://blog.travis-ci.com/2017-11-21-xcode8-3-default-image-announce

kusma:
Another option could be to use `osx_image: xcode7.3` to get the old behavior... I like staying with a modern version, though.

yupferris:
Yeah, and installing it explicitly feels a bit better than relying on what's installed by that package already to some degree, in case it gets added/removed again. With so few install steps currently it shouldn't be a problem.

PR #815, opened 2017-11-30
Add node as a stuff dependency

nekronos:
FuseJS requires node to function properly. Unfortunately, Node got a deep hierarchy of files when installed, which is larger than MAX_PATH on Windows, so this means `git clean -dxf` will complain. However `stuff clean` should work fine. `Stuff.exe` also had to be updated to support long paths.

The node.stuff file will also be shipped with fuselibs, so that Studio can use it as well.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


sebbert:
Again, why is this against fuselibs?

Tapped:
I think a compromise here is that fuselibs got the stuff file, but doesn't install it. But I think it makes sense that fuselibs owns node since fuselibs are depending on it, and gonna depend more on it.

kusma:
@Tapped: Fuselibs doesn't even know about node, it's uno that depends on node.

mortoray:
@nekronos Was a decision made about this being the right place for this?

kusma:
This should either go in Uno or Fuse, because:
- Uno is what actually uses Node, not fuselibs
- Fuse is what installs user-dependencies (like Android SDK etc)

I haven't heard a good argument for adding this here, so I'm closing this.

PR #835, opened 2017-12-04
Fuse.Reactive.Expressions: fixup ToUpper / ToLower tests

kusma:
This broke when merging release-1.4 into master, due to divergent changes


kusma:
@mortoray: Yeah. My plan here was to add tests on master for the functionality.

Issue #840, opened 2017-12-05
Intermitant test hanging on android


cbaggers:
- fuselibs master (ed476aeb at time of testing)
- uno (1597f6fea as newer has separate problems)
- `uno test -tandroid --run-local ./AllTests.unoproj`

tests run file for a while but eventually get caught in one test and test runner times out:

```
FAILED:  AllTests.Fuse.MarshalTests.SizeParserTest
         Exception was thrown: Test timed out after 10.780831 seconds
FAILED:  AllTests.UnknownTest351
         Exception was thrown: Previous test timed out
FAILED:  AllTests.UnknownTest352
         Exception was thrown: Previous test timed out
.. etc ..
```

Saw the issue in __ previously and @kusma suggested https://github.com/fusetools/fuselibs-public/commit/35c1984b1ba91d63696f60ae97dadd84981bf7c7 but give the `SizeParserTest` this could be something else.

cbaggers:
General notes on problem:

cbaggers:
I found the issue, details and fix coming real soon

cbaggers:
Please see this commit https://github.com/fusetools/fuselibs-public/pull/846/commits/b64095302a4fce1a5abe07b68927abbd81c20daf for something that 'fixed' the issue for these specific tests, the fundamental issue of errors at module-eval-time not being emitted is not fixed however and will need addressing.

I'll leave this issue open to track that

cbaggers:
Hmm may be better to look at the PR that commit was in https://github.com/fusetools/fuselibs-public/pull/846 There is some strangeness going on that is probably not explained by our original theory

cbaggers:
I'm having to put this down to focus on other work. These spurious failures still remain and we will want to fix them before we turn on mobile testing on TC

PR #846, opened 2017-12-06
'Fix' hang in ScriptPromise tests

cbaggers:
Previous there were 3 promise tests sharing the same test panel.

the panel has a javascript tag

the javascript tag would, at module eval time, try and dispatch some
promises.

.e.g this top level expression `customPanel.getString().then(...)`

however the test only populated the promise it was interested in, so
I believe the other expressions would try and call `then` on null.

This seems to leave the js in a rather undefined that so that a few
times in a hundred the promise never resolves and we get a hang
resulting in the tests timing out.

By refactoring these tests to each own their own panel and
bootstrapping we avoid the issue.

However the lack of an error is a deeper issue that is being tracked
here: https://github.com/fusetools/fuselibs-public/issues/840

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests


kusma:
I'm not a big fan of this approach. The reason is that we end up duplicating code. All we would have needed would be to fill in these promises with some dummies, no?

cbaggers:
@kusma Sounds fair. I'll try that now

cbaggers:
@kusma I think my theory of what was wrong may be wrong. I tried both 'set the promises at the start of each test' approach e.g.
```
 		{
 			child.customPanel.StringFuture = new Promise<string>("this is a string");
+			child.customPanel.ObjectFuture = new Promise<TestObject>(new TestObject());
+			child.customPanel.NumberFuture = new Promise<double>(0.0);
```
and also the 'give the properties defaults' approach
```
-		public Future<string> StringFuture { get; set; }
-		public Future<double> NumberFuture { get; set; }
-		public Future<TestObject> ObjectFuture { get; set; }
+		Future<string> _stringFuture = new Promise<string>("ignored promise");
+		Future<double> _numberFuture = new Promise<double>(0.0);
+		Future<TestObject> _objectFuture = new Promise<TestObject>(new TestObject());
+
+		public Future<string> StringFuture
+		{
+			get { return _stringFuture; }
+			set { _stringFuture = value; }
+		}
+		public Future<double> NumberFuture
+		{
+			get { return _numberFuture; }
+			set { _numberFuture = value; }
+		}
+		public Future<TestObject> ObjectFuture
+		{
+			get { return _objectFuture; }
+			set { _objectFuture = value; }
+		}
```

and both still resulting in the fail test hanging fairly frequently.

However the approach in this pr still works. Very odd

mortoray:
You can probably simplify this by using `FuseTest.DudElement`. It lets you capture bindings by various types without creating a custom class.  Often the ux:Property's can interfere with what you're actually trying to test.

Source/Fuse.Scripting/Tests/ScriptClass/ScriptStringPromise.Test.uno
```diff
@@ -0,0 +1,46 @@
...
+	public class ScriptStringPromiseTest : TestBase
+	{
+		[Test]
+		public void StringTest()
+		{
+			var child = new UX.StringBasicTest();
+			child.customStringPanel.StringFuture = new Promise<string>("this is a string");
+			using (var root = TestRootPanel.CreateWithChild(child))
+			{
+				while (string.IsNullOrEmpty(child.StringProperties.StringResult))
```
mortoray: I'd like to see a tighter requirement on this. I'm thinking a `MultiStepFrameJS(2)` would be correct here. We expect only the first JS to run and request the future, then the UI thread produce the value, then one more JS step to get the value out.

kusma: Yeah, this was my crappy attempt at making this more robust, but I think it's been shown that I misidentified the problem.

mortoray:
@cbaggers Were you going to return to this PR or is it no longer relevant?

mortoray:
I'll assume this is abandoned.

PR #859, opened 2017-12-08
Fuse.Video.CIL: Build separately for Mac and Win

ollef:
Create separate configurations for building the Fuse.Video.CIL C#
project on Mac and Windows, and load the right assemblies in Uno
depending on the platform.

This allows us to conditionally include only the projects that are
needed for the specific platform depending on the configuration we're
building, and also means that the Platform detection module is not
needed anymore since it's all statically determined.

This fixes an issue where the Fuse.Video.CIL assemblies failed to load
on Mac with recent versions of Uno because it tried to load
Windows-specific assemblies that weren't there.

This PR contains:
- [ ] ~Changelog~ Not user-facing, just means we can build with upcoming versions of Uno.
- [ ] ~Documentation~ N/A
- [ ] ~Tests~ We notice this pretty quickly as all dotnet builds fail on Mac without it.


ollef:
I opted for not porting this code to Uno (which @kusma has said to me elsewhere that he preferred) since it uses some C# PInvoke stuff which isn't trivial to port. That's not to say it can't be done, it just means it'd require some work.

kusma:
@ollef: That's totally fair :)

Besides, a smaller change is always less scary.

PR #861, opened 2017-12-08
Model: Attach to parent on property update

### LINKED ISSUE #844: Model is not updating correctly in the UI
In [this sample project](https://github.com/mortoray/fork-sandbox/tree/master/Dec2017/dice), reduced from a real-world test, the data on-screen isn't updating correctly. 

Reproduction:
- You have a list with one item
- Click `Standard` to add a standard item
- Click `Modifier` to add a modifier item
- the list shows three correct items
- Click `Standard` to add another standard item
- Error: The third item now also reads `standard`, it should be `modifier` as the item has not changed
- Clicking on the item (invoking the JS edit call) corrects the issue

sebbert:
Fixes #844, as well as a number of other bugs relating to how Model.js manages parent-child relationships.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests


kusma:
@sebbert: Any chance for a test-case to prevent future regressions?

mortoray:
The code in #844 should be small enough for a test case.

PR #875, opened 2017-12-15
merge redundant code in expressions

mortoray:
This removes a lot of the redundant code from the expressions. It makes it simpler to implement new expressions, and fixes a few hidden defects in the previously divergent code.

Deprecations warnings will be produced at runtime in most cases (since it can't be detected statically).

There is one removal of old "technically public" API from UnaryOperator. A compatbile addition could be made, but it'd be messy, and those funcions should not have been exposed to begin with (they don't work correctly, nor did Binary/Ternary/QuaternaryOperator expose them).

I didn't manage to simplify all the bits I wanted yet, but this seems like a good merge point.

This PR contains:
- [x] Changelog
- [x] Documentation
- [ ] ~Tests~ Already covered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,12 +1,13 @@
 # Unreleased
 
-## Navigation
-- Fixed an issue with `Navigator.Pages` not registering pages correctly in certain initialization orders
-- Added `$navigationRequest` to `Navigation.Pages` objects. This can be used to fine-tune the navigation.
 ## Expressions
 - Deprecated `UnaryOperator.OnNewOperand` and `OnLostOperand`.  These are part of a broken pattern of using unary expressions (and were not present on Binary/Ternary/QuaternaryOperator). You generally shouldn't need this, and should implement `Compute` instead. In the rare cases you need the vrituals you'll need to extend Expression and implement `Subscribe`, using `ExpressionListener` as a way to capture the correct functionality.
 - Moved `VarArgFunction.Argument` to `Expression.Argument`. It's in a base class so still has visibility in `VarArgFunction`.
 - `VarArgFunction.Subscription.OnNewData` and `OnLostData` have been sealed. They should not have been open for overriding before as it conflicts with the inner workings on the class. Only the `OnNewPartialArguments` and `OnNewArguments` should be overridden.
+
+## Navigation
+- Fixed an issue with `Navigator.Pages` not registering pages correctly in certain initialization orders
+- Added `$navigationRequest` to `Navigation.Pages` objects. This can be used to fine-tune the navigation.
 
 ## Instance
 - Added `Instance.Item` to work similar to an `Each` with a single data item
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 - Fixed an issue with `Navigator.Pages` not registering pages correctly in certain initialization orders
 - Added `$navigationRequest` to `Navigation.Pages` objects. This can be used to fine-tune the navigation.
 ## Expressions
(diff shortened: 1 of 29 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased
 
+## Expressions
+- Removed `UnaryOperator.OnNewOperand` and `OnLostOperand`.  This were part of a broken pattern of using unary expressions (and were not present on Binary/Ternary/QuaternaryOperator). You will need to extend Expression and implement `Subscribe`, using `ExpressionListener` as a way to capture the correct functionality. It was already broken before, this forces a fix to be made.
```
kusma: Can't we mark this as obsolete for a few releases first so code implementing this will still compile, but give a warning?

mortoray: Is it possible, but it's relatively tricky to manage, and would only provide a runtime warning. I need to pass a deprecated flag to the base-class that changes how it works, basically breaking out into a different model (Which might be a tad bit broken -- it was broken before as well)

mortoray: I'll push my attempt as a separate commit, then you can see if it's a good idea to retain.

mortoray:
@kusma Introduced the removed functions as deprecated. Added a test since the implementation is a bit mad.

Source/Fuse.Reactive.Expressions/UnaryOperator.uno
```diff
@@ -34,6 +37,12 @@ namespace Fuse.Reactive
 		/** @deprecated Override the other `Compute` function. 2017-11-29 */
 		protected virtual object Compute(object operand) { return null; }
 
+		protected override sealed bool Compute(Argument[] args, out object result)
+		{
+			return Compute(args[0].Value, out result);
+		}
+		
+		/** @deprecated Override `Compute` or don't derive from `UnaryOperator` if you need argument tracking (which is rare). The typical base would be `Expression` and create a `Subscription` derived from `ExpressionListener`. 2017-12-14 */
 		protected virtual void OnNewOperand(IListener listener, object operand)
```
kusma: The reason why you can't mark it as `[Obsolete("Override Compute instead")]` is that this would introduce warnings in the calling code, `InternalOnNewOperand`, right?

kusma: By the way, I really wish we used `bool TryCompute(Argument[] args, out object result)` rather than `bool Compute(Argument[] args, out object result)`, so this wouldn't be so confusing. Is that too late to change? I don't think we've shipped this API-change yet...

mortoray: Yes, I have no strict path of obsolete entry points that I could mark the functions as obsolete. Non-obsolete ones need to call them for them to work. :/

As to `TryCompute`, yeah, it'd be better. There just used to be a `Compute` method in some of the operators. Let me verify we never shipped anything with `Compute` yet (since having `Compute` and `TryCompute` might be confusing)

mortoray: The `Unary/Binary/Ternary/Quaternary.Compute` virtuals have been there a long time. Some were marked `deprecated` before.  I'll see if there are enough new methods to justify calling it `TryCompute`.

mortoray: Or maybe not, the latest methods were introduced in https://github.com/fusetools/fuselibs-public/commit/ed7ed4f401dbb1ae0108dc15c94bb6705dd4ea6a  Has that been released?

kusma: By the way, I guess we should consider adding support `#pragma warning disable XXX` etc to Uno so we could use obsolete code without having to rewire code like crazy until the full support can be removed...

mortoray: Yes, some pragma support would be nice. I'd love to see a language with actual features for library release cycles at some point.

BTW, I changed `Compute` to `TryCompute`

Issue #878, opened 2017-12-18
move custom mono-build to CDN-neutral URLs


kusma:
The mono-build referred to by [.travis.yml](../blob/master/.travis.yml) needs to be hosted on a CDN-neutral URL.

Issue #877, opened 2017-12-18
move XamMac-related stuff-file to CDN-neutral URLs


kusma:
The following file needs to be hosted on CDN-neutral URLs:
* [Source/Fuse.Controls.Video/CIL/XamMac.stuff](../blob/master/Source/Fuse.Controls.Video/CIL/XamMac.stuff)

Issue #497, opened 2017-09-26
move websocket-related stuff-file to CDN-neutral URLs


kusma:
The following file needs to be hosted on CDN-neutral URLs:

[Source/Fuse.WebSockets/ios/SocketRocket/SocketRocket-iOS.stuff](../blob/master/Source/Fuse.WebSockets/ios/SocketRocket/SocketRocket-iOS.stuff)

Issue #496, opened 2017-09-26
move V8-related stuff-files to CDN-neutral URLs


kusma:
The following files needs to be hosted on CDN-neutral URLs:
* [Source/Fuse.Scripting.V8/lib/Android.stuff](../blob/master/Source/Fuse.Scripting.V8/lib/Android.stuff)
* [Source/Fuse.Scripting.V8/lib/DotNet.stuff](../blob/master/Source/Fuse.Scripting.V8/lib/DotNet.stuff)
* [Source/Fuse.Scripting.V8/lib/OSX.stuff](../blob/master/Source/Fuse.Scripting.V8/lib/OSX.stuff)
* [Source/Fuse.Scripting.V8/lib/Windows.stuff](../blob/master/Source/Fuse.Scripting.V8/lib/Windows.stuff)

PR #882, opened 2017-12-19
Move dependencies to new CDN

kusma:
Fixes issue #878.
Fixes issue #877.
Fixes issue #497.
Fixes issue #496.
Fixes issue #495.


kusma:
Don't merge yet. I'm currently running a custom TC-build to ensure that all build-targets still work.

PR #884, opened 2017-12-22
New <Desktop /> trigger

LuisRodriguezLD:
Triggers if run on an Desktop preview. 

Now you can do: 
```
<Panel ux:Name="panel" Background="#F00" >
    <Desktop>
        <Change panel.Background="#0F0" />
    </Desktop>
</Panel>
```

**Use case:** I want some elements to be visible in iOS while others in Android, since I am using both `<Android />` and `<iOS/>` the Desktop preview becomes empty.
Now I can using `<Desktop />` to show whatever I need.

### DIFF
--- a/Source/Fuse.Triggers/Platforms.uno
+++ b/Source/Fuse.Triggers/Platforms.uno
@@ -72,4 +72,40 @@ namespace Fuse.Triggers
 			base.OnUnrooted();
 		}
 	}
+
+	/**
+		Triggers if run on Desktop
+
+		## Example
+
+		This example sets a panel's background color to green if the app is
+		running on Desktop. If the app is ran on another platform, it will be
+		red:
+
+			<Panel ux:Name="panel" Background="#F00" >
+				<Desktop>
+					<Change panel.Background="#0F0" />
+				</Desktop>
+			</Panel>
+	*/
+	public class Desktop: Trigger
+	{
+		protected override void OnRooted()
+		{
+			base.OnRooted();
+			if defined(!Android && !iOS)
+			{
+				Activate();
+			}
+		}
+
+		protected override void OnUnrooted()
+		{
+			if defined(!Android && !iOS)
+			{
+				Deactivate();
+			}
+			base.OnUnrooted();
+		}
+	}
 }


Sunjammer:
On the surface this looks good to me as a fix, but I think a better long term solution would be a more general mechanism that allows mixing of these flags, say a `<Platform Is="iOS, Desktop"/>` or similar (names are hard).

kusma:
I don't like this. It makes it look like desktop is a supported export-target, which it isn't really.

I also think the semantics will be technically wrong; if running in Android emulation on ChromeOS, we'd be incorrectly reporting as non-desktop.

In general, I expect the lines between mobile and desktop to blur more in the future, making distinctions like these hard to support. We should instead strive to make tools that makes it easy to detect the platform-features that matter, and respond reasonably to that.

LuisRodriguezLD:
I agree. It _might_ confuse some users. How about changing the name? Maybe, `<Fallback />`?  Or I can try @Sunjammer suggestion `<Platform Is=“iOS, Desktop”`

kusma:
I think `ux:Condition` is really what you want here. Sadly, that's a long-standing, unimplemented planned feature.

But, we need to get out of the "exposing discrete platforms"-mindset, and focus on high-level, semantic tools. I'd prefer if we didn't cement the "desktop" vs "mobile" idea any more...

mortoray:
I think we have do occassionally have need of this trigger. Full cross-platform support will always be lagging slightly behind and such a trigger could be a decent stop-gap measure.

That said, I agree with @kusma that there is a naming problem. "Desktop" isn't right here, especially since Preview is meant to mimic Mobile. But I don't know what a good name for this is -- I'm okay with something long/ugly since it should rarely be used.

LuisRodriguezLD:
How about `<NotMobile />`? I don't know, like @Sunjammer said, names _are_ hard. I personally find many uses for this implementation, however it's your call guys.

mmkevin:
how about `<Simulator> </Simulator>` ??

mortoray:
If we do such a trigger I'd prefer a generic `Platform` trigger that provides for a series of tests and conditions as properties allowing for combinations as desired.

kusma:
Yeah, I think a Trigger that takes (for instance) an UX expression would make sense for this purpose.

mortoray:
@kusma Yes, expression functions are probably the best. To make them effective I need to add an `IsEnabled` or `IsActive` to the `Instance` behaviour.  Using them with `WhileTrue` would be bad since it by default takes instances, not templates, thus preinstantiating per-platform stuff.  And option to make triggers use templates is also probably a good thing (General purpose).

mortoray:
I think we're in disagreement on this PR and would prefer a solution like https://github.com/fusetools/fuselibs-public/issues/962

PR #886, opened 2017-12-28
Fix/instantiator sanity

mortoray:
This is a long-needed refactoring of the way `Instantiator` works. This clarifies how it works and makes it possible to add new features, and fix some defects.

The first commit was squashed. The inidividual bits done during the refactoring didn't produce logically useful in between steps, but can still be viewed at: https://github.com/fusetools/fuselibs-public/compare/master...mortoray:fix/instantiator-sanity-steps?expand=1
There was no real way to do any sort of part-way PRs in this refactoring as the original code was far too intertwined.

The refactoring fixed two minor defects (test-cases added), and ended up introducing one minor new functionality about `MatchKey` and `Observable` (this was just the easiest way to do some of the simplification in the refactoring). It means that `Instance` could also be greatly reduced now.

This PR contains:
- [x] Changelog - minimal
- [x] Documentation
- [x] Tests - mostly covered by existing tests


Source/Fuse.Reactive.Bindings/ItemsWindowList.uno
```diff
@@ -0,0 +1,328 @@
...
+							RemovedDataAt(op.A + Offset);
+							InsertWindowItem(DataToWindowIndex(op.A + Offset), op.Data);
+						}
+						break;
+				}
+			}
+		}
+		
+		
+		void IObserver.OnSet(object newValue)
```
nekronos: `newValue` should not be used here?

mortoray: These interface values aren't necesarily needed. The underlying Observable has already gogten the new value, and `ItemWindowsList`  updates itself from the backing observable.

Should this be indicated somehow in this code? I agree it looks a bit weird.

PR #905, opened 2018-01-03
Feature `Element.Position` and `Size`, and `size()` expression function

mortoray:
Adds some new functionality for simplified Element positioning/animation in some scenarios.

The Position/Size are hidden by derived classes in a few places. A few have new names/deprecated the old. Only `Video` doesn't have a new name. We can't solve this immediately (other than to deprecate Video.Progress with a new name). User's will have to use the old properties, or wrap the item in a panel.

This PR contains:
- [x] Changelog
- [x] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Marshal/Tests/Marshal.Test.uno
+++ b/Source/Fuse.Marshal/Tests/Marshal.Test.uno
@@ -169,6 +169,7 @@ namespace Fuse
 			Assert.AreEqual(r*Marshal.ToFloat4(v), Marshal.ToFloat4(Marshal.Multiply(vr, v)));
 			Assert.AreEqual(r/Marshal.ToFloat4(v), Marshal.ToFloat4(Marshal.Divide(vr, v)));
 
+			//It's somewhat questionable that a float4 can convert to a Size/Size2, but alas, it was supported before :(
 			Assert.AreEqual(new Size2(r.X, r.Y), Marshal.ToSize2(v));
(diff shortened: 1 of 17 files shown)


Source/Fuse.Marshal/Tests/Marshal.Test.uno
```diff
@@ -80,42 +80,42 @@ namespace Fuse
...
-
-			TestVector(7.0f, float4(7.0f, 7.0f, 7.0f, 7.0f));
-			TestVector(7.0, float4(7.0f, 7.0f, 7.0f, 7.0f));			
-			TestVector(7, float4(7.0f, 7.0f, 7.0f, 7.0f));
-
-			TestVector((Size)7.0f, float4(7.0f, 7.0f, 7.0f, 7.0f));
-			TestVector(new Size2(7.0f, 8.0f), float4(7.0f, 8.0f, 7.0f, 8.0f));
-
-			TestVector(float2(7.0f, 8.0f), float4(7.0f, 8.0f, 7.0f, 8.0f));
+// 			TestScalar(13.0f);
```
nekronos: Why are these tests commented out?

mortoray: Because I foolishly forgot to comment them back in after searching for a defect. I'll fix that.

kusma:
This broke our [master integration](https://tc.outracks.com/viewType.html?buildTypeId=MasterIntegration_Build).

PR #909, opened 2018-01-04
ImageSource: added Transform property

nekronos:
Useful for ImageSources that needs to be transformed to display
correctly. For example photos captured with frontfacing camera

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+## ImageSource
+- Added `Transform` property. Useful for ImageSources that requires a transform to be displayed correctly (For example photos caputred with front faceing camera)
+
 ## Let
 - Added specific type version of `Let`, such as `LetFloat` and `LetString`. This improve the ability to connect pieces of the UX together and do animation/transitions in UX without using JavaScript.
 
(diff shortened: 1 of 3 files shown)


kusma:
Can't the same thing already be done by modifying `Source.Orientation` instead?

nekronos:
[`Source.Orientation`](https://github.com/fusetools/fuselibs-public/blob/master/Source/Fuse.Elements/Resources/Exif/Exif.uno#L12) is based on EXIF and cannot provide a custom transform.

Although, in my usecase the textures from camera are mirrored, so `ImageOrientation` could be changed to be general and I can add mirroring as a flag

mortoray:
I think extended `Orientation` would also make sense here. General purpose transforms on ImageSource would have unintended consequences for layout/drawing/render calculations.

kusma:
@nekronos: `Source.Orientation` can already express all flipping and 90-degress rotation variations. You'll have to combine vertical flipping and 180 degree rotation to do horizontal flipping (or mirroring as you call it).

mortoray:
@nekronos Are you going to change the approach here to use the existing orientation as Erik describes?  Can this PR be closed?

mortoray:
I'll assume the lack of activity means this has been merged elsewhere, or a new approach has been used. I'll close this.

PR #919, opened 2018-01-08
ScrollView: fix native scrolling bugs

nekronos:
This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Native/Interfaces.uno
+++ b/Source/Fuse.Controls.Native/Interfaces.uno
@@ -123,6 +123,7 @@ namespace Fuse.Controls.Native
 	public interface IScrollViewHost
 	{
 		float PixelsPerPoint { get; }
+		float2 ContentSize { get; }
 		void OnScrollPositionChanged(float2 newScrollPosition);
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+### ScrollView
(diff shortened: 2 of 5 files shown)


Source/Fuse.Controls.Native/Interfaces.uno
```diff
@@ -123,6 +123,7 @@ namespace Fuse.Controls.Native
 	public interface IScrollViewHost
 	{
 		float PixelsPerPoint { get; }
+		float2 ContentSize { get; }
```
mortoray: This needs a CHANGLEOG entry... though I suspect this interface should actually be internal.

nekronos: I'll add that. But yeah, everything in that file should have been `internal`

mortoray: Or chnage them to internal and note that in the CHANGELOG.

kusma: Yeah, I think IScrollViewHost should be made internal.

nekronos: `IScrollViewHost` is used in public `UXConstructor`'s. Cannot easily be made internal

nekronos: I'll do a pass on `Fuse.Control.Native` and make whats need to be internal internal and make a separate PR

nekronos:
Fixes https://github.com/fusetools/fuselibs-public/issues/917

PR #926, opened 2018-01-09
Remove FuseJS/DI

### LINKED ISSUE #925: Deprecate FuseJS/DI
@bondehagen and @sebbert and I agree this isn't worth having. It may confuse users, and it also can't work correctly due to having no defined order to the execution of JavaScript.  Users can instead create their own common module and use exports from it instead.

Sebastian Reinhard 11:29
@mortoray I don't know, I probably would not depend on it
edA-qa mort-ora-y 11:30
That’s what I thought, in which case FuseJS/DI is not  usable. :disappointed:
11:33
Maybe if we use it always via a module that does the init it would work. `var DI = require("MyDI")`
Anders Bondehagen 11:35
I say we remove it
Sebastian Reinhard 11:35
I agree
11:36
I've only seen it lead to confusion among users
edA-qa mort-ora-y 11:37
Yes. Remove it. An easy alternative (Which works) is this:
11:37

```var DI = require("MyDI")
class Mymodel {
   constructor( service = DI.service ) { ... }
```

11:37
Then in MyID:
11:37

```exports.service = new SomeService()
```

11:38
It’s simple and serves the same purpose I think.
11:38
Doesn’t have any startup order concerns either.
11:40
_Though I should add that i would personally never use this approach since I don’t actually find value in constructor-based dependency injection._
11:41
_I would call DI.service() as needed, not added to constructor._

sebbert:
This removes the FuseJS/DI module - key reasons being:
- It is actually a service locator, not a dependency injector
- It is broken wrt. initialization order
- It causes unnecessary confusion among users who think it's an essential part of FuseJS
- Its functionality can be easily mirrored using standard CommonJS modules

Fixes #925 

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


sebbert:
@mortoray Hmm, maybe we should soft-deprecate it instead? Just give a warning when trying to `require()` the module?

mortoray:
Yes, I think a deprecation warning is required instead. Someobdy could be using this.
There should be a module for emitting deprecated messages from JS (I used it in OBservable)

kusma:
Hm. 1.6 has not yet gone out. Perhaps we could just sneak this in? If so, the only users are from the beta, and people should know better than to depend on beta-features...

mortoray:
Ah, yes, if this was not released yet then let's just kill it!
Low risk of breaking something since it wasn't used for anything we did. :)

sebbert:
There are probably a few example projects that will need changing as well. I'll add a ticket for myself.

kusma:
@sebbert: If the plan is to remove it from the release, this should be against release-1.6, not master.

mortoray:
@kusma Do you still manage merges to release branches?

kusma:
@mortoray: No, that'd be @nekronos now.

PR #928, opened 2018-01-10
Model: Bring back model expressions

sebbert:
This adds back the model expression system that was removed to reduce the scope of the Model feature.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


sebbert:
@mortoray No, there were only UX and JS changes to that test.

sebbert:
After discussing this in person during the daily, we arrived at a different solution that is much less error-prone, and doesn't rely on doing ugly hacks with the UX Expression system. I will open a new PR when it is ready.

PR #929, opened 2018-01-11
Update uno.stuff

ollef:
Update Uno.stuff to the latest Uno master so I can later update the tool's stuff files and preemptively fix some build errors seen when integrating masters.


kusma:
@ollef: Please don't push directly to the fuselibs-public repo. We push to personal forks, and send PRs from there.

PR #938, opened 2018-01-15
Negate fix

yupferris:
The `UXConstructor` attribute is necessary here for the UX compiler to
understand that the `Operand` property value should be passed as a ctor
arg, not set later.

This PR contains:
- [x] Changelog
- [x] Tests


Source/Fuse.Reactive.Expressions/Tests/UX/UnaryOperator.Basic.ux
```diff
@@ -1,7 +1,10 @@
...
 	<FuseTest.Data Value=" 'hi' " ux:Name="v"/>
+	<FuseTest.Data FloatValue="10" ux:Name="someFloat"/>
 	
 	<FuseTest.DudElement StringValue="{= _unJoin('ab') }" ux:Name="a"/>
 	<FuseTest.DudElement StringValue="{= _unJoin({none}) ?? 'x' }" ux:Name="b"/>
 	<FuseTest.DudElement StringValue="{= _unJoin({v}) }" ux:Name="c"/>
-</Panel>
\ No newline at end of file
+
+	<FuseTest.DudElement StringValue="{= -{someFloat} }" ux:Name="d"/>
```
mortoray: Can you use just `Value="-{someFloat}"`, or as an additional test.  To ensure it works without the escape sequence as well.

CHANGELOG.md
```diff
@@ -1,7 +1,8 @@
 # Unreleased
 
 ## Expressions
-- Added support for boolean `==` and `!=` expressions, which can be used for things like negating expressions.
+- Added support for boolean `==` and `!=` expressions, which can be used for things like negating boolean expressions.
```
mortoray: Can we get a simple test for these as well?  I fear regression on expressions a lot.

kusma: @mortoray: This is just a clarification of the changelog-message. There's already tests for these, as the feature landed earlier.

Source/Fuse.Reactive.Expressions/Tests/UX/UnaryOperator.Basic.ux
kusma: Seems like a good idea.

yupferris: I was already rewriting the tests as this was written; see new code, it better matches existing math tests and is written like this.

CHANGELOG.md
yupferris: yep, that was part of another change; it showed up here as I added the word `boolean` for clarification after seeing that and the new entry side-by-side.

PR #940, opened 2018-01-16
Fuse.Reactive.Expressions: implement LogicalNot

### LINKED ISSUE #915: Feature discussion: Adding a negation reactive expression
If I'm correct, there's currently no direct way to negate a boolean value in UX Expressions. I'm suggesting an expression, `not ( bool ) : bool` that negates a given boolean value.

One of the simpler use cases I've found is probably best described through code;

```XML
<Rectangle ux:Class="Toggle" Width="50" Height="50" Background="Blue" State="False">
  <bool ux:Property="State" />
  <Tapped>
    <Set Target="this.State" Value="{not({Property State})}" />
   </Tapped>
</Panel>
```

_Note: This might not be the best use case scenario since there is an actual Toggle class, but the point still stands._

I have working code that I could push, given that the feature is wanted.

_Note: I've chosen to implement the functionality as a function rather than an arithmetic expression (such as `!`) as I find it more fit for the readable nature of UX markup._

kusma:
Being able to use the logical not operator seems neat, and we already
have support in the parser. Seems we're just missing the actual
implementation, so let's add that.

Fixes #915.

This PR contains:
- [x] Changelog
- [x] Tests

### DIFF
--- a/Source/Fuse.Reactive.Expressions/Tests/UX/UnaryOperator.Basic.ux
+++ b/Source/Fuse.Reactive.Expressions/Tests/UX/UnaryOperator.Basic.ux
@@ -1,7 +1,14 @@
 <Panel ux:Class="UX.UnaryOperator.Basic">
 	<FuseTest.Data Value=" 'hi' " ux:Name="v"/>
+	<FuseTest.Data StringValue="{= true }" ux:Name="truth"/>
+	<FuseTest.Data StringValue="{= false }" ux:Name="falsehood"/>
 	
 	<FuseTest.DudElement StringValue="{= _unJoin('ab') }" ux:Name="a"/>
 	<FuseTest.DudElement StringValue="{= _unJoin({none}) ?? 'x' }" ux:Name="b"/>
 	<FuseTest.DudElement StringValue="{= _unJoin({v}) }" ux:Name="c"/>
+
+	<FuseTest.DudElement BoolValue="{truth}" ux:Name="d"/>
+	<FuseTest.DudElement BoolValue="!{truth}" ux:Name="e"/>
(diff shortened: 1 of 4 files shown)


Source/Fuse.Reactive.Expressions/Tests/UX/UnaryOperator.Basic.ux
```diff
@@ -1,7 +1,14 @@
 <Panel ux:Class="UX.UnaryOperator.Basic">
 	<FuseTest.Data Value=" 'hi' " ux:Name="v"/>
+	<FuseTest.Data StringValue="{= true }" ux:Name="truth"/>
+	<FuseTest.Data StringValue="{= false }" ux:Name="falsehood"/>
 	
 	<FuseTest.DudElement StringValue="{= _unJoin('ab') }" ux:Name="a"/>
 	<FuseTest.DudElement StringValue="{= _unJoin({none}) ?? 'x' }" ux:Name="b"/>
 	<FuseTest.DudElement StringValue="{= _unJoin({v}) }" ux:Name="c"/>
+
+	<FuseTest.DudElement StringValue="{= {truth} }" ux:Name="d"/>
```
mortoray: DudElement also has `BoolValue` to test directly, avoiding the string conversion.

kusma: @mortoray: Ah, thanks for the pointer!

kusma: Fixed, much appreciated :)

PR #941, opened 2018-01-16
Android: fix ImageView transform issue

nekronos:
The native imageview was transformed incorrectly, making native images
in certain cases be clipped and not match how it would look in a
GraphicsView

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+### Image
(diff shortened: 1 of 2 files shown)


nekronos:
Wups, found another case!

mortoray:
It'd als be good if a test page for all these alignment issues was created in the native test app.

PR #943, opened 2018-01-16
Less draw statements

kusma:
Here's some patches to centralize the draw-statements we have, with the goal of making it easier to replace the draw-statements with something else down the road. I have some patches for the latter, but they aren't ready yet.

This PR contains:
- [x] Changelog


kusma:
@yupferris: I think that's a good idea to consider. I'm not going to do anything in that direction right now, as I have a lot more to clean up in this area as it is, bit it sounds like a neat little project to do between other big things.

There's some other cleanups I'm working on that's going to also make this even easier :)

Issue #936, opened 2018-01-12
LinearGradient in Path shows discrete colors instead of a gradient


knatten:
The following:

```xml
<App>
    <StackPanel>
        <Path Data="M 0 0 L 1 0 L 1 1 L 0.262521114864865 1 L 0 1 L 0 0" StretchMode="Fill" Height="100">
            <LinearGradient StartPoint="0, 1" EndPoint="1">
                <GradientStop Color="#f00" Offset="0" />
                <GradientStop Color="#00f" Offset="1" />
            </LinearGradient>
        </Path>
        <Rectangle Height="100">
            <LinearGradient StartPoint="0, 1" EndPoint="1">
                <GradientStop Color="#f00" Offset="0" />
                <GradientStop Color="#00f" Offset="1" />
            </LinearGradient>
        </Rectangle>
    </StackPanel>
</App>
```

Produces this:
![screen shot 2018-01-12 at 16 02 08](https://user-images.githubusercontent.com/727543/34882046-91e8f080-f7b5-11e7-8140-a80b4e2ca4ca.png)

Tested on macOS local preview with installed Fuse, build 15133.

mortoray:
This works on Target=Native, but not in preview (DotNet). We've had other issues with DotNet not doing gradients correctly. 

https://github.com/fusetools/fuselibs-public/issues/841

mortoray:
<img width="787" alt="screen shot 2018-01-15 at 10 36 12" src="https://user-images.githubusercontent.com/1308999/34935962-f205e5dc-f9df-11e7-9aa1-ecf222100711.png">

It works fine for more complex layouts --- stop trying to do simple things! :P

kusma:
To be fair, I've had a look at the code, and it's completely bonkers. I can't really imagine it working under any circumstances besides some synthetic test-cases...

kusma:
In particular, that CalcGradientPoints-method. Stuff like `if (Math.Round(finalBoundDiff) == 1)` just screams nasty work-around for buggy code to me.

mortoray:
I recall Noah had a lot of issues with weird corner cases in DotNet.  It also had a strange notion of what the gradient end points were, differing from the other platforms. It's unfortunate the code doesn't document what output is actually expected from this function. :(

Issue #841, opened 2017-12-05
LinearGradient direction on Path is incorrect


drslem:
Consider this example where the only difference between the 3 paths is a _tiny_ variation of the X coordinate of the `LinearGradient.StartPoint`

Replacing the Path's with Rectangles gives the expected results (e.g. virtually no difference between the 3 instances)

```
<App>
	<Grid RowCount="3">
		<Path Data="M 1 1 L 0 1 L 0 0 L 0 0" StretchMode="Fill" Width="100%" Height="100%" >
			<LinearGradient StartPoint="0.5, 0" EndPoint="0.5, 1">
				<GradientStop Color="#f00" Offset="0" />
				<GradientStop Color="#00f" Offset="1" />
			</LinearGradient>
		</Path>

		<Path Data="M 1 1 L 0 1 L 0 0 L 0 0" StretchMode="Fill" Width="100%" Height="100%" >
			<LinearGradient StartPoint="0.50001, 0" EndPoint="0.5, 1">
				<GradientStop Color="#f00" Offset="0" />
				<GradientStop Color="#00f" Offset="1" />
			</LinearGradient>
		</Path>

		<Path Data="M 1 1 L 0 1 L 0 0 L 0 0" StretchMode="Fill" Width="100%" Height="100%" >
			<LinearGradient StartPoint="0.501, 0" EndPoint="0.5, 1">
				<GradientStop Color="#f00" Offset="0" />
				<GradientStop Color="#00f" Offset="1" />
			</LinearGradient>
		</Path>
	</Grid>
</App>
```

![image](https://user-images.githubusercontent.com/5551628/33615292-0cc122f4-d9da-11e7-897e-de60d1a92f29.png)

kusma:
@drslem: What back-end? I believe this has separate implementations per platform...

kusma:
OK. Looking at CalcGradientPoints in DotNetSurface.uno, I'm not very surprised. This code can't be correct...

drslem:
It should just use the minimum bounding box, right?

mortoray:
Yeah, the code looks a bit weird, but I recall we were fighting several compatibility problems on dotNet. It's also a bit tricky to get hese points right, as they're expressed in Element bounds, but may be outside or inside, but the actual drawing may not share the same bounds (see VectorLayer, or simply have an object exceeding the bounds).

mortoray:
But yeah, that tiny variation in X should make no visual difference.

PR #947, opened 2018-01-17
fixes the linearGradient on dotnet

mortoray:
This needs to be tested on Windows: I tested only DotNet on OSX.  I'm including  reference photo (CoreGraphics output, which is the same as the DotNet output on OSX).

_Note: The fix isn't complete per-se, you can still create gradients that fail by making the gradient line not cross the fill region. It's not a normal use-case, and would take more effort to fix than I think is desired now._

Test project: https://github.com/mortoray/fork-sandbox/tree/master/Jan2018/LinearGrad

screen shot 2018-01-17 at 11 36 10

fixes https://github.com/fusetools/fuselibs-public/issues/936
fixes https://github.com/fusetools/fuselibs-public/issues/841

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [ ] ~Tests~

### DIFF
--- a/Source/Fuse.Drawing.Surface/DotNetSurface.uno
+++ b/Source/Fuse.Drawing.Surface/DotNetSurface.uno
@@ -567,95 +567,14 @@ namespace Fuse.Drawing
 			return rightMin + (valueScaled * rightSpan);
 		}
 
-		/* identify the start and end of the gradient, along with the start of bounds and end of bounds
+		/**
+			A Clamped WrapMode for ColorBlend is not supported (despite being in the docs). We workaround by duplicating the start/end colors and extending the start/end points to cover the entire drawing region.
+			
+			This is done quite roughly now and assume the input is reasonably sane (not some/both points for outside of the rectangle, or of tiny distance).  It relies on DotNet donig the trimming of excessive start/end lines on its own (the calc here will almost always produce a line that is excessively long).
 		*/
-		static float4 CalcGradientPoints(RectangleF bounds, float startX, float startY, float endX, float endY) 
+		static ColorBlend CreateColorBlend(LinearGradient lg, RectangleF bounds, float2 inStart, float2 inEnd,
+			out float2 gStart, out float2 gEnd)
 		{
-			var finalBoundX = bounds.X + bounds.Width;
-			var finalBoundY = bounds.Y + bounds.Height;
-
-			var xDiff = endX - startX;
-			var yDiff = endY - startY;
-
-			// this is the "best case" gradient of the line
-			// in the cases where the diff is not best case, we adjust it
-			// accordingly 
-			var finalBoundDiff = xDiff / yDiff;
-
-			// this is the point the gradient _should_ start at
-			// Sometimes it may not if the gradient only goes in one direction, or equally in two directions
-			var startBoundPoint = - ( (finalBoundDiff * bounds.X) - bounds.Y  );
-			var endBoundPoint = - ( (finalBoundDiff * (bounds.X + bounds.Width)) - (bounds.Y + bounds.Height));
-
-
-			// when our gradient only changes in the y
-			if (Math.Round(xDiff) == 0)
-			{
-				finalBoundDiff = 0;
-				startBoundPoint = bounds.Y;
-				endBoundPoint = bounds.Y + bounds.Height;
-			} 
-			// when our gradient only changes in the x
-			else if (Math.Round(yDiff) == 0)
-			{
-				finalBoundDiff = 0;
-				startBoundPoint = bounds.X;
-				endBoundPoint = bounds.X + bounds.Width;
-			}
-			// when our gradient goes equally in X and Y
-			else if (Math.Round(finalBoundDiff) == 1) 
-			{
-				finalBoundDiff = 0;
-
-				// this should not matter, but accounts for rounding errors that may occur
-				if (yDiff <= xDiff)
-				{
-					startBoundPoint = startY;
-					endBoundPoint = endY;
-				}
-				else 
-				{
-					startBoundPoint = startX;
-					startBoundPoint = endX;
-				}
-			}
-			else 
-			{
-				// catch either bound point being 0 - default to X
-				if (startBoundPoint == 0)
-				{
-					startBoundPoint = startX;
-				}
-
-				if (endBoundPoint == 0)
-				{
-					endBoundPoint = endX;
-				}	
-			}
-
-
-			var gradientStartPoint = - ( (finalBoundDiff * startX) - startY);
-			var gradientEndPoint = - ( (finalBoundDiff * endX) - endY);
-
-			// when we failed to calculate the right gradient start point
-			if (gradientStartPoint == 0)
-			{
-				gradientStartPoint = xDiff == 0 ? startY : startX;
-			} 
-
-			if (gradientEndPoint == 0)
-			{
-				gradientEndPoint = xDiff == 0 ? endY : endX;
-			}
-
-			return float4(gradientStartPoint, gradientEndPoint, startBoundPoint, endBoundPoint);
-		}
-
-		static ColorBlend CreateColorBlend(LinearGradient lg, RectangleF bounds, float startX, float startY, float endX, float endY)
-		{
-
-			var calculatedPoints = CalcGradientPoints(bounds, startX, startY, endX, endY);
-
 			var stops = lg.SortedStops;
 
 			// we have 2 extra points, one at each end. This enables clamping
@@ -674,6 +593,22 @@ namespace Fuse.Drawing
 			colors[numberOfColorPoints - 1] = endColor;
 			offsets[numberOfColorPoints - 1] = 1.0f;
 
+			var inVec = Vector.Normalize(inEnd - inStart);
+			var inLen = Vector.Length(inEnd - inStart);
+			//safety
+			if (inLen < 0.1f) {
+				inVec = inEnd - inStart; //this affects only gStart/gEnd so okay to be incorrect in nonsense case
+				inLen = 1;
+			}
+
+			var minLen = Vector.Length( float2(bounds.Width, bounds.Height) );
+			//extra length needed to extend line to cover whole region
+			var extraLen = Math.Max(minLen - inLen,0);
+			var exOff = extraLen / inLen;
+			var exLen = inLen * exOff;
+			//the 0/1 input offsets as offsets in the adjusted line
+			var inOffStart = exOff / (1 + 2 * exOff);
+			var inOffEnd = (1+exOff) / (1 + 2 * exOff);
 
 			for (int i=0; i < stops.Length; ++i)
 			{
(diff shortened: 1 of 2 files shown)


kusma:
Looks reasonable to me on Windows:
image

Source/Fuse.Drawing.Surface/DotNetSurface.uno
```diff
@@ -674,6 +593,22 @@ namespace Fuse.Drawing
 			colors[numberOfColorPoints - 1] = endColor;
 			offsets[numberOfColorPoints - 1] = 1.0f;
 
+			var inVec = Vector.Normalize(inEnd - inStart);
+			var inLen = Vector.Length(inEnd - inStart);
+			//safety
+			if (inLen < 0.1f) {
```
kusma: I think this is a bit large of an epsilon here. Shouldn't the code work even if this is way smaller (like 1e-5 or something)?

mortoray: The code might start misbehaving if this gets too small as it alters the offset ranges too much.  0.1 is chosen as kind of a cut-off where anything below this you can't visually see a difference anyway (this is a measure of pixel-spread of the gradient).

kusma:
I now get this when running manualtestingapp:
```
Error: System.Exception: Unexpected exception ---> System.OutOfMemoryException: Out of memory.
   at System.Drawing.Drawing2D.LinearGradientBrush..ctor(PointF point1, PointF point2, Color color1, Color color2)
   at Fuse.Drawing.DotNetHelpers.StrokePathLinearGradient(Graphics graphics, GraphicsPath path, LinearGradient lg, Single startX, Single startY, Single endX, Single endY, Single width, Single miterLimit, LineJoin lineJoin, LineCap lineCap) in C:\Users\erik\src\fuselibs\Source\Fuse.Drawing.Surface\DotNetSurface.uno:line 719
   at Fuse.Drawing.DotNetSurface.StrokePath(SurfacePath path, Stroke stroke) in C:\Users\erik\src\fuselibs\Source\Fuse.Drawing.Surface\DotNetSurface.uno:line 259
   at Fuse.Controls.Shape.Fuse.Drawing.ISurfaceDrawable.Draw(Surface surface) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Primitives\Shapes\Shape.Surface.uno:line 105
   at Fuse.Drawing.Surface.DrawLocal(ISurfaceDrawable drawable) in C:\Users\erik\src\fuselibs\Source\Fuse.Drawing.Surface\Surface.uno:line 150
   at Fuse.Drawing.Surface.Draw(DrawContext dc, Element elm, ISurfaceDrawable drawable) in C:\Users\erik\src\fuselibs\Source\Fuse.Drawing.Surface\Surface.uno:line 135
   at Fuse.Controls.Shape.DrawVisual(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Primitives\Shapes\Shape.Visual.uno:line 22
   at Fuse.Controls.Control.OnDraw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls\Control.uno:line 43
   at Fuse.Elements.Element.DrawWithChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 113
   at Fuse.Elements.Element.CompositEffects(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 184
   at Fuse.Elements.Cache.DrawHeuristically(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Caching\Cache.uno:line 83
   at Fuse.Elements.Element.Composit(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 103
   at Fuse.Elements.Element.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Drawing.uno:line 109
   at Fuse.Elements.Element.DrawNonUnderlayChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 71
   at Fuse.Elements.Element.DrawWithChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 113
   at Fuse.Elements.Element.CompositEffects(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 184
   at Fuse.Elements.Cache.DrawHeuristically(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Caching\Cache.uno:line 83
   at Fuse.Elements.Element.Composit(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 103
   at Fuse.Elements.Element.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Drawing.uno:line 109
   at Fuse.Controls.Panel.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Panels\Panel.Freeze.uno:line 208
   at Fuse.Elements.Element.DrawNonUnderlayChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 71
   at Fuse.Elements.Element.DrawWithChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 113
   at Fuse.Elements.Element.CompositEffects(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 184
   at Fuse.Elements.Cache.DrawHeuristically(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Caching\Cache.uno:line 83
   at Fuse.Elements.Element.Composit(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 103
   at Fuse.Elements.Element.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Drawing.uno:line 109
   at Fuse.Controls.Panel.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Panels\Panel.Freeze.uno:line 208
   at Fuse.Elements.Element.DrawNonUnderlayChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 71
   at Fuse.Elements.Element.DrawWithChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 113
   at Fuse.Elements.Element.CompositEffects(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 184
   at Fuse.Elements.Cache.DrawHeuristically(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Caching\Cache.uno:line 83
   at Fuse.Elements.Element.Composit(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 103
   at Fuse.Elements.Element.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Drawing.uno:line 109
   at Fuse.Controls.Panel.Draw(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Controls.Panels\Panel.Freeze.uno:line 208
   at Fuse.Elements.Element.DrawNonUnderlayChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 71
   at Fuse.Elements.Element.DrawWithChildren(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Batching.uno:line 113
   at Fuse.Elements.Element.CompositEffects(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Element.Effects.uno:line 184
   at Fuse.Elements.ElementAtlas.FillFramebuffer(DrawContext dc, Framebuffer fb, Boolean drawAll, Rect scissorRectInClipSpace) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Caching\ElementAtlas.uno:line 212
   at Fuse.Elements.ElementAtlas.PinAndValidateFramebuffer(DrawContext dc) in C:\Users\erik\src\fuselibs\Source\Fuse.Elements\Caching\ElementAtlas.uno:line 149
```

mortoray:
Ah man, out-of-memory inside DotNet libraries isn't fun.  I'll see if it happens on OSX/DotNet.  It probably relates to the length of the gradient line -- if you can emit the gStart/gEnd prior to line 719 it might indicate the issue.

kusma:
gStart: 0, 0
gEnd: 0, 0

:/

kusma:
This confuses me... It seems like we should have exited earlier than this due to `if (bounds.IsEmpty)` at line 709...

kusma:
Ah, the problem is that inVec gets zero-length.

kusma:
So, StrokePathLinearGradient() gets called with startX=0, startY=0, endX=0, endY=0. That can't be very well-defined, can it?

mortoray:
Completely zero length is weird for `inVec`... my epsilon check would catch low values, but I guess zero values still result in nonsense.

I can't reproduce this on DotNet/OSX, so Mono appears to check against this case.

kusma:
The question is, what do we expect to happen here?

mortoray:
No, if Start == End the gradient is bascially undefined. We could fix it however we want, I suggest setting start to 0,0 and End to 1,1.

This may happen if an element is currently a zero-sized (happens during scaling, but that'd be unusual for the test app)

mortoray:
Sampling the middle of the gradient would be a lot more work in the surface implementation, since it's already in the gradient branch.

kusma:
Sure, it just seems like a more well-behaved fallback than 0,0 <-> 1,1. That angle seems pretty arbitrary to me.

kusma:
By the way `bounds: {X=-7,Y=-7,Width=14,Height=14}`, so it's not an invisible element. It's the gradient-points who are problematic, not the size of the element itself.

mortoray:
The angle is arbitrary, but the default on DotNet, thus I assume it'll work on all sizes.  Plus, given the small size it'll likely sample the middle color, or the end color or any color, but since we say it's undefined we don't care which one.

mortoray:
Aha, so the gradient points are too close together in that case (After my translation). I can probably improve my fix so that doesn't happen.

kusma:
This cheesy fix gets things working for me again:
```patch
diff --git a/Source/Fuse.Drawing.Surface/DotNetSurface.uno b/Source/Fuse.Drawing.Surface/DotNetSurface.uno
index caba0a92fe..fefde896dc 100644
--- a/Source/Fuse.Drawing.Surface/DotNetSurface.uno
+++ b/Source/Fuse.Drawing.Surface/DotNetSurface.uno
@@ -597,7 +597,7 @@ namespace Fuse.Drawing
 			var inLen = Vector.Length(inEnd - inStart);
 			//safety
 			if (inLen < 0.1f) {
-				inVec = inEnd - inStart; //this affects only gStart/gEnd so okay to be incorrect in nonsense case
+				inVec = float2(1); //this affects only gStart/gEnd so okay to be incorrect in nonsense case
 				inLen = 1;
 			}
 
```
@mortoray: If you don't mind, I'd like to push this to master to un-break stuff. Is that OK?

mortoray:
I'm fine with that fix. It's just an unusual corner-case anyway, and I said my code doens't handle them all correctly (like apparently DotNet doesn't either ;)

mortoray:
If it works, and my test app renders the same, I'd say it's a suitable long-term fix -- until we find a legitimate use of a corner-case that produces the wrong gradient.

kusma:
@mortoray: It looks correct as far as I can tell. I just went ahead and pushed out the fix, since things don't currently even start.

PR #950, opened 2018-01-18
Make stuff/uno pack use same versioning scheme

knatten:
This should change the versioning of artifacts built with `stuff pack`
to be the same as for artifacts built with `uno pack`. In particular,
`AutomaticTestApp.stuff` and `ProjectTemplates.stuff` will now get the
same version numbers as `fuselibs.packages`. This will make things less
confusing for consumers of these packages.

#### Locally
I've tested this locally, here are examples from applying the patch to master, a release branch and another branch:

##### master
Before:
```
Creating upload/Polyfills.Window-1.6.0-rc1-master-070b904.upk  0.26 s
Creating ProjectTemplatesUpload/ProjectTemplates-1.6.0-rc1-local.zip
```
After:
```
Creating upload/Polyfills.Window-1.6.0-rc1-master-bc5baca.upk  0.24 s
Creating ProjectTemplatesUpload/ProjectTemplates-1.6.0-rc1-master-bc5baca.zip
```

##### release branch
Before:
```
Creating upload/Polyfills.Window-1.6.0-rc1.upk  0.24 s
Creating ProjectTemplatesUpload/ProjectTemplates-1.6.0-rc1-local.zip
```
After:
```
Creating upload/Polyfills.Window-1.6.0-rc1.upk  0.23 s
Creating ProjectTemplatesUpload/ProjectTemplates-1.6.0-rc1.zip
```

##### feature branch
Before:
```
Creating upload/Polyfills.Window-1.6.0-rc1-dev-070b904.upk  0.24 s
Creating ProjectTemplatesUpload/ProjectTemplates-1.6.0-rc1-local.zip
```
After:
```
Creating upload/Polyfills.Window-1.6.0-rc1-dev-793254c.upk  0.23 s
Creating ProjectTemplatesUpload/ProjectTemplates-1.6.0-rc1-dev-793254c.zip
```

#### In CI

Obviously, I'm not able to test the behaviour of my PR in branches with these names in our CI servers, which is where it really matters. But I've checked which types of version numbers our CI server produces for each type of branch:

```
master: -1.6.0-rc1-master-c8bc166
release: -1.6.0-rc1
other: -1.4.0-rc1-dev-7d6fc20
```

So the new versioning scheme should work both locally and in CI.

### DIFF
--- a/pack.sh
+++ b/pack.sh
@@ -5,12 +5,7 @@ cd "`dirname "$SELF"`" || exit 1
 PATH="Stuff:$PATH"
 OUT="upload"
 
-# Detect version
-if [ -n "$RELEASE_VERSION" ]; then
-    VERSION="$RELEASE_VERSION"
-else
-    VERSION=$(cat VERSION.txt)"-local"
-fi
+VERSION=$(cat VERSION.txt)
 
 # Detect revision
 if [ -n "$BUILD_VCS_NUMBER" ]; then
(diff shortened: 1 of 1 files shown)


kusma:
I'm not a huge fan. The "-local" label is useful, because it tells us that the build can also have *local changes*. This cannot happen in CI for obvious reasons.

knatten:
Sure, but the `-local` label was only used for `AutomaticTestApp.stuff` and `ProjectTemplates.stuff`, not for `fuselibs.packages`.

I think it's more important that the artifacts use the same versioning scheme than that some them has a somewhat useful feature in a corner case, which the main artifact lacks.

If we want to use that label, it should be used for both `fuselibs.packages` too.

kusma:
Yes, I agree. But this *removes* the local-label rather than adding it.

However, thinking about this a bit more, maybe that part is OK; we don't really run this script locally, and you need API keys to upload the result. So it's not like locally built packages have a very high chance of spreading.

TL;DR you've got me convinced ;)

Issue #966, opened 2018-01-24
Text doesn't render TextColor opacity correctly (Preview/iOS)


mortoray:
The opacity of the TextColor does not apply to emoji on Desktop preview and iOS (Native and non-native).

On Android is does apply, and logiclaly it should (even if it can't apply the color).  That is, this rendering (with emoji opacity).  This is what Android shows, as well as the correct desktop preview.
<img width="194" alt="screen shot 2018-01-24 at 10 23 32" src="https://user-images.githubusercontent.com/1308999/35324321-b1895aa8-00f0-11e8-9fff-e50d5858690e.png">

Seems more correct than this one (without emoji opacity). This is what iOS currently shows.

<img width="194" alt="screen shot 2018-01-24 at 10 24 25" src="https://user-images.githubusercontent.com/1308999/35324357-c773eb58-00f0-11e8-9746-81d4334b427b.png">

kusma:
I actually think Android is the one in the wrong here, based on how we've previously defined this to work.

Issue #918, opened 2018-01-08
PlaceholderColor does not blend alpha right on local targets


eksperts:
Fuse 1.6.0 build 15070, macOS High Sierra.

Repro:
```
<App Background="Gray">
    <Panel Alignment="Center">
        <StackPanel ItemSpacing="20">
            <TextInput PlaceholderText="This black placeholder text has alpha" PlaceholderColor="#0005"/>
            <TextInput PlaceholderText="While this white text does not" PlaceholderColor="#fff2"/>
        </StackPanel>
    </Panel>
</App>
```
On local preview, both the black and white placeholders are expected to handle the alpha channel so that they're partially transparent. While the black one gets faded about right, the white one doesn't.

Native and dotnet build targets affected as well.
Neither Android nor iOS targets are affected, they render exactly as expected.

fusebuild:
The forum thread [No alpha on placeholder text](https://www.fusetools.com/community/forums/bug_reports/no_alpha_on_placeholder_text) posted by Paul was linked to this issue.

danhollick:
Video demonstrating the issue

[Opacity issue.mov.zip](https://github.com/fusetools/fuselibs-public/files/1635008/Opacity.issue.mov.zip)

danhollick:
This seems to be a problem beyond `PlaceHolderColor. Happens with normal `<Text/>` elements in some cases, using both `Color` and `TextColor`.

knatten:
@mortoray you wanted to be notified about issues related to the Sketch importer, this seems to be one. (It seems to be the cause of #86)

mortoray:
That's Strange :tm: !?

mortoray:
@knatten That linked issue 86 appears to be the wrong one, from wrong repo I guess?

knatten:
Yeah, sorry. I meant https://github.com/fusetools/Sketch2Fuse/issues/86.

mortoray:
<img width="295" alt="screen shot 2018-01-24 at 09 07 19" src="https://user-images.githubusercontent.com/1308999/35320991-08219dcc-00e6-11e8-9856-8e81ede888f7.png">

Extended example showing how it affects all non-block colors. THe other colors appear to blend with a white, or bright color, not the background. 

```
<App Background="#AAA">
    <Panel Alignment="Center">
        <StackPanel ItemSpacing="5">
            <Text Value="While this white text does not" TextColor="#fff4"/>
[...]

Issue #967, opened 2018-01-24
Native Text on iOS doesn't support emoji opacity


mortoray:
I was able to fix https://github.com/fusetools/fuselibs-public/issues/966 for all targets except native iOS text. Their native renderers just don't support it. For graphics mode I used opacity on the blitting, but for the UITextView I have no simple option.  I thought about setting the view opacity, but that would affect more than intended (children and background for example).

Perhaps using an attribute RichText string (NSAttributeText or whatever it is called), might be an option.

kusma:
I think we would have to add a separate child *just* for the text, put it in it's own layer, and set the opacity for that.

PR #968, opened 2018-01-24
Fixes text opacity on desktop and iOS graphics mode

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/966
fixes https://github.com/fusetools/fuselibs-public/issues/918

Unfortunately https://github.com/fusetools/fuselibs-public/issues/967  remains open (iOS support for emoji transparency)

Android was working previously in all situations it appears.

#918 was introduced by Harfbuzz rendering -- so it's been around for a while now, but technically still a regression.

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [x] Tests


Source/Fuse.iOS.TextRenderer/TextRenderer.uno
```diff
@@ -47,7 +48,9 @@ namespace Fuse.iOS.Bindings
 
 			_font = Fuse.Controls.Native.iOS.FontCache.Get(control.Font.Descriptors[0], control.FontSize * control.Viewport.PixelsPerPoint);
 
-			_textColor = ToUIColor(control.TextColor);
+			//iOS Text rendering fails to apply opacity to emoji, therefore we apply opacity on our own
+			_textColor = ToUIColor(float4(control.TextColor.XYZ,1));
+			TextOpacity = control.TextColor.W;
```
kusma: I don't think this is correct. We've previously defined TextColor to only apply to non-colored glyphs (typically emoji, but can also apply to colorfonts in the future). That is, it can be thought of as a "default text color unless the font specifies otherwise", not as a multiplier on the font-data.

If the user wants to fade out all text, `Opacity` should be used instead.

Source/Fuse.Text/Renderer.uno
```diff
@@ -176,7 +176,8 @@ namespace Fuse.Text
 
 						ClipPosition: Vector.Transform(float4(Position, 0, 1), pixelToClipSpaceMatrix);
 						float4 sampleColor: sample(texture, TexCoord, SamplerState.LinearClamp);
-						PixelColor: grayScale ? color * float4(sampleColor.X) : sampleColor;
+						float4 outColor: grayScale ? color * float4(sampleColor.X) : float4(sampleColor.XYZ, color.W * sampleColor.W);
+						PixelColor: outColor * outColor.W;
```
kusma: OK, so it looks like there's two "problems" this tries to fix. One is that in the grayScale case `color` is non-premultiplied, which could be better fixed by just pre-multiplying `color` before the shader:
```patch
diff --git a/Source/Fuse.Text/Renderer.uno b/Source/Fuse.Text/Renderer.uno
index 9fd28b53e3..a0bbd21c50 100644
--- a/Source/Fuse.Text/Renderer.uno
+++ b/Source/Fuse.Text/Renderer.uno
@@ -139,6 +139,8 @@ namespace Fuse.Text
 
 		public void Draw(float4 color, float4x4 pixelToClipSpaceMatrix)
 		{
+			color = float4(color.XYZ * color.W, color.W);
+
 			lock (_glyphAtlasMutex)
 			{
 				var batches = GetBatches();
```
The other is that color isn't taken into account in the non-grayScale at all. As I said below, I think this is actually correct, at least in the current definition.

Source/Fuse.iOS.TextRenderer/TextRenderer.uno
mortoray: I don't see that we actively defined what should happen before, there's no notes on it and it behaves differently per-platform. On Android the colored glyphs fade with the provided text opacity.  Given Android is the dominant platform, and the behaviour seems reasonable, I think it's the correct definition.

I can't see any use-case why the user would want the rest of the text faded, but not the colored glyphs -- it leads to a jarring visual contrast.

If we instead adopt iOS's approach, we're left with Android, thus most devices, not rendering consistently.

mortoray: HTML/CSS also fades out the emoji opacity (tested on Firefox).

kusma: This has been defined, but probably not documented anywhere.

However, if HTML/CSS does it, I think that's a good argument for us to do it as well. But please fix these things as separate commits.

kusma: I'm talking separate commits, not separate PRs. It's just easier to read what happened if there's one concern at the time.

Source/Fuse.Text/Renderer.uno
kusma: If you do this change, then the second change would be reduced to:
```diff
diff --git a/Source/Fuse.Text/Renderer.uno b/Source/Fuse.Text/Renderer.uno
index 9fd28b53e3..2231ffd20a 100644
--- a/Source/Fuse.Text/Renderer.uno
+++ b/Source/Fuse.Text/Renderer.uno
@@ -176,7 +176,7 @@ namespace Fuse.Text
 
 						ClipPosition: Vector.Transform(float4(Position, 0, 1), pixelToClipSpaceMatrix);
 						float4 sampleColor: sample(texture, TexCoord, SamplerState.LinearClamp);
-						PixelColor: grayScale ? color * float4(sampleColor.X) : sampleColor;
+						PixelColor: grayScale ? color * float4(sampleColor.X) : color * sampleColor;
 					};
 				}
 			}
```

kusma: Also, I don't think the change you propose here is correct either. AFAIK, sampleColor is pre-multiplied, so it's alpha would end up being multiplied in twice in this case. It's probably not easy to see, because sampleColor.W mostly contains anti-aliasing details, but thin details should become even thinner than they should be with this approach.

mortoray: I've pushed the change. I went with a modified form of your first change. Note we only `sampleColor * color.W` since otherwise we'd be affecting the tinting/coloring as well, not just the opacity.

CHANGELOG/textcolor.opacity.md
```diff
@@ -0,0 +1,3 @@
+## TextColor Opacity
+- Fixed a failure to render translucent `TextColor` correctly
+- Fixed the rendering of opaque emoji when `TextColor` is translucent (they will not also be translucent, though still  use the font coloring)
```
kusma: I assume "they will not also be" was meant to be "they will *now* also be"?

CHANGELOG/textcolor.opacity.md
```diff
@@ -0,0 +1,3 @@
+## TextColor Opacity
+- Fixed a failure to render translucent `TextColor` correctly
+- Fixed the rendering of opaque emoji when `TextColor` is translucent (they will now be translucent, though still  use the font coloring)
```
kusma: Also, is there consensus to switch to the changelog-directory approach? I haven't heard anything about that...

mortoray: Not yet, as there was no daily. But I will collapse these into CHANGELOG.md at the time I merge if we don't agree on this.

kusma: Fair enough!

kusma:
@mortoray: I just pushed out a split-up version (the result is identical), so the history will be a bit easier to read.

mortoray:
@kusma We can merge it then?

kusma:
@mortoray: Yeah, this looks good to me.

PR #969, opened 2018-01-24
Model: Fix pushing an array onto an array

sebbert:
This fixes a severe bug in Model where we get a crash when pushing an array onto an array.

`Array.prototype.concat()` is recursive up to one level.
This means that if one of the arguments is an array, then its elements,
rather than the array itself, will be appended to the original array.
This changes it to always wrap the value in a single-element array,
which gives us the intended behavior.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Models/Tests/Model.Test.uno
+++ b/Source/Fuse.Models/Tests/Model.Test.uno
@@ -12,6 +12,20 @@ namespace Fuse.Models.Test
 	public class ModelTest : ModelTestBase
 	{
 		[Test]
+		public void NestedArray()
+		{
+			var e = new UX.Model.NestedArray();
+			using(var root = TestRootPanel.CreateWithChild(e))
+			{
+				root.StepFrameJS();
+				Assert.AreEqual("", GetRecursiveText(e));
+				e.push.Perform(); // Throws if test fails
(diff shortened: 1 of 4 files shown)


kusma:
@sebbert: How about a test to prove the bugfix?

sebbert:
@kusma Can do!

sebbert:
@kusma There ya go

kusma:
@sebbert: Could you also squash them into one commit? Otherwise, we'll get a temporarily failing test while bisecting...

Source/Fuse.Models/Tests/Model.Test.uno
```diff
@@ -11,6 +11,17 @@ namespace Fuse.Models.Test
...
 	public class ModelTest : ModelTestBase
 	{
+		[Test]
+		public void NestedArray()
+		{
+			var e = new UX.Model.NestedArray();
+			using(var root = TestRootPanel.CreateWithChild(e))
+			{
+				root.StepFrameJS();
+				e.push.Perform(); // Throws if test fails
```
kusma: Could we also validate that the code did what we expected here?

sebbert:
@kusma Done

Issue #955, opened 2018-01-19
<PageControl />: System.NullReferenceException for Active page with variables


ianychoi:
Fuse version: 1.5.0 (build 15046)

The following source generates `System.NullReferenceException` in Problems section in Fuse Studio.
When I use static Active parameter & page names (e.g., `Active="abc"` and `Page Name="abc"`), it works well. However, when I try to use those parameters as dynamic like the following source code, it looks working well however it generates `System.NullReferenceException` on preview mode (exporting to Android & iOS apps seem working well).

```
<App>
	<JavaScript>
		exports.activePage = "222";
		exports.pageData = [ { "pageName": "111", "color": "#FF0000" },
						{ "pageName": "222", "color": "#00FFFF" },
						{ "pageName": "333", "color": "#0000FF" } ];
	</JavaScript>

	<PageControl Active="{activePage}">
		<Each Items="{pageData}">
			<Page Name="{pageName}" Background="{color}">
				<Text Value="{pageName}" Alignment="Center"/>
			</Page>
		</Each>
	</PageControl>
</App>

```

Since I cannot copy & paste in Problems section, I took a screenshot and am attaching now:
![System.NullReferenceException](https://user-images.githubusercontent.com/6965982/35141464-56f589aa-fd3e-11e7-8675-347f4d1ac335.png)

mortoray:
`PageControl.Active` cannot be bound to a JS variable this way. You'd have to bind to a `Node` object, not a string. You should instead bind to `ActiveIndex`, which is an integer, and/or use the functions to navigate (or add a feature request for a `ActiveName` binding).

I'll leave this issue open as the resulting error is a bit ridiculous.  A nice error message would be expected instead.

mortoray:
We apparently do support the one-way binding of string->node, so this should have worked.

PR #971, opened 2018-01-25
Fix Pagecontrol null crash on binding by name

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/955

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [x] Tests


kusma:
Should this be based on release-1.6 instead? I'm not sure, just asking if that might be the case...

mortoray:
I didn't see any indication it's a regression or blocking issue that would require it on 1.6.  If there is one I can switch it.

kusma:
OK, fair enough.

PR #980, opened 2018-01-26
Fuse.Scripting: restore the old NativeEvent.RaiseAsync API

### LINKED ISSUE #979: Error on NativeEvent.RaiseAsync Method after upgrading to 1.6
- Fuse version: 1.6.0 (build 15273)

A Korea community member reported that some existing codes using `NativeEvent.RaiseAsync Method (object[])` did not work after upgrading Fuse to 1.6.0. The error message is:

`E3128: Call to 'Fuse.Scripting.NativeEvent.RaiseAsync(Fuse.Scripting.IThreadWorker,object[])' has some invalid arguments (string)`

I tested this method using some code: https://www.fusetools.com/community/forums/howto_discussions/push_from_uno_to_javascript and https://www.fusetools.com/docs/native-interop/native-js-modules#nativeevent . I am attaching my sample code: 
[NativeEventTest.zip](https://github.com/fusetools/fuselibs-public/files/1666733/NativeEventTest.zip)
 and also the full error message.

```
Configuring  2.25 s
Compiling syntax tree
/Users/ian/Fuse/NativeEventTest/Chat.uno(28.30): E3128: Call to 'Fuse.Scripting.NativeEvent.RaiseAsync(Fuse.Scripting.IThreadWorker,object[])' has some invalid arguments (string)
/Users/ian/Fuse/NativeEventTest/Chat.uno(28,31): Error E3128: Call to 'Fuse.Scripting.NativeEvent.RaiseAsync(Fuse.Scripting.IThreadWorker,object[])' has some invalid arguments (string)/Users/ian/Fuse/NativeEventTest/Chat.uno(30.30): E3128: Call to 'Fuse.Scripting.NativeEvent.RaiseAsync(Fuse.Scripting.IThreadWorker,object[])' has some invalid arguments (string)
/Users/ian/Fuse/NativeEventTest/Chat.uno(30,31): Error E3128: Call to 'Fuse.Scripting.NativeEvent.RaiseAsync(Fuse.Scripting.IThreadWorker,object[])' has some invalid arguments (string)1.00 s

Build completed in 3.28 seconds
    2 errors
```

kusma:
This was an unintended hard API-breakage, but luckily it's pretty
simple to fix.

We mark the API as obsolete, so people can forward-port their code
to be correct.

Fixes #979.

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [ ] ~Tests~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 
 # 1.6
 
+## 1.6.1
+
+### Fuse.Scripting
+- Restore accidentally broken NativeEvent.RaiseAsync() API.
+
 ## 1.6.0
 
 ### Image
--- a/Source/Fuse.Scripting/NativeEvent.uno
+++ b/Source/Fuse.Scripting/NativeEvent.uno
@@ -50,6 +50,15 @@ namespace Fuse.Scripting
 				threadWorker.Invoke(new CallDiscardingResultClosure(_jsFunction, _eventArgsQueue.Dequeue()).Run);
 		}
 
+		[Obsolete("Use `RaiseAsync(IThreadWorker, params object[])` instead")]
+		public void RaiseAsync(params object[] args)
+		{
+			if(Context != null || _queueEventsBeforeEvaluation)
+				_eventArgsQueue.Enqueue(args);
+
+			DispatchQueue(Context.ThreadWorker);
+		}
+
 		public void RaiseAsync(IThreadWorker threadWorker, params object[] args)
 		{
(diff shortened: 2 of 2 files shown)


kusma:
This should probably have changelog before merging.

kusma:
Argh, no. This doesn't work; now it's impossible to call the latter method, because `params object[] args` can (and gets) matched before.

I'm not sure this is great behavior of Uno, but it seems to be what happens... :(

PR #982, opened 2018-01-26
Model: Fix parent refs on array manipulation

### LINKED ISSUE #958: Model: Attempted to detach a parent that was not attached to us
Not linking the test case here as this came from a user, but we hit another case of detaching a node that is not our child.

sebbert:
Whenever an element is inserted or removed from an array,
the indices of later elements are also adjusted.
However, for every node in the model graph, we also keep track of which
nodes have a property that refers to this node. These parent references
were not adjusted accordingly.

Now, when we do an `insertAt` or `removeRange`, the parent references
of elements that appear after the index we insert/remove at will be
adjusted accordingly.

Fixes #958

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [x] Tests

### DIFF
--- a/Source/Fuse.Models/FuseJS/Internal/Model.js
+++ b/Source/Fuse.Models/FuseJS/Internal/Model.js
@@ -436,11 +436,29 @@ function Model(initialState, stateInitializer)
 			TreeObservable.set.apply(store, argPath);
 		}
 
+		function updateArrayParentIndices(start, offset) {
+			// Updates the indices of parent references after removing or inserting elements in the middle of an array
+			for(var i = start; i < node.length; ++i) {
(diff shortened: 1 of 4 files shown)


kusma:
@sebbert: should this be done against release-1.6 instead, perhaps?

sebbert:
Ah, that's a good idea, forgot that we of course do patch releases 😄

sebbert:
@kusma Done

kusma:
@sebbert: seems Appveyor is not loving it :/

kusma:
Ah, yeah.

mortoray:
@sebbert I see the comments and won't merge -- you can do it if you've resolved that issue.

PR #998, opened 2018-02-02
reintroduce `Invoke(Action ..)` to Fuse.Scripting.Context

cbaggers:

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 
 # 1.6
 
+## 1.6.2
```
kusma: 1.6.1 isn't shipped yet, no need to introduce 1.6.2 just for this...

PR #997, opened 2018-02-02
Upgrade to Mono 5.4.1

BeeWarloc:
Everything required for upgrading Mono from 4.4.2 to 5.4.1.

This PR contains:
- [ ] Changelog - not really feature related, do we still put this in the changelog?
- [x] Documentation - changed the mono version pointed to from README.md
- [x] Tests - covered by same tests as before

PR #1002, opened 2018-02-05
Upgrade to Uno master

knatten:
This might not work until #997 is merged, but let's make a PR so we can get it through CI at least.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
I believe this can *technically speaking* be merged without #997; we're still compatible without Mono 5, we just don't get to upgrade yet.

However, if we do so, we'll end up down-grading the stuff-files.

kusma:
I think we should ignore that part, and just update #997 to not have new stuff files, just the mono5-enablement. Then we can merge it after 1.7 and do one thing at the time.

PR #1009, opened 2018-02-06
NativeMember thread-cleanup

kusma:
This restores the same thread-safety that the change that got reverted in #993 introduced, but without a hard API breakage.

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [ ] ~Tests~


kusma:
This is going to be affected a bit by #1011, so I'll rebase and adjust this to not conflict once that's reached master. Don't merge yet :)

PR #1016, opened 2018-02-12
Feature/platform functions, cleanup SystemUI

mortoray:
Adds `Device` and `window()`

Cleans up a lot of the internals of SystemUI.

This PR contains:
- [x] Changelog
- [x] Documentation
- [ ] ~Tests~ Unclear, I might add another PR for a comprehensize ManualTestPage later.


mortoray:
Wait on this. Though compiling this no longer works on an actual iPhone5s (emulators work though).

mortoray:
Note that this does not work for dynamic looking as `objc_msgSend_stret` is not supported on ARM64:

```
UIEdgeInsets (*getEdge)(id receiver, SEL operation);
getEdge = (UIEdgeInsets(*)(id,SEL))objc_msgSend_stret;
UIEdgeInsets insets = getEdge(view, @selector(safeAreaInsets));
```

Travis doesn't have an SDK which recognizes `safeAreaInsets`, which was what the above tried to fix.

Source/Fuse.Controls.Native/iOS/KeyboardView.uno
```diff
@@ -75,8 +75,5 @@ namespace Fuse.Controls.Native.iOS
 			if (FocusHelpers.IsFirstResponder(_handle))
 				FocusHelpers.ResignFirstResponder(_handle);
 		}
-
```
yupferris: Not sure we need these whitespace removals

Source/Fuse.Platform/iOS/SystemUI.uno
```diff
@@ -161,74 +216,68 @@ namespace Fuse.Platform
...
-					handler(null, args);
-			}
+			if (MarginsChanged != null)
+				MarginsChanged();
 		}
 
 		static Rect GetStatusBarFrame()
 		{
-			var density = Uno.Platform.Displays.MainDisplay.Density;
+			var density = 1;//Uno.Platform.Displays.MainDisplay.Density;
```
yupferris: This seems bad. :)

mortoray: The density stuff was intentional as I changed the contract on SystemUI to return units in the device unit size. You  need access to a `Viewport` objecct to convert correctly, and that is done inside `window()` now. There are no other users of SystemUI.  Plus, most of the old interfaces to it were removed.

I don't think it should cause any ill-effects.

yupferris: That sounds good; it looked a bit suspect due to the comments though.

mortoray: Yeah, I'll cleanup the comments.

mortoray:
I don't think we can avoid using a different image for Travis. We need XCode 9 to build the iPhone X features.

yupferris:
> I don't think we can avoid using a different image for Travis. We need XCode 9 to build the iPhone X features.

Yep, I agree, and I also like specifying explicitly which image we need for a given CI configuration. I'm still trying to think of any unintended consequences this might have, but unless the XCode 9 APIs have broken in ways we haven't predicted (which we surely would have seen locally at least), I'm struggling to come up with any. Unless something totally bogus was also changed with the image, which is hard to foresee anyways.

mortoray:
This image only affects testing on Travis, so as long as it runs there it should be fine. It doesn't influence any binaries we use, release, or otherwise test. It's isolated to Travis.

Users needed to have XCode 9 already to export to iOS with Fuse.

mortoray:
This should be tested on a few more hardware devices before merging. I've done several iOS combinations in an emulator, and two actual devices, but no hardware iPhone X.

I've done only two Android tablets, but do not have any bevelled/margin/weird ones to test borders on. The changes to Android were less significant, mainly shuffling code and hiding stuff, the same backing APIs were used as before.

yupferris:
I tested this on a few devices; it looks sensible to me.

PR #1031, opened 2018-02-20
Change some members from protected internal to just internal

ollef:
These protected internal members returned or took as parameter internal entities, which will soon be an error in Uno (matching C#). A member being protected internal means that it's visible to derived classes outside the package, but this wasn't really possible for these members since they were mentioning internal entities.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Common/Tests/FuseTest/ObservableList.uno
+++ b/Source/Fuse.Common/Tests/FuseTest/ObservableList.uno
@@ -82,7 +82,7 @@ namespace FuseTest
 			get { return _observers != null && _observers.Count > 0; }
 		}
 		
-		virtual protected void OnSubscribe(IObserver observer)
+		virtual void OnSubscribe(IObserver observer)
 		{
 			observer.OnNewAll(this);
 		}
@@ -250,7 +250,7 @@ namespace FuseTest
 			_value = initialValue;
 		}
 		
-		protected override void OnSubscribe(IObserver observer)
+		override void OnSubscribe(IObserver observer)
 		{
 			observer.OnSet(_value);
 		}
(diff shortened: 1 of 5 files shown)


Source/Fuse.Common/Tests/FuseTest/ObservableList.uno
```diff
@@ -10,7 +10,7 @@ namespace FuseTest
 		
 		NOTE: Taken from the charting in premiumlibs. Stuck in FuseTest for now since there is no use in fuselibs yet other than for testing.
 	*/
-	public abstract class ObservableData : IObservable
+	abstract class ObservableData : IObservable
```
mortoray: Why have these test classes been changed to `internal`?

ollef: Otherwise these classes expose the internal class IObserver publicly, which will become a compilation error soon.

mortoray: They don't expose it. The implementation methods are private. It's valid to implement an internal internal interface so long as you don't make public implementation functions. C# supports implementing internal interfaces, and we depend on this in many places.

ollef:
Changelog added.

Source/Fuse.Common/Tests/FuseTest/ObservableList.uno
ollef: Sure, but e.g. this line basically means it's exposed publicly if the surrounding class is public.

```
virtual protected void OnSubscribe(IObserver observer)
```

mortoray: I think you can just make that function, `OnSubscribe` internal. It's only used within that package.

These classes in test are supposed to be public, to simplify testing, otherwise we'll end up with a bunch of InternalsVisibleTo.  If they use an internal method that's okay, but that bit wasn't meant to be leaking.

ollef: Yeah, that's better. Fixed!

PR #1035, opened 2018-02-21
Refactor ModelJavaScript constructor

sebbert:
- ModelJavaScript's constructor made private rather than internal
- Now takes a ModelData instance rather than destructuring

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Models/ModelJavaScript.uno
+++ b/Source/Fuse.Models/ModelJavaScript.uno
@@ -56,7 +56,7 @@ namespace Fuse.Models
 			if (md.NameTable == null || md.ModulePath == null)
 				return;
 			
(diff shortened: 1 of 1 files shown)


sebbert:
```
Source\Fuse.Models\ModelJavaScript.uno(139.12): E4128: Fuse.Models.ModelJavaScript.ModelData is less accessible than 'Fuse.Models.ModelJavaScript..ctor(Fuse.Models.ModelJavaScript.ModelData,[string])'
```

Weird, I'm not getting that error when building locally

sebbert:
Seems like I was on an older version of Uno. Fixed by making the constructor private rather than internal (which I'd argue it should have been anyway).

PR #1036, opened 2018-02-21
Model: Introduce ModelArgs property

sebbert:
Adds support for passing constructor arguments to the model class.

`MyComponent.ux`

    <Panel
        ux:Class="MyComponent"
        Model="Models/MyComponent"
        ModelArgs="this, {foo}, 123, {ReadProperty Title}"
    />

`Models/MyComponent.js`

    export default class {
        constructor(view, foo, oneTwoThree, titleProperty) {
            // ...
        }
    }

This is currently only implemented for component models, since
app-level model arguments would be a bit more convoluted
(state migration would need to be disabled in this case).

This PR contains:
- [x] Changelog
- [ ] Documentation (N/A, docs live in an article on fuse-docs)
- [x] Tests

### DIFF
--- a/Source/Fuse.Models/ModelJavaScript.uno
+++ b/Source/Fuse.Models/ModelJavaScript.uno
@@ -16,6 +16,8 @@ namespace Fuse.Models
 		{
 			public string ModulePath;
 			public NameTable NameTable;
+			public IExpression Arguments;
+			public bool HasArguments = false;
 		}
 		static PropertyHandle _modelHandle = Properties.CreateHandle();
 		
@@ -46,6 +48,21 @@ namespace Fuse.Models
 		public static NameTable GetModelNameTable(Visual v)
 		{
 			return GetOrCreateModelData(v).NameTable;
+		}
+
+		[UXAttachedPropertySetter("ModelArgs")]
+		public static void SetModelArgs(Visual v, IExpression args)
+		{
+			var md = GetOrCreateModelData(v);
+			md.Arguments = args;
+			md.HasArguments = true;
+			OnModelDataChanged(md, v);
+		}
+
+		[UXAttachedPropertyGetter("ModelArgs")]
+		public static IExpression GetModelArgs(Visual v)
+		{
+			return GetOrCreateModelData(v).Arguments;
 		}
 
 		static void OnModelDataChanged(ModelData md, Visual v)
@@ -80,6 +97,39 @@ namespace Fuse.Models
 			rootVisualProvider.Root.Children.Add(appModel);
 		}
 
+		static IExpression[] UnpackArgs(IExpression argsExpr)
+		{
+			var vector = argsExpr as Reactive.Vector;
+			if (vector != null)
+			{
+				var vectorArgs = vector.Arguments;
+				var outputArgs = new IExpression[vectorArgs.Count];
+				for (var i = 0; i < vectorArgs.Count; ++i)
+					outputArgs[i] = vectorArgs[i];
+
+				return outputArgs;
+			}
+			
+			return new[] { argsExpr };
+		}
+
+		string GenerateArgsStringAndPopulateDependencies()
+		{
+			var argsString = "";
+			if (!_hasArgs)
+				return argsString;
+
+			var args = UnpackArgs(_args);
+			for (var i = 0; i < args.Length; ++i)
+			{
+				var depName = "__modelArg" + i;
+				argsString += ", " + depName;
+				Dependencies.Add(new Dependency(depName, args[i]));
+			}
+
+			return argsString;
+		}
+
 		void SetupModel()
 		{
 			if (_modulePath == null)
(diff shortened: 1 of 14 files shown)


Source/Fuse.Models/ModelJavaScript.uno
```diff
@@ -80,6 +97,39 @@ namespace Fuse.Models
 			rootVisualProvider.Root.Children.Add(appModel);
 		}
 
+		static IExpression[] UnpackArgs(IExpression argsExpr)
+		{
+			var vector = argsExpr as Reactive.Vector;
```
mortoray: IS this guaranteed to only trigger if the expression is something like  `a,b,c` instead of say a single value `list` that happens to be a vector?

Source/Fuse.Models/ModelJavaScript.uno
```diff
@@ -126,22 +177,38 @@ namespace Fuse.Models
...
+			_modulePath = md.ModulePath;
+			_args = md.Arguments;
+			_hasArgs = md.HasArguments;
 			FileName = "(model-script)";
-			SetupModel();
 		}
 		
 		protected override void OnRooted()
 		{
+			// TODO: Doing anything prior to base.OnRooted is questionable.
```
mortoray: Can you address this TODO?  Perhaps make a new virtual function in the base class called by OnRooted.

Source/Fuse.Models/Tests/Model.Test.uno
```diff
@@ -11,6 +11,53 @@ namespace Fuse.Models.Test
 {
 	public class ModelTest : ModelTestBase
 	{
+
+		[Test]
+		public void Args_Each()
```
mortoray: Unusual naming convention, we usually don't have the underscore.

Source/Fuse.Models/Tests/Model.Test.uno
```diff
@@ -11,6 +11,53 @@ namespace Fuse.Models.Test
...
+				Assert.AreEqual(3, e.innerInstanceCount.Value);
+				Assert.AreEqual(1, e.outerInstanceCount.Value);
+			}
+		}
+
+		[Test]
+		public void Args_ScrollView()
+		{
+			var e = new UX.Model.Args.ScrollView();
+			using (var root = TestRootPanel.CreateWithChild(e))
```
mortoray: As this test requires a suitable height you should provide an explicit size to CreateWithChild.

sebbert: Good call. By the way, this test is really only there to check if we can call methods on a ScriptClass instance that was passed via an argument, so if you know of any nodes with a ScriptClass that would be better suited for this purpose, that'd be great :)

Source/Fuse.Models/Tests/Model.Test.uno
sebbert: 👍

Source/Fuse.Models/ModelJavaScript.uno
sebbert: I assume you are referring to the case where I have a binding (like `ModelArgs="{list}"`), where the binding would resolve to a `Fuse.Reactive.Vector`? If so, I don't see how that could be the case. Since `SetModelArgs` takes an `IExpression`, the UX compiler will provide it with the expression tree rather than the result of evaluating the expression. In the case of `{list}`, that would become a `new Fuse.Reactive.Data("list")`, the dynamic cast would evaluate to `null`, and we'd hit the other case, where we return an array of just that one expression.

sebbert: Oh, I guess you mean something like `ModelArgs="list"` where `list` is a `Fuse.Reactive.Vector` that is in scope? I'm fairly certain  that would become a `new Fuse.Reactive.Constant(list)`, but I'm not 100% sure. I'll have to test this a bit further.

mortoray: Yes, those two caess, where `{list}` binds to some kind of vector type. But I guess you're correct, you're checking the actual IExpression type, not the resulting type.

The other case `ModelArgs="list"` where `list` is a global would be interesting to test. I don't think it'd come up much, but I think we'd be safer if we had some tests covering these cases.

sebbert: Indeed. I'll get a test in for that.

sebbert: @mortoray Yes, given the following UX:

```
<Fuse.Reactive.Vector ux:Name="vec" ux:AutoBind="false" />
<Panel Model="UX/Args.VectorInScope" ModelArgs="vec" />
```

The UX compiler will generate something like the following (cleaned up for brevity)

```
vec = new global::Fuse.Reactive.Vector();
var temp = new global::Fuse.Controls.Panel();
var temp1 = new global::Fuse.Reactive.Constant(vec);
global::Fuse.Models.ModelJavaScript.SetModelArgs(temp, temp1);
```

PR #1038, opened 2018-02-21
onParamaterChanged crash-fix

kusma:
Seems our change to always reflect the arguments to ScriptClass-methods lead to breakage of the `onParameterChanged`-API. But we had no test that would notice.

So let's fix that. In general, we should make FunctionMirror a bit less insane (somehow). But that's not so good for a stable-branch. So let's use some internal API trickery for now.

This PR contains:
- [x] Changelog
- [x] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 
 # 1.7
 
+### 1.7.1
+
+# Fuse.Nodes
+- Fixed a bug where triggering `onParameterChanged` on an element would lead to a crash.
+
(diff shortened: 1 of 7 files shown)


kusma:
I'm not surprised by the error at all; `Context.Reflect()` really shouldn't wrap `Scripting.Function` at all. However, ATM avoiding that would break EventBinding, so we can't get rid of the wrapping without rewriting that code.

yupferris:
Nice one!

Issue #1005, opened 2018-02-05
`Path` isn't rendering in exported view


trusis666:
[Reported on a forum.](https://www.fusetools.com/community/forums/fuse_studio/path_isnt_rendering_in_exported_view)

I have a simple component that renders out QR code using SVG <Path/>.

This works perfectly when running in preview mode (locally & on device), but it shows nothing when I export that view and embed it in a native app (tested only on iOS so far).

I am attaching isolated component:

https://dsc.cloud/vv/TestProject-standalone.zip - for testing local preview
https://dsc.cloud/vv/TestProject-exported-views.zip - for testing exported view

On a side note, when I mark view for export using <ExportedView/> + ux:Class property on a <Panel/> (as in attached), I am losing ability to preview locally, so I have to keep adding/removing those bits whenever I want to switch between local testing and embedded testing. Documentation doesn't provide answers/hints, am I missing some tricks there?

If necessary, I can provide iOS test project too (however, it doesn't do anything unusual!)

nekronos:
If you want to preview locally, just do:
```xml
<App>
	<ExportedViews>
		<MyClass ux:Template="MyClass" />
	</ExportedViews>

	<Panel ux:Class="MyClass" />

	<MyClass />
</App>
```

nekronos:
Looks like something goes wrong with the native UI implementation of path rendering. As a workaround you can do: 
```xml
<ExportedViews>
    <GraphicsView ux:Template="QRCode">
        <QRCodePanel />
    </GraphicsView>
</ExportedViews>
```

mortoray:
You should be able to test that native rendering using a `NativeViewHost` on the type, without an exported view.

The backend rendering will be the same on the target regardless whether it's Native, GL, or the exported-Native/GL.

nekronos:
@mortoray the native surface test app is working, could there be invalidation issues?

mortoray:
I don't know. How hard is it to setup the test of using this exported view. Do you have a unoproj and XCode project I could use as a reference?  (I've never created a non-fuse mobile app, so don't know how to do that bit)

Invalidation does sound like a likely possibility. But if other exported views work that's a bit weird. Espeically since it's the native controls failing, not the GL ones.

Potentially this relates to a `Viewport`. I assume FuseViews must expose a `Viewport`, but maybe the control isn't finding it.

nekronos:
the issue happens in normal build aswell with this app

mortoray:
[...]

PR #1059, opened 2018-02-28
TreeRenderer: add RenderBoundsChanged

nekronos:
The forwarding of renderoubnds relied on the Placed event triggering,
which might not happen when only RenderBounds are invalidated

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Elements/Element.Validation.uno
+++ b/Source/Fuse.Elements/Element.Validation.uno
@@ -45,7 +45,22 @@ namespace Fuse.Elements
 			if (ElementBatchEntry != null)
 				ElementBatchEntry.InvalidateRenderBounds();
 
+			if (!_hasNotifiedRenderBoundsChanged)
+			{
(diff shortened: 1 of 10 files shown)


nekronos:
Fixes https://github.com/fusetools/fuselibs-public/issues/1005

mortoray:
Why is `interface ITreeRenderer`? Should this be an `internal` class? I know this PR didn't touch that, but perhaps we should consider changing it. It's not somethign we want users implementing.

Source/Fuse.Elements/Element.Validation.uno
```diff
@@ -45,9 +45,24 @@ namespace Fuse.Elements
 			if (ElementBatchEntry != null)
 				ElementBatchEntry.InvalidateRenderBounds();
 
+			if (!_hasNotifiedRenderBoundsChanged)
+			{
+				UpdateManager.AddDeferredAction(NotifyRenderBoundsChanged, UpdateStage.Layout, LayoutPriority.Post);
```
mortoray: Why is this dispatch done in the Layout stage and not just in the current stage?

PR #1061, opened 2018-03-01
Model: Use Set to track visited nodes in evaluateDerivedProps

sebbert:
Given an efficient implementation, this should bring lookup time
down to O(1), instead of always doing a linear search.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


sebbert:
@mortoray Ah, it does indeed use the ES6 `Set` type. We use a `Set` for tracking the visited node in the main diff algoritm as well. I'll write a polyfill.

PR #1063, opened 2018-03-01
Model: Add Set polyfill

sebbert:
This adds an implementation of ES6 Set that uses an array internally,
and chooses this implementation if the native Set is unavailable.

The polyfill is only applied locally in Model.js, and does not pollute
the global object.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Models/Fuse.Models.unoproj
+++ b/Source/Fuse.Models/Fuse.Models.unoproj
@@ -21,5 +21,6 @@
     "FuseJS/Internal/ZonePatches.js:Bundle",
     "FuseJS/Internal/Model.js:Bundle",
     "FuseJS/Internal/ViewModelAdapter.js:Bundle",
+    "FuseJS/Internal/ArraySet.js:Bundle",
   ]
 }
--- a/Source/Fuse.Models/FuseJS/Internal/ArraySet.js
+++ b/Source/Fuse.Models/FuseJS/Internal/ArraySet.js
@@ -1,0 +1,72 @@
+"use strict";
+
+function ArraySet(initialItems) {
+	if (initialItems == null) {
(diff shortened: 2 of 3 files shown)


sebbert:
@mortoray I found [this library](https://github.com/medikoo/es6-set) which seems to have good test coverage and is MIT licensed. I think using something more mature like that might be a better idea.

Re. preformance: I guess we could do a Uno implementation instead. However, as far as I've understood, `Scripting.Object` does not (yet) have well-defined equality semantics (`Equals` and `GetHashCode` look at the handle, not the object reference itself). So this would have to be done first.

mortoray:
That lib might be overkill (it adds size to our distro then) -- and a confusing structure. :/
I'm happy with just a few tests for this minimal Set. Maybe hold off on a full implementation until somebody requests `Set` support (what happens now if you compile using Set with Babel, does it perhaps add one?)

sebbert:
Hmm, thinking about it, for this one case I guess we could actually use object keys. Instead of storing a reference to the node in question, we could add its ID number as a key to an object. The object would act as a hash map, and in most cases, membership tests should be O(1).

sebbert:
@mortoray That sounds much saner indeed. I'll get some basic tests up and running.

Re. babel: I guess that would depend on your specific babel configuration, but as far as I know, the standard is to include the polyfills you need yourself.

sebbert:
@mortoray Turns out, Set has existed in JavaScriptCore since iOS 8. I'll test it a bit on an older device and see if there are any issues. I'll close this until it's deemed necessary.

sebbert:
Yep, Set verified to work under iOS 8.

PR #1079, opened 2018-03-07
Add optional onProperties callback to Model viewmodels

Sunjammer:
Making view models that use ux properties is currently very awkward. There is no signal to tell you when all properties are ready to be read, there is no interface to get at the observables so you can do a combineLatest of your own. This makes view model code that reads properties sort of a mess of derived state, resulting in idiosyncratic code full of null checks and other things that are boring and bad.

This PR adds a pass to the view model adapter that, if observables are found on the view and the model has a function called onProperties, sets up a combineLatest subscription that calls onProperties with a dictionary of propertyName:propertyValue.

This creates an alternative constructor for viewmodels using ux properties where the developer can reliably access all ux properties without hoping the diffing will settle into a good state in some indeterminate future, or respond to unexpected differences between the model and the view.  
This makes ux properties a reliable and useful way to ferry data into viewmodels without introducing a user-code dependency on the Observable API.


kusma:
Seems like this isn't moving forward. @Sunjammer: are you planning on following this up, or should we just close it for now?

kusma:
OK, I'll close it for now. @Sunjammer: if you ever get the interest to pick this up again, please re-open when you have the test-cases @mortoray asked for...

PR #1099, opened 2018-03-13
Video: Use OpenGL interface from Uno

Tapped:
The old implementation was loading its own GL entrypoints, however to support for example Angle it has to use Uno's GL interface.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
OK, I've rebased it now! Let's see how CI goes!

Issue #1096, opened 2018-03-13
Android Viewport.Size is reporting 0,0 at startup


mortoray:
It appears that we're getting a 0,0 size on startup of an app on Android in 1.8.  This is leading to the error in #1093 .

The defect in 1093 can be fixed, but we still shouldn't be getting this 0,0 size. It causes an unecessary layout calculation pass, which can be quite costly, and the results will just be discarded once the real size comes.

It's not possible to just ignore a 0,0 size in Layout. It must be honoured as there are scenarios when components truly have 0,0 size. Though the RootViewport probably should never be 0,0, it feels like the wrong fix to just bypass it there.

PR #1101, opened 2018-03-13
Fixes an empty layout on Android during startup

mortoray:
fixes https://github.com/fusetools/fuselibs-public/issues/1096

This PR contains:
- [x] Changelog
- [ ] ~Documentation~
- [x] Tests - ManualTestApp will go all red if it detects one of the symptoms

### DIFF
--- a/Source/Fuse.Android/AndroidApp.uno
+++ b/Source/Fuse.Android/AndroidApp.uno
@@ -53,7 +53,7 @@ namespace Fuse
 
 			Uno.Platform.Displays.MainDisplay.Tick += OnTick;
 		}
-
+		
 		public sealed override IList<Node> Children
 		{
 			get { return RootVisual.Children; }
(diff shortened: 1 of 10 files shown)


Source/Fuse.Android/AndroidApp.uno
```diff
@@ -53,7 +53,7 @@ namespace Fuse
 
 			Uno.Platform.Displays.MainDisplay.Tick += OnTick;
 		}
-
+
```
kusma: This seems silly

mortoray:
@kusma I thought about a hard error, but it makes the fix more challenging. We have several controls that end up reading frame state prior to the viewport being ready, things like `ClientPanel`, `StatusBarFrame`, or triggers like `WhileWindowPortrait`, and the `window()`. There's no [easy] way to defer their creation, nor do they gain anything from having "unknown" values -- they all still need actual values.

So while not perfect, this patch fixes the most costly issue: the layout pass before ready. All those other bits don't gain us anything in performance, and will actually make the code more complex, thus I don't think it's worth it.

kusma:
Right. If so, I think the right thing is to not mark the issue as fixed, but reduce the severity of it after this PR.

mortoray:
I've created a new issue instead https://github.com/fusetools/fuselibs-public/issues/1102  It better describes the remaining situation.

PR #1108, opened 2018-03-15
Move "Safe Layout" article to fuse-docs, fix broken link, update old …

sebbert:
Seems that our docs pipeline does not support linking to markdown files inside fuselibs, just including them inline.

This moves the "Safe Layout" article to `fuse-docs`, changes doc comments that "include" that article so that they link to it instead, and fixes a broken link.

This PR contains:
- [ ] Changelog
- [x] Documentation
- [ ] Tests


sebbert:
Note that this is **not** critical for release, as it's only docs changes.

PR #1113, opened 2018-03-16
travis: update osx image

kusma:
It seems that the default OSX image on travis is having some security
issues, causing builds to fail. So let's try to upgrade the OSX image to
the latest non-beta one, so builds will hopefully work again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 os: osx
 
+osx_image: xcode9.2
+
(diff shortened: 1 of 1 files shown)


kusma:
This is just a test to see if this would solve what's blocking #1112.

The right move might instead be to upgrade python or SSL using brew before running `sudo easy_install pip`. But let's see if this helps first.

PR #1118, opened 2018-03-19
Fixup warnings

kusma:
This is just a warning clean-up, shouldn't have any behavioral changes.


kusma:
@mortoray: `Math.Round` with float and float{2,3,4} as argument, as well as `Random.NextInt(x)` are deprecated, yes.

kusma:
Do note that the `Math.Round` overloads aren't deprecated in the uno-version we're using in fuselibs yet. But they are on uno master.

PR #1124, opened 2018-03-20
Improve language in docs

kusma:
This is based on a code-review in the uno-repo (fusetools/uno#1638
for those who have access), but I thought I'd make sure we update
fuselibs as well.

This PR contains:
- [x] Documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,12 +1,11 @@
 # Contributing
 
-When contributing to this repository, please first discuss the change you
-wish to make via an issue, in our [slack](https://fusecommunity.slack.com)
-community, or any other method with the owners of this repository before
-making a change.
+If you want to contribute to this repository, please first discuss the
+changes you want to make with the repository's owners, for example via
+an issue or on [Slack](https://fusecommunity.slack.com).
 
-Please note we have a [code of conduct](#code-of-conduct), please follow
-it in all your interactions with the project.
+This project has a [code of conduct](#code-of-conduct). Please follow
+it in all your interactions related to the project.
 
 We also have a [contributor license agreement][1], that we require
 contributors to sign in order to have their changes considered for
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -15,24 +14,24 @@ inclusion.
...
-   to ensure it isn't reintroduced.
+   silently break. Bug fixes should generally have regression tests added
+   to ensure they're not reintroduced later.
 4. New features must be documented. Please read [this](Documentation/WritingDocumentation.md)
    for details on how to write documentation.
-5. If there's any functional changes, update CHANGELOG.md with a brief
+5. If there are any functional changes, update CHANGELOG.md with a brief
    explanation.
-6. Ensure that tests pass locally before sending a pull-request.
+6. Ensure that tests passes locally before sending a pull-request.
```
ollef: "pass" is correct in this case. My comment in the other PR was only meant to cover the next one.

CONTRIBUTING.md
```diff
@@ -15,24 +14,24 @@ inclusion.
 ## Pull Request Process
 
 1. Features should generally be pulled against the `master` branch.
-   Bug-fixes, depending on the urgency and risk, goes either against an
-   active release-branch, or `master`.
+   Bug-fixes, depending on the urgency and risk, go either against an active
```
ollef: Should be "Bug fixes" to match the change further down.

kusma:
I'm going to fixup some compound words also, so there's little point in reviewing right now.

CONTRIBUTING.md
kusma: Yeah, I'll fix it :)

ollef:
Too late. :) I can have another look later though, no worries!

kusma:
@ollef: OK, I think I'm done :)

Documentation/CodingStyle.md
```diff
@@ -1,24 +1,23 @@
...
-text-editor. Plug-ins exist for most popular text-editors. This ensures
+text editor. Plug-ins exist for most popular text editors. This ensures
 that our [configuration file](../.editorconfig) gets respected, which
 reduces needless commit-noise.
 
-Generally, try to follow the local code-style of the source-file you're
-editing. It's much more disturbing to swtich between multiple code-styles
-within a source-file, than between source-files.
+Generally, try to follow the local code style of the source file you're
+editing. It's much more disturbing to swtich between multiple code styles
```
ollef: Missed a `swtich`!

kusma: Whoops, good catch! Seems this is a common typo, it's two other places in the fuselibs source!

> $ git grep swtich
> Documentation/CodingStyle.md:editing. It's much more disturbing to swtich between multiple code styles
> Source/Fuse.Navigation/WhileNavigationTriggers.uno:                     As the user swipes from B to C the progress of B will reduce from 1 towards 0 and the progress of C increases from 0 towards 1. The `Threadhold="0.5"` here causes the trigger swtich at the mid-way point of the transition. In this setup this means that one page is green and the rest are red -- the one closest to active is green.
> Source/Fuse.Triggers/StateGroup.uno:                    //this is needed in case a state swtich results in a state switch, see `ChainedSwitch` test case

kusma:
@ollef: Fixed :)

PR #1150, opened 2018-04-19
remove ProjectTemplates

kusma:
These now live in the Fuse-repo instead, so let's get rid of this
unused copy.


kusma:
@anettegj: No problem, we're not in a hurry :)

Issue #1156, opened 2018-05-02
Desciption parameter is not managed in Share.shareFile(path, mimetype, description)


telatao:
1/ Create a script that calls this line:
Share.shareFile(image.path, "image/*", "Text To Share", ShareOrigin);
2/ Run it
=> Description is never shown, even in a simple email.

PR #1157, opened 2018-05-02
Fix for the ticket #1156

telatao:
1/ Create a script that calls this line:
Share.shareFile(image.path, "image/*", "Text To Share", ShareOrigin);
2/ Run it
=> Description is never shown, even in a simple email.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Share/ShareModule.uno
+++ b/Source/Fuse.Share/ShareModule.uno
@@ -142,7 +142,7 @@ namespace Fuse.Share
 
 			if defined(android || iOS)
 			{
-				var description = args.Length>1 ? "" + args[1] : "";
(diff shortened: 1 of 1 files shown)


Source/Fuse.Share/ShareModule.uno
```diff
@@ -151,7 +151,7 @@ namespace Fuse.Share
 				else if defined(iOS)
 				{
 					float2 position = float2(0);
-					if (args.Length > 2 && TryGetPosition(args[2], out position))
+					if (args.Length > 2 && TryGetPosition(args[3], out position))
```
kusma: You'll need to bump the args.Length-check here also...

PR #1158, opened 2018-05-02
Description is not managed in both NewShareFileActivity and NewShareT…

telatao:
Description is not managed in both NewShareFileActivity and NewShareTextActivity

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
@telatao: Would you care to explain a bit more about what this does and why?

telatao:
Sure!
What this does? 
ShareImage
Linking description to the given image. It works with mail, copy/past, Twitter... on FaceBook, only image will be manage, but this belongs to FaceBook, not us.

ShareText
Display the given description.

Why?
The description was not given to the SharedArray and it must be given in order to be displayed

PR #1165, opened 2018-05-09
Fixup pack.sh

kusma:
This fixes two things:
- pack.sh was broken since we started using uno from nuget, as we no longer have uno.stuff.
- we also failed to pass --version to `uno pack`, leaving all packages with version 0.0.0.


kusma:
@BeeWarloc: I believe this fixes the problems you were having with the latest generated packages...

Issue #752, opened 2017-11-16
multiple push notifications have no way to distinguish which one was clicked on


eksperts:
attn. @cbaggers 

[Reported on community](https://www.fusetools.com/community/forums/general/push_notifications_body_json_structure?page=1&highlight=42379112-d6f4-405f-a80c-50d62f40b1d9#post-42379112-d6f4-405f-a80c-50d62f40b1d9). Start reading at the highlighted post to the end of the thread.

In particular: `How to detect on which notification I have clicked ? When I click on one notification, all pending notification are delievered in order as they arrived (first the oldest then others) and they all have value for fromNotificationBar "true" and I dont have none information which notification I have clicked.`

Suggestion for a solution:
it is perfectly OK that all of the pending notifications are all delivered to the app when it enters foreground - this is the behaviour we have always had.
it is NOT OK that all of the delivered notification have `fromNotificationBar` set to `true`, and supposedly only the one clicked on should have that.

fusebuild:
The forum thread [Push Notifications Body JSON Structure](https://www.fusetools.com/community/forums/general/push_notifications_body_json_structure) posted by Levi Olson was linked to this issue.

AndrewEQ:
Hi there, just reposting my feedback from forum: https://forums.fusetools.com/t/android-push-notifications-payload-event-not-fired-if-there-are-more-than-1-notifications-in-the-list/7125

Fuse 1.8.1
Marshmallow

If I send the device a single push notification at a time and open it each time before sending the next one, the payload is 100%, tested via lock screen and via tray.

If I send the device several push notifications, the following happens:

- after opening any notification (could open the latest received) via lock screen or via tray, the payload of the 1st notification received is returned.

- after opening subsequent notifications via the lock screen or via tray, the payload event is not fired.

PR #1185, opened 2018-05-22
Fix Push Notifications

devadiab:
Regarding to forum posts talking about a problem with unintended behavior on `Fuse.PushNotifications` that represent the following scenario : 
1- Our app received multiple notifications on the notifications bar.
2- When we click on the first one, the notification will be clickable and `` will fired also `push.on("receivedMessage" .. `.
3- The rest of notifications won't work.

I have fixed the flags for the pending intent in `Impl.uno` so that the app will receive multiple notifications and all notifications are working and doing fine

### DIFF
--- a/Source/Fuse.PushNotifications/Android/Impl.uno
+++ b/Source/Fuse.PushNotifications/Android/Impl.uno
@@ -296,13 +296,14 @@ namespace Fuse.PushNotifications
 		[Foreign(Language.Java)]
 		static void SpitOutNotification(Java.Object _listener, string title, string body, string bigTitle, string bigBody, string notificationStyle, string featuredImage, string sound, Java.Object _payload)
 		@{
+			int id = PushNotificationReceiver.nextID();
 			Context context = (Context)_listener;
 			Bundle payload = (Bundle)_payload;
 			Intent intent = new Intent(context, @(Activity.Package).@(Activity.Name).class);
-			intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+			intent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP);
 			intent.setAction(PushNotificationReceiver.ACTION);
 			intent.replaceExtras(payload);
-			android.app.PendingIntent pendingIntent = android.app.PendingIntent.getActivity(context, 0, intent, android.app.PendingIntent.FLAG_ONE_SHOT);
+			android.app.PendingIntent pendingIntent = android.app.PendingIntent.getActivity(context, id, intent, android.app.PendingIntent.FLAG_UPDATE_CURRENT);
 			android.app.NotificationManager notificationManager = (android.app.NotificationManager)context.getSystemService(Context.NOTIFICATION_SERVICE);
 
 			NotificationCompat.Builder notificationBuilder = new NotificationCompat.Builder(context)
@@ -318,7 +319,6 @@ namespace Fuse.PushNotifications
 				notificationBuilder.setSound(defaultSoundUri);
 			}
 
-			int id = PushNotificationReceiver.nextID();
 
 			if (notificationStyle != null && !notificationStyle.isEmpty())
 			{


devadiab:
This PR solves the problem of this issue #752

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -296,13 +296,15 @@ namespace Fuse.PushNotifications
 		[Foreign(Language.Java)]
 		static void SpitOutNotification(Java.Object _listener, string title, string body, string bigTitle, string bigBody, string notificationStyle, string featuredImage, string sound, Java.Object _payload)
 		@{
+	        int requestID = (int) System.currentTimeMillis();
```
kusma: Please indent with tabs here...

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -296,13 +296,15 @@ namespace Fuse.PushNotifications
...
 			Context context = (Context)_listener;
 			Bundle payload = (Bundle)_payload;
 			Intent intent = new Intent(context, @(Activity.Package).@(Activity.Name).class);
-			intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+        	intent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP);
```
kusma: Please indent with tabs here...

kusma: Hmm, [the documentation](https://developer.android.com/reference/android/content/Intent.html#FLAG_ACTIVITY_CLEAR_TOP) for `FLAG_ACTIVITY_CLEAR_TOP` says

> If set, and the activity being launched is already running in the current task, then instead of launching a new instance of that activity, all of the other activities on top of it will be closed and this Intent will be delivered to the (now on top) old activity as a new Intent.

...and [the documentation](https://developer.android.com/reference/android/content/Intent#FLAG_ACTIVITY_SINGLE_TOP) for `FLAG_ACTIVITY_SINGLE_TOP` says:

> If set, the activity will not be launched if it is already running at the top of the history stack.

...these two immediately sounds a bit like they are incompatible to me. However, further down on the documentation for `FLAG_ACTIVITY_CLEAR_TOP`, it seems to suggest they can be combined. :confused: 

@cbaggers: Does this make sense to you? I'm probably not smart enough to wrap my head around this right now ;)

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -296,13 +296,15 @@ namespace Fuse.PushNotifications
...
 			intent.setAction(PushNotificationReceiver.ACTION);
 			intent.replaceExtras(payload);
-			android.app.PendingIntent pendingIntent = android.app.PendingIntent.getActivity(context, 0, intent, android.app.PendingIntent.FLAG_ONE_SHOT);
+
+			android.app.PendingIntent pendingIntent = android.app.PendingIntent.getActivity(context, requestID, intent, android.app.PendingIntent.FLAG_UPDATE_CURRENT);
```
kusma: Excuse me if I'm being stupid here, but let me check if I'm understanding this correctly...

requestCode = 0 is replaced with System.currentTimeMillis(), to make each request unique, and then replace `FLAG_ONE_SHOT` is replaced with `FLAG_UPDATE_CURRENT` to deal with the off-chance of two push-notifications triggering within the same millisecond (or whatever granularity System.currentTimeMillis() has)?

I think a short explanation about the intent of the changes would make sense to put in the commit-message.

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -318,7 +320,7 @@ namespace Fuse.PushNotifications
 				notificationBuilder.setSound(defaultSoundUri);
 			}
 
-			int id = PushNotificationReceiver.nextID();
+			int id = requestID;
```
kusma: OK, reading this a bit more, I feel relatively sure I've misunderstand what the code tries to do in my explanation above. It would seem to me like `PushNotificationReceiver.nextID()` would be pretty much the perfect way of making each request unique. So why don't we want to use it?

Source/Fuse.PushNotifications/Android/Impl.uno
cbaggers: @kusma It is a bit odd it's one i'd like to test later this week when I have a free evening. 
@devadiab thanks for the PR! does this work without the change on this line? if not, what problem is caused and how did it manifest?

Source/Fuse.PushNotifications/Android/Impl.uno
cbaggers: @devadiab I'd prefer using `PushNotificationReceiver.nextID()` here.

cbaggers: @devadiab if possible it would also be great to hear why we had to change the `FLAG_UPDATE_CURRENT` flag. As @kusma says having that in the commit message would be great. Android is such a sea of weirdness that trying to understand the logic of changes like this gets really hard 8 months down the line :)

devadiab: @kusma , @cbaggers  Thanks guys for you follow-up. about the Id i was thinking to return it to `PushNotificationReceiver.nextID()` .
About the  `FLAG_UPDATE_CURRENT` flag, android says: 
> Flag indicating that if the described PendingIntent already exists, then keep it but replace its extra data with what is in this new Intent. 
which mean that it will let the app received the new payload and extras.

Source/Fuse.PushNotifications/Android/Impl.uno
devadiab: Okay, done i have changed it on my own local repo.

devadiab:
@kusma Can you review that last commit ?

kusma:
@devadiab: That looks a lot better, thanks a lot! But would you mind squashing it all down into one commit, and give it a bit better commit-message?

kusma:
@cbaggers: I think you're a bit better suited than me to review the remaining bits ;)

Source/Fuse.PushNotifications/Android/Impl.uno
```diff
@@ -296,13 +296,14 @@ namespace Fuse.PushNotifications
 		[Foreign(Language.Java)]
 		static void SpitOutNotification(Java.Object _listener, string title, string body, string bigTitle, string bigBody, string notificationStyle, string featuredImage, string sound, Java.Object _payload)
 		@{
+			int id = PushNotificationReceiver.nextID();
 			Context context = (Context)_listener;
 			Bundle payload = (Bundle)_payload;
 			Intent intent = new Intent(context, @(Activity.Package).@(Activity.Name).class);
-			intent.addFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP);
+			intent.setFlags(Intent.FLAG_ACTIVITY_CLEAR_TOP | Intent.FLAG_ACTIVITY_SINGLE_TOP);
```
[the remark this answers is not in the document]
devadiab: `Intent.FLAG_ACTIVITY_CLEAR_TOP` used to clear the Stack while the `FLAG_ACTIVITY_SINGLE_TOP` will not be launched if it is already running at the top of the history stack.

So clearing the stack then run without any previous information their.
@cbaggers still need your sign-off 👍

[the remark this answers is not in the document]
kusma: I think there's a limit to how long I'm willing to wait. Let's just merge this as-is, if something is wrong, we can always fix this later...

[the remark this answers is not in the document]
devadiab: Appreciated 💯

PR #1204, opened 2018-06-30
Upgrade ScrollViewPager to support multi-column layouts

devadiab:
Dear Sirs,

I have made a small changes on `ScrollViewPager` by adding an attribute called `GridSize` for supporting multi-column layouts which will fix all scroll shaking and layout problems. 

Only you have to set the `GridSize` attribute of the current `ScrollViewPager`  that represent the columns count you have inside your `ScrollView`.

Best,
Ahmad

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+### ScrollViewPager
+- Added a new attribute called (`GridSize` = 1 by default) to let `ScrollViewPager` supports multi-column layout. 
+
 ### Fuse.Deprecated.CameraView
 - This obsolete package has been removed. All functionality should be present in `Fuse.Controls.CameraView` instead.
 
--- a/Source/Fuse.Controls.ScrollView/ScrollViewPager.uno
+++ b/Source/Fuse.Controls.ScrollView/ScrollViewPager.uno
@@ -70,19 +70,19 @@ namespace Fuse.Controls
 			}
 		}
 
-		int _gridSize = 1;
+		int _columnCount = 1;
 		/**
 			An optional attribute that will maintain the layout in-case of using Grids or any layout seems to be columns.
 		*/
-		public int GridSize
+		public int ColumnCount
 		{
-			get { return _gridSize; }
+			get { return _columnCount; }
 			set
 			{
-				if (_gridSize == value)
+				if (_columnCount == value)
 				return;
 				
-				_gridSize = value;
+				_columnCount = value;
 			}
 		}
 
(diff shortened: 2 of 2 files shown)


mortoray:
With a WrapPanel though you won't know how many items are on the row, it may also differ per row. Only the GridLayout and ColumnLayout have a readable attribute that says how many items are on the line.

devadiab:
@mortoray I can set how many items per row by defining the width for each one related to whole container width like 
```
<StackPanel ux:Name="container">
	<WrapPanel>
			<Panel Width="width(container)/3"/>
			<Panel Width="width(container)/3"/>
			<Panel Width="width(container)/3"/>
	</WrapPanel>
</StackPanel>
```

That what i meant 👍

devadiab:
@mortoray about the `GridLayout` columns it's easy to get the number of columns but when we using soming like this `<Grid Columns="60,1*,auto" />` i think we will turned into too many cases. we can have future work on it to make it somehow automatically.

mortoray:
That `WrapPanel` example is really not a good way of doing the layout. A `Grid Columns="1*,1*,1*"` would be more appropriate.

The Grid always has the number of columns in it. It doesn't matter what sizes those columns are. Your example would have 3 columns, and that's all that's needed for the paging to work.

devadiab:
I totally understand what you mean and i expected that, but this way is still active and works fine what i mean is cover all cases of UX.
Then we have to notice that we should use `Grid` with `Pager` unless a problems may occurs.
The `GridSize` attribute then can be optional and we will set its value to the Grid Columns value sounds good ?

mortoray:
I'd be okay with an explicit option as a quicker solution.  However, the name should be `ColumnCount` or similar, not `GridSize`. Be explicit about what the value is.

I'd also like to see the code handle the situation when the Offset or Limit aren't multiples somehow. This could happen during dynamic loading of items or recycling of views. It's sufficient to recover from this situation, not actually support it.

Test cases showing this works are needed -- this is historically the type of stuff that fails easily when stuff changes.

devadiab:
Thanks @mortoray for your comments, I have already implemented dynamically setting the `GridSize` or 'ColumnCount' when the parent of the `Each` is `<Grid />` Parent but i failed to have the same for `GridLayout` or `ColumnLayout` due to the enumurator for the Ancestors and Successors applies only for `Nodes` and both `GridLayout` and `ColumnLayout` are `PropertyObject` which is the parent of `Node` as well but the template methods signature such as `NextSibling` have the following signature :
```
public T NextSibling<T>() where T: Node
```
which let me fail.
BTW I will push another commit for the `ScrollViewPager` right now and it should be have the new name for the attribute called `ColumnCount` and will update the Changelog file as well.

devadiab:
@mortoray  can you review last commit please?

kusma:
It's been more than a year without feedback from the author, I'm closing this.

@devadiab: If you still want to pursue this, please reopen and address the feedback!

Issue #932, opened 2018-01-11
FileUriExposedException on Android >=8 using FuseJS/Camera


eksperts:
[Reported on community](https://www.fusetools.com/community/forums/general/issue_camera_crash).

## Version fuse and OS
Fuse 1.5.0, macOS High Sierra

## What do I expect?
I expect the URL of the photo to be printed on the console correctly when i use takephoto of Fuse.Camera Module

## What's going on?
An exception thrown `System.err: android.os.FileUriExposedException`. This happens with Android versions > = 8.0.  With earlier versions and with iOS it works properly.
I have noticed that the cause is having added these specs in the project file:
```
"Android": {
    "SDK": {
      "BuildToolsVersion": "25.0.3",
      "CompileVersion": 25,
      "MinVersion": 16,
      "TargetVersion": 25
    }
  }
```

## Reproduce the issue
Mainview.ux
```
<App>
	<JavaScript>
		
		var camera = require('FuseJS/Camera');
		var cameraRoll = require("FuseJS/CameraRoll");

		function takePhoto(){
			console.log("take photo");
			camera.takePicture(1920, 1080).then(function(newImage) {
		//HERE CRASH 
				console.log(newImage.path);
				}).catch(function(error) {
					cameraRoll.getImage().then(function(imageRoll) {
					console.log(imageRoll.path);
					}.bind(this),
						function(error) {
							console.log("error", JSON.stringify(error));
						});
			});
		}
		module.exports={takePhoto:takePhoto}
	</JavaScript>

	<Button Clicked="{takePhoto}" Text="Take photo"/>
</App>

```
*.unoproj
```
{
  "RootNamespace":"",
  "Packages": [
    "Fuse",
    "FuseJS",
    "Fuse.Camera",
    "Fuse.CameraRoll"
  ],
  "Includes": [
    "*"
  ],
  "Android": {
    "SDK": {
      "BuildToolsVersion": "25.0.3",
      "CompileVersion": 25,
      "MinVersion": 16,
      "TargetVersion": 25
    }
  }
}
```

## EXCEPTION
```
System.err: android.os.FileUriExposedException: file:///storage/emulated/0/Android/data/com.sandos.precheckin/cache/images/IMG_696f3530-2789-419a-aa44-1f9a788925b9.jpg exposed beyond app through ClipData.Item.getUri()

01-09 18:23:22.701 22775 22775 W System.err: at android.os.StrictMode.onFileUriExposed(StrictMode.java:1960)
[...]

PR #1207, opened 2018-08-05
simple fix for sharing file uris to other apps

shnupta:
After experiencing a similar error as the one in #932 I've read the link specified there:
https://developer.android.com/reference/android/os/FileUriExposedException.html

And changed the `ShareModule.uno` file to share with the `content:///` prefix rather than `file:///`

It has fixed my problem when trying to use `shareFile()` on android.

### DIFF
--- a/Source/Fuse.Share/ShareModule.uno
+++ b/Source/Fuse.Share/ShareModule.uno
@@ -145,7 +145,7 @@ namespace Fuse.Share
 				var description = args.Length>2 ? "" + args[2] : "";
 				if defined(android)
 				{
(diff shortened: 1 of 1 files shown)


kusma:
@shnupta: Why was this closed?

shnupta:
Sorry forgot to add a comment.

My change fixed the exception from happening, but you need to use a
`FileProvider` for Android to serve files to other applications, with
temporary permissions.

See here:
https://developer.android.com/reference/android/support/v4/content/FileProvider

I was going to continue making changes to my branch but I really don't have
time at the moment. I may come back to it though.

kusma:
I see, thanks for clarifying :)

PR #1208, opened 2018-08-10
fix iOS Camera roll permission issue

### LINKED ISSUE #983: CameraRoll permission issue when sellecting an image
[Reported on forum](https://www.fusetools.com/community/forums/bug_reports/cameraroll_permission_issue_when_sellecting_an_ima)

MacOS Siera High
Fusetools 1.5 (build 150046)
Reproduced: iPhone X, iPhone 7, iPhone SE, iPhone 6
Not affected: Nexus 5 Android 6.0.1

We're having an issue with the permissions for the cameraroll. When selecting a image, permission is asked after selecting image and this clears the selected image. Therefore to select a image first time users have to take the same steps twice.

Is this a bug in Fuse or is there some kind of work around this issue?

Video: To make things clear: https://youtu.be/nwrvvy3HTJ0

# Expected behavior:
 Permission should be asked, before accessing images on a device. After that, I can select an image. 

# Minimal code to reproduce:
```
<App>

    <JavaScript>

        var cameraRoll = require("FuseJS/CameraRoll");
        var Observable = require("FuseJS/Observable");

        var TheImage = Observable();


        module.exports = {
            TheImage:TheImage,
            SetImage : function() {

                cameraRoll.getImage()
                .then(function(image) {
                    TheImage.value = image;
                }, function(error) {
                });
            }
        }    

    </JavaScript>
    <Button Text="test" Clicked="{SetImage}" />
    <Image File="{TheImage}"/>
</App>
```
When ruuning `fuse build --target=iOS -adebug` in Xcode I get this error:
```
[discovery] errors encountered while discovering extensions: Error Domain=PlugInKit Code=13 "query cancelled" UserInfo={NSLocalizedDescription=query cancelled}
```

LuisRodriguezLD:
Fixes #983. Before accessing the image picker I checked if `PHAuthorizationStatus` was determined. If not then get permissions, if  `PHAuthorizationStatusAuthorized` go for it, else (denied or restricted) return `onFail`.

More info on [PHAuthorizationStatus](https://developer.apple.com/documentation/photokit/phauthorizationstatus?language=objc)

Later, I want to implement `cameraRoll.accessPhotos()` to get user's permissions on demand and if it's false then redirect them to phone settings.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.CameraRoll/iOS/iOSCameraRoll.uno
+++ b/Source/Fuse.CameraRoll/iOS/iOSCameraRoll.uno
@@ -41,11 +41,24 @@ namespace Fuse.CameraRoll
 		[Foreign(Language.ObjC)]
 		static void SelectPictureInternal(Action<string> onComplete, Action<string> onFail)
 		@{
-			dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
-				[[CameraRollHelper instance] selectPictureWithCompletionHandler:onComplete onFail:onFail];
-			});
+			PHAuthorizationStatus status = [PHPhotoLibrary authorizationStatus];
+			if (status == PHAuthorizationStatusNotDetermined)
+				[PHPhotoLibrary requestAuthorization:^(PHAuthorizationStatus status){
+					if (status == PHAuthorizationStatusAuthorized)
+						dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+							[[CameraRollHelper instance] selectPictureWithCompletionHandler:onComplete onFail:onFail];
+						});
+					else
+						onFail(@"PHAuthorizationStatusDenied");
+				}];
+			else if (status == PHAuthorizationStatusAuthorized)
+				dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+					[[CameraRollHelper instance] selectPictureWithCompletionHandler:onComplete onFail:onFail];
+				});
+			else if (status == PHAuthorizationStatusRestricted)
+				onFail(@"PHAuthorizationStatusRestricted");
+			else if (status == PHAuthorizationStatusDenied)
+				onFail(@"PHAuthorizationStatusDenied");
 		@}
-
-
 	}
 }


Source/Fuse.CameraRoll/iOS/iOSCameraRoll.uno
```diff
@@ -41,11 +41,28 @@ namespace Fuse.CameraRoll
...
+					} else {
+						onFail(@"User cancelled");
+					}
+				}];
+			} else if (status == PHAuthorizationStatusAuthorized)
+			{
+				dispatch_async(dispatch_get_global_queue(DISPATCH_QUEUE_PRIORITY_DEFAULT, 0), ^{
+					[[CameraRollHelper instance] selectPictureWithCompletionHandler:onComplete onFail:onFail];
+				});
+			} else {
```
kusma: I know we don't have a defined ObjC [coding-style defined](https://github.com/fuse-open/fuselibs/blob/master/Documentation/CodingStyle.md), but this seems even self-inconsistent.

I *think* most of the ObjC code I've seen uses K&R style braces. This isn't important at all for this PR, but perhaps we should consider standardizing this somehow?

Source/Fuse.CameraRoll/iOS/iOSCameraRoll.uno
```diff
@@ -41,11 +41,28 @@ namespace Fuse.CameraRoll
...
+				onFail(@"User cancelled");
```
kusma: I don't think "User cancelled" is accurate here. Here's the full list of return-values, taken from [here](https://developer.apple.com/documentation/photokit/phauthorizationstatus?language=objc):
- `PHAuthorizationStatusNotDetermined`: Explicit user permission is required for photo library access, but the user has not yet granted or denied such permission.
- `PHAuthorizationStatusRestricted`: Your app is not authorized to access the photo library, and the user cannot grant such permission.
- `PHAuthorizationStatusDenied`: The user has explicitly denied your app access to the photo library.
- `PHAuthorizationStatusAuthorized`: The user has explicitly granted your app access to the photo library.

So, the first and last ones are handled, but the difference between `PHAuthorizationStatusRestricted` and `PHAuthorizationStatusDenied` seems relevant here.

Perhaps we should check for either, and pass the string from the documentation instead?

LuisRodriguezLD: Done. It now passes a string. Maybe we should put these values in the docs so Fusers can now what they mean.

PR #1209, opened 2018-08-13
New camera roll features

LuisRodriguezLD:
Sometimes automatically asking for access to photos may not be a good idea.
These new features allow the programmer to know the status and request access when the timing is appropriate.

More on that [here](https://uxplanet.org/getting-to-yes-best-practices-for-ios-permissions-dialogs-9d62892142cc).

### New
`checkPermissions()` - Checks if your app has access to the gallery.
`requestPermissions()` - Requests permission to access the gallery. In iOS, if the user has previously denied access this function will redirect them to your app settings in their iPhone / iPad.

This PR contains:
- [X] Changelog
- [X] Documentation
- [X] Tests

PR #1211, opened 2018-08-15
New camera features

LuisRodriguezLD:
Similar to #1209 

Sometimes automatically asking for access to the camera may not be a good idea.
These new features allow the programmer to know the status and request access when the timing is appropriate.

More on that [here](https://uxplanet.org/getting-to-yes-best-practices-for-ios-permissions-dialogs-9d62892142cc).

### New

`checkPermissions()` - Checks if your app has access to the camera.
`requestPermissions()` - Requests permission to access the camera. In iOS, if the user has previously denied access this function will redirect them to your app settings in their iPhone / iPad.

This PR contains:
- [X] Changelog
- [X] Documentation
- [X] Tests

### DIFF
--- a/Tests/ManualTests/ManualTestingApp/Singles/CameraPage.ux
+++ b/Tests/ManualTests/ManualTestingApp/Singles/CameraPage.ux
@@ -116,9 +116,9 @@
 			);
 		};
 
-    exports.checkPermissions = function()
+		exports.requestCameraPermissions = function()
 		{
-			CameraRoll.checkPermissions().then(
+			Camera.requestPermissions().then(
 				function(res)
 				{
 					console.log("response: "+res)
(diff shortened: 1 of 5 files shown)


Tests/ManualTests/ManualTestingApp/Singles/CameraPage.ux
```diff
@@ -102,7 +102,21 @@
...
+					console.log("response: "+res)
+				}
+			).catch(
+				function(error){
+					console.log("error: "+error)
+				}
+			);
+		};
+
+    exports.checkPermissions = function()
```
kusma: Hmm, seems something went wrong when I resolved the conflict here... Would you mind fixing it up? :pray:

PR #1212, opened 2018-08-20
Change version of play services of Maps

ckarmy:
Change the version of play services of Fuse.Maps to **12.0.1** To work and compile with the last commit of Fuse.Firebase(https://github.com/fuse-compound/Fuse.Firebase)

SOLVE this problem: https://stackoverflow.com/questions/41408514/com-google-android-gms-common-internal-safe-parcel-safe-parcelable-not-found

### DIFF
--- a/Source/Fuse.Maps/Android/MapView.uno
+++ b/Source/Fuse.Maps/Android/MapView.uno
@@ -14,7 +14,7 @@ namespace Fuse.Maps.Android
 		MOVE = 2
 	}
 
-	[Require("Gradle.Dependency.Compile", "com.google.android.gms:play-services-maps:9.2.0")]
+	[Require("Gradle.Dependency.Compile", "com.google.android.gms:play-services-maps:12.0.1")]
 	extern (Android) public class MapView : Fuse.Controls.Native.Android.LeafView, IMapView
 	{
 		public bool IsReady { get; private set; }


AndrewEQ:
Error Building Android App (Error with Google Play Services Maps) ~ https://github.com/fuse-open/fuselibs/issues/1216

kusma:
Hmmm...  Should I revert it, then?

AndrewEQ:
Well, it looks like the version of the "play-services-maps" is not installed? How does one make sure that fuse installs it? ...and is that version compatible with say API 22 and above? ~ https://developer.android.com/about/dashboards

kusma:
I don't think "fuse install android" is actively maintained at the moment...

AndrewEQ:
Yeah, about that, I'm still waiting for a proper hand off of all these "little" gotchyas, do you know where we can find and update that?

kusma:
I think the general consensus at Fuse before the pivot was that we wanted to remove "fuse install android" altogether, and refer to the android installer and documentation. The reason we did "fuse install android" was to make the on-boarding experience smooth. But since we got the preview app, the on-boarding experience didn't depend on "fuse install android" any more.

I'm not saying "fuse install android" doesn't have value; just that the effort it takes to keep this up-to-date might not be worth it any more. In either case we need to document some more; either how to do it "by hand", or how to update "fuse install android".

I've saved an article from the internal wiki about this [here](https://gist.github.com/kusma/7030a33ee5f7044dde38bbbfd317e64d). Perhaps it would be helpful.

kusma:
FWIW, #1197 tries to upgrade this dependency even further...

AndrewEQ:
Ah see, thanks @kusma, we just tested this, @ckarmy needs to add this above the services require, so if a fuser didn't previously update their libs, it would auto install the dependency upon using ```fuse build```. 

I believe first-time users would benefit from "fuse install android" in terms of its "magic" but the more responsible way would be to educate them or give them the "30 second install" (which is listing all the necessary commands to get up and running with one-liners explaining what its going to do)

```[Require("Gradle.Repository", "maven { url 'https://maven.google.com' }")]```

AndrewEQ:
#1197 sounds about right, I think the phone will ask the user to update their Google Play Services or it will auto-update via the Play Store ~ https://developers.google.com/android/guides/overview

Although, if the version is outdated, we should popup a dialog to the user to update via the Play Store, where to put this check or checks is upto you guys: 

Ensure Devices Have the Google Play services APK ~ https://developers.google.com/android/guides/setup

PR #1214, opened 2018-08-30
Added "Tap to Focus" to CameraView for iOS and Android

AndrewEQ:
- added setCameraFocusPoint
- added to example usage

:cake:

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.CameraView/Android/Camera.uno
+++ b/Source/Fuse.Controls.CameraView/Android/Camera.uno
@@ -155,6 +155,32 @@ namespace Fuse.Controls.Android
 			return new AndroidPhotoOptionPromise(this).Visit(options);
 		}
 
+		class AndroidCameraFocusPointPromise : Promise<CameraFocusPoint> {
+			Camera _camera;
+			string _parameters;
+
+			public AndroidCameraFocusPointPromise(Camera camera) {
+				_camera = camera;
+				_parameters = _camera.SaveParameters();
+			}
+				
+			public Future<CameraFocusPoint> doFocus(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) {
+				try {
+					_camera.SetCameraFocusPointNow(x,y, cameraWidth, cameraHeight, isFocusLocked);
+					
+					Resolve((CameraFocusPoint)1);
+					return this;
+				} catch (Exception e) {
+					Reject(e);
+					return this;
+				}
+			}
+		}
+
+		public Future<CameraFocusPoint> SetCameraFocusPoint(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) {
+			return new AndroidCameraFocusPointPromise(this).doFocus(x, y, cameraWidth, cameraHeight, isFocusLocked);
+		}
+
 		class RecordingSessionPromise : Promise<RecordingSession>
 		{
 			Action _doneCallback;
@@ -295,6 +321,11 @@ namespace Fuse.Controls.Android
 		void SetFlashMode(Java.Object handle, string flashMode)
 		@{
 			((CameraImpl)@{Fuse.Controls.Native.ViewHandle:Of(_this).NativeHandle:Get()}).setFlashMode(flashMode);
+		@}
+
+		[Foreign(Language.Java)]
+		void SetCameraFocusPointNow(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) @{
+			((CameraImpl)@{Fuse.Controls.Native.ViewHandle:Of(_this).NativeHandle:Get()}).setCameraFocusPoint(x,y,cameraWidth,cameraHeight, isFocusLocked);
 		@}
 
 		[Foreign(Language.Java)]
--- a/Source/Fuse.Controls.CameraView/Android/Camera.uno
+++ b/Source/Fuse.Controls.CameraView/Android/Camera.uno
@@ -155,29 +155,36 @@ namespace Fuse.Controls.Android
 			return new AndroidPhotoOptionPromise(this).Visit(options);
 		}
 
-		class AndroidCameraFocusPointPromise : Promise<CameraFocusPoint> {
+		class AndroidCameraFocusPointPromise : Promise<CameraFocusPoint> 
+		{
 			Camera _camera;
 			string _parameters;
 
-			public AndroidCameraFocusPointPromise(Camera camera) {
+			public AndroidCameraFocusPointPromise(Camera camera) 
+			{
 				_camera = camera;
 				_parameters = _camera.SaveParameters();
 			}
 				
-			public Future<CameraFocusPoint> doFocus(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) {
-				try {
+			public Future<CameraFocusPoint> doFocus(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
+			{
+				try 
+				{
 					_camera.SetCameraFocusPointNow(x,y, cameraWidth, cameraHeight, isFocusLocked);
 					
 					Resolve((CameraFocusPoint)1);
 					return this;
-				} catch (Exception e) {
+				} 
+				catch (Exception e) 
+				{
 					Reject(e);
 					return this;
 				}
 			}
 		}
 
-		public Future<CameraFocusPoint> SetCameraFocusPoint(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) {
+		public Future<CameraFocusPoint> SetCameraFocusPoint(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
+		{
 			return new AndroidCameraFocusPointPromise(this).doFocus(x, y, cameraWidth, cameraHeight, isFocusLocked);
 		}
 
@@ -324,7 +331,8 @@ namespace Fuse.Controls.Android
 		@}
 
 		[Foreign(Language.Java)]
-		void SetCameraFocusPointNow(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) @{
+		void SetCameraFocusPointNow(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
+		@{
 			((CameraImpl)@{Fuse.Controls.Native.ViewHandle:Of(_this).NativeHandle:Get()}).setCameraFocusPoint(x,y,cameraWidth,cameraHeight, isFocusLocked);
 		@}
 
--- a/Source/Fuse.Controls.CameraView/Android/Camera.uno
+++ b/Source/Fuse.Controls.CameraView/Android/Camera.uno
@@ -155,7 +155,7 @@ namespace Fuse.Controls.Android
 			return new AndroidPhotoOptionPromise(this).Visit(options);
 		}
 
-		class AndroidCameraFocusPointPromise : Promise<CameraFocusPoint> 
+		class AndroidCameraFocusPointPromise : Promise<Nothing> 
 		{
 			Camera _camera;
 			string _parameters;
@@ -166,13 +166,13 @@ namespace Fuse.Controls.Android
 				_parameters = _camera.SaveParameters();
 			}
 				
-			public Future<CameraFocusPoint> doFocus(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
+			public Future<Nothing> doFocus(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
 			{
 				try 
 				{
 					_camera.SetCameraFocusPointNow(x,y, cameraWidth, cameraHeight, isFocusLocked);
 					
-					Resolve((CameraFocusPoint)1);
+					Resolve(default(Nothing));
 					return this;
 				} 
 				catch (Exception e) 
@@ -183,7 +183,7 @@ namespace Fuse.Controls.Android
 			}
 		}
 
-		public Future<CameraFocusPoint> SetCameraFocusPoint(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
+		public Future<Nothing> SetCameraFocusPoint(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) 
 		{
 			return new AndroidCameraFocusPointPromise(this).doFocus(x, y, cameraWidth, cameraHeight, isFocusLocked);
 		}
--- a/Source/Fuse.Controls.CameraView/Android/Camera.uno
+++ b/Source/Fuse.Controls.CameraView/Android/Camera.uno
@@ -90,6 +90,7 @@ namespace Fuse.Controls.Android
 		public Java.Object CameraHandle { get; private set; }
 		public CameraFacing Facing { get; private set; }
 
+		
 		FlashMode _flashMode = FlashMode.Off;
 		public FlashMode FlashMode
 		{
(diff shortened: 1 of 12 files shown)


Source/Fuse.Controls.CameraView/iOS/CameraViewImpl.mm
```diff
@@ -400,6 +401,116 @@ void setCameraFacing(CameraView* cameraView, int cameraFacing, void(^onResolve)(
...
+			    if ([currentDevice isFocusPointOfInterestSupported]) {
+			    	[currentDevice setFocusPointOfInterest:focusPoint];	
+			    }
+					
+					if ([currentDevice isExposurePointOfInterestSupported]) {
+				    [currentDevice setExposurePointOfInterest:focusPoint];
+				  }
+
+					[currentDevice setFocusMode:AVCaptureFocusModeContinuousAutoFocus];
+			    [currentDevice setExposureMode:AVCaptureExposureModeContinuousAutoExposure];
```
kusma: Seems something went wrong with the indentation here... In general, this commit seems to introduce a lot of whitespace issues in this file... It seems this is due to inconsistent indentation style. Please fix that.

Just a tip: I would suggest coding with visible white-space so you can see what's going on...

Source/Fuse.Controls.CameraView/iOS/CameraViewImpl.mm
```diff
@@ -400,6 +401,116 @@ void setCameraFacing(CameraView* cameraView, int cameraFacing, void(^onResolve)(
...
+			if([currentDevice isFocusPointOfInterestSupported] 
+				&& [currentDevice isFocusModeSupported:AVCaptureFocusModeContinuousAutoFocus]
+				&& [currentDevice isExposureModeSupported:AVCaptureExposureModeContinuousAutoExposure]
+				)
+			{
+				NSError* error;
+
+				//check to unlock
+				if (isFocusLocked != 1 && [currentDevice lockForConfiguration:&error])
+				{
```
kusma: The other code in this file seems to use Egyptian braces, please be consistent with that. This applies elsewhere in this patch as well.

Source/Fuse.Controls.CameraView/Android/Camera.uno
```diff
@@ -90,6 +90,7 @@ namespace Fuse.Controls.Android
 		public Java.Object CameraHandle { get; private set; }
 		public CameraFacing Facing { get; private set; }
 
+
```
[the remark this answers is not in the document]
AndrewEQ: My bad, will remove those...

Source/Fuse.Controls.CameraView/iOS/CameraViewImpl.mm
AndrewEQ: Yeah I dunno how that happened, looks fine on my side, will go through each line...

Source/Fuse.Controls.CameraView/iOS/CameraViewImpl.mm
AndrewEQ: Yeah, was a little confused between languages, one was using egyptian, the other wasn't, I personally like egyptian, so am gonna use it all round...

AndrewEQ:
Ok, made the changes!

Source/Fuse.Controls.CameraView/Android/Camera.uno
```diff
@@ -155,6 +155,32 @@ namespace Fuse.Controls.Android
 			return new AndroidPhotoOptionPromise(this).Visit(options);
 		}
 
+		class AndroidCameraFocusPointPromise : Promise<CameraFocusPoint> {
```
kusma: We don't use egyptian braces in Uno, as per [the coding style](https://github.com/fuse-open/fuselibs/blob/master/Documentation/CodingStyle.md). Please fix this elsewhere in this PR also.

Source/Fuse.Controls.CameraView/Android/Camera.uno
```diff
@@ -155,6 +155,32 @@ namespace Fuse.Controls.Android
...
+			public AndroidCameraFocusPointPromise(Camera camera) {
+				_camera = camera;
+				_parameters = _camera.SaveParameters();
+			}
+				
+			public Future<CameraFocusPoint> doFocus(double x, double y, int cameraWidth, int cameraHeight, int isFocusLocked) {
+				try {
+					_camera.SetCameraFocusPointNow(x,y, cameraWidth, cameraHeight, isFocusLocked);
+					
+					Resolve((CameraFocusPoint)1);
```
kusma: This seems to cast 1 to a non-existent enum-value, which doesn't make a whole lot of sense to me. How about adding a new corresponding enum value and using that instead? Then some semantics would be attached to this, instead of a magical "1"...

kusma: Actually, shouldn't this just be `Resolve(CameraFocusPoint.OK)`? This looks like a success-path...

Source/Fuse.Controls.CameraView/CameraViewBase.uno
```diff
@@ -33,6 +33,10 @@ namespace Fuse.Controls
 		Front,
 	}
 
+	public enum CameraFocusPoint {
+		Ok,
+	}
```
kusma: An enum with a single value seems nonsensical to me... What's the point of this?

Source/Fuse.Controls.CameraView/iOS/CameraViewImpl.h
```diff
@@ -29,6 +29,10 @@ namespace fcv {
 		CAMERA_FACING_BACK = 1,
 	};
 
+	enum CameraFocusPoint {
+		OK = 1,
+	};
```
kusma: Again, the intention of this is kinda unclear... Also, why does this have a value of 1, as opposed to the Uno-definition?

Source/Fuse.Controls.CameraView/Android/Camera.uno
AndrewEQ: Yeah, I think initially we were gonna return an x and y point but it wasn't necessary, so I guess we can remove the enums?

kusma: Sounds good

Source/Fuse.Controls.CameraView/CameraViewBase.uno
AndrewEQ: Looking over the code again, I didn't understand how to write in uno that the value returned for a function could be an int if successful or a string if error, so I think that's why I created the enum to promise "something"

Source/Fuse.Controls.CameraView/iOS/CameraViewImpl.h
AndrewEQ: Yeah, its not necessary for the iOS and android implementations, so I'll remove em

Source/Fuse.Controls.CameraView/Android/Camera.uno
AndrewEQ: Aight, I think I'm getting the styles now...

AndrewEQ:
Ok, made the updates!

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                    case 0: //portrait
+                        rect = new Rect(left, top, left + focus_area_size, top + focus_area_size);
+                        break;
+                    case 3: //landscape right
+                        rect = new Rect(top, left, top + focus_area_size, left + focus_area_size);
+                        break;
+                    case 2: //portrait upsidedown
+                        rect = new Rect(top, left, top + focus_area_size, left + focus_area_size);
+                        break;
+                    case 1: //landscape left
```
kusma: maybe use `Surface.ROTATION_0`, `Surface.ROTATION_270`, `Surface.ROTATION_180` and `Surface.ROTATION_90` instead of 0, 3, 2, 1? And perhaps sort the cases also?

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                        rect = new Rect(left, top, left + focus_area_size, top + focus_area_size);
+                        break;
+                    default:
+                        rect = new Rect(left, top, left + focus_area_size, top + focus_area_size);
```
kusma: Is this actually needed, or are you just playing safe here? If the latter, I'd just throw an exception or something instead. If not, I'd reuse the `case 0`-code to avoid repetition...

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                        break;
+                }
+
+
+                List<Camera.Area> meteringAreas = new ArrayList<Camera.Area>();
+                meteringAreas.add(new Camera.Area(rect, 1000));
```
kusma: So, 1000 is the max weight... Does that matter? If it's just a value to please the API, I'd pass 0...

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+
+                        parameters.setAutoWhiteBalanceLock(false);
+                    }
+
+                    _camera.setParameters(parameters);
+                }
+
+                int focus_area_size = 300;
+                int left = clamp(Float.valueOf((float)(x / cameraWidth) * 2000 - 1000).intValue(), focus_area_size);
+                int top = clamp(Float.valueOf((float)(y / cameraHeight) * 2000 - 1000).intValue(), focus_area_size);
```
kusma: This whole calculation seems a bit strange. Would you mind adding some comments explaining what it does?

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                }
+
+                if (parameters.getMinExposureCompensation() != 0 
+                    && parameters.getMaxExposureCompensation() != 0
+                    && parameters.isAutoExposureLockSupported()
+                ) {
+                    
+                    int exposureAmount = (parameters.getMinExposureCompensation() - parameters.getMaxExposureCompensation()) * -1;
+                    int designAdjustment = 1; //added 1 to account for a control panel covered area on the bottom of the camera that can't be tapped on for focus
+                    if ((y/cameraHeight) < 0.5) { //apply only to upper half of camera screen
```
kusma: I think `if (y < cameraHeight / 2)` would be clearer....

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                        designAdjustment = 0;
+                    }
+
+                    int amountApplied = (int)Math.round( (exposureAmount * (y/cameraHeight)) ) + designAdjustment; 
+                    amountApplied = (amountApplied > exposureAmount) ? exposureAmount : amountApplied;
```
kusma: Is this code based on some algorithm worth noting? I'm not really sure what's going on here... Some more comments would be nice.

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                    parameters.setExposureCompensation(parameters.getMinExposureCompensation() + amountApplied);
+                    
+                    parameters.setAutoExposureLock(false);
+                }
+
+
+                _camera.setParameters(parameters);
+
+
+
```
kusma: I'm not sure what's up with the excessive amounts of blank lines here...

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+                        parameters.setAutoWhiteBalanceLock(true);
+                    }
+
+                    _camera.setParameters(parameters);
+
+
+                    _autoFocus = false;
+                } else {
+                    _autoFocus = false;
+                }
```
kusma: Here you can lift out the `_autoFocus = false;` assignment, because it happens in both blocks, and get rid of the else.

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+        int result;
+        if (Math.abs(touchCoordinateInCameraReper)+focusAreaSize/2>1000) {
+            if (touchCoordinateInCameraReper>0) {
+                result = 1000 - focusAreaSize/2;
+            } else {
+                result = -1000 + focusAreaSize/2;
+            }
+        } else {
+            result = touchCoordinateInCameraReper - focusAreaSize/2;
+        }
```
kusma: I'm not sure what's going on here... care to explain?

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -253,6 +257,164 @@ public void setFlashMode(String flashMode) {
...
+        return result;
```
kusma: You don't need result. Just return when you assign to it instead, should do the same thing.

Source/Fuse.Controls.CameraView/CameraViewBase.ScriptClass.uno
```diff
@@ -141,6 +142,28 @@ namespace Fuse.Controls
...
+						.then(function(isSet) {  })
+						.catch(function(error) { });
+				</JavaScript>
+		*/
+		static Future<CameraFocusPoint> setCameraFocusPoint(Context context, CameraViewBase self, object[] args) 
+		{
+			if (args.Length != 5)
+				return new Promise<CameraFocusPoint>().RejectWithMessage("Arguments for CameraFocusPoint must be provided");
+
+			return self.SetCameraFocusPoint( Marshal.ToDouble(args[0]), Marshal.ToDouble(args[1]), Marshal.ToInt(args[2]), Marshal.ToInt(args[3]), Marshal.ToInt(args[4]) );
```
kusma: I would wrap this line so it'd be easy to see all arguments even in githubs narrow view...

Source/Fuse.Controls.CameraView/Example/CameraViewExample/.unoconfig_example
```diff
@@ -0,0 +1 @@
+Packages.SourcePaths += /www/github/fuselibs/Source
```
kusma: This file looks like a mistake. I doubt that path is useful on other people's machines ;)

AndrewEQ: that's why its an example :D

Source/Fuse.Controls.CameraView/CameraViewBase.uno
```diff
@@ -33,6 +33,11 @@ namespace Fuse.Controls
 		Front,
 	}
 
+	public enum CameraFocusPoint 
+	{
+		Ok,
+	}
```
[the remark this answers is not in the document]
AndrewEQ: I'm not an UNO expert yet, I tried to do Fuse.Scripting.Object but I couldn't get the typing to work; I had casting issues everywhere, you can transplant it afterwards if you like but I really don't know how to make it work.

Source/Fuse.Controls.CameraView/Example/CameraViewExample/.unoconfig_example
AndrewEQ: ...plus its great for testing the example with latest source

AndrewEQ:
Ok, made the updates!

Source/Fuse.Controls.CameraView/CameraViewBase.uno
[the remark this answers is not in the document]
kusma: See what Fuse.FileSystem does...

Source/Fuse.Controls.CameraView/Example/CameraViewExample/.unoconfig_example
kusma: I don't understand. This path is only useful on your computer. Put it somewhere else, and you should be fine without adding noise like this...

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
```diff
@@ -8,21 +8,27 @@
...
-    final boolean _autoFocus;
+    boolean _autoFocus;
     final int _maxWidth;
     final int _maxHeight;
     final int _cameraId;
     final OrientationEventListener _orientationListener;
 
     int _cameraRotation = 0;
+    int previewWidth;
+    int previewHeight;
```
kusma: Private fields start with an underscore, according to the coding standard...

Source/Fuse.Controls.CameraView/Example/CameraViewExample/.unoconfig_example
AndrewEQ: Its useful for the testing of new source workflow, which you guys haven't mentioned; after pulling a change to the libs or making a change, the project doesn't include the updated source files without using the .unoconfig file and "uno doctor" command on a project. The fuse libs work the first time when following the instructions but not again thereafter. This is for mac: https://forums.fusetools.com/t/run-fuselibs-locally/7250

Source/Fuse.Controls.CameraView/Android/CameraImpl.java
AndrewEQ: Oh hey, I just realised, that those are not my variables!

AndrewEQ: ...I shall change them for you while I'm at it ;)

AndrewEQ:
Hokay updates made!

PR #1221, opened 2018-09-27
Fuse.Maps Fix

AndrewEQ:
Need the google repo to download the maps play services

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Maps/Android/MapView.uno
+++ b/Source/Fuse.Maps/Android/MapView.uno
@@ -14,6 +14,7 @@ namespace Fuse.Maps.Android
 		MOVE = 2
 	}
 
+	[Require("Gradle.Repository", "maven { url 'https://maven.google.com' }")]
 	[Require("Gradle.Dependency.Compile", "com.google.android.gms:play-services-maps:12.0.1")]
 	extern (Android) public class MapView : Fuse.Controls.Native.Android.LeafView, IMapView
 	{


kusma:
8 commits (half of them merges) for this little amount of code-changes doesn't seem quite right. Could you at least rebase this on master so the history gets somewhat understandable?

Source/Fuse.Maps/Android/MapView.uno
```diff
@@ -14,6 +14,7 @@ namespace Fuse.Maps.Android
 		MOVE = 2
 	}
 
+	[Require("Gradle.Repository", "maven { url 'https://maven.google.com' }")]
 	[Require("Gradle.Dependency.Compile", "com.google.android.gms:play-services-maps:12.0.1")]
```
kusma: Why exactly is Gradle.Repository needed? We use the same pattern in other places (Fuse.Elements, Fuse.Maps, Fuse.PushNotifications, Fuse.Scripting.JavaScript...) without specifying Gradle.Repository...  In fact, lib/UnoCore/Targets/Android/Dependencies.uxl in Uno already specify this, which AFAIK means that it gets required for everything...

What behavior does this PR change?

kusma: Uno had that "Gradle.Repository"-stuff since before last release, so that explanation doesn't satisfy me. This should AFAICT work without it. What problem are you seeing?

kusma: No, you're right (I deleted my comment, I thought you hadn't read it yet), the old version doesn't have this. Because old versions didn't need it.

I think we should just upgrade uno and take it from there. Peppering repeated requirements all over the code-base isn't going to be fun to maintain.

AndrewEQ: I don't think gradle but the ndk is impacted: https://developer.android.com/ndk/guides/cpp-support.html

AndrewEQ: I was wondering, how do we keep older builds compatible and working as they still make up the majority of the devices out there: https://developer.android.com/about/dashboards

AndrewEQ: Android 8+ is only 19%

kusma: I don't see anything obvious at that link hinting at compatibility issues. What am I missing?

AndrewEQ: Oh sorry, it looks like they updated their own docs, never mind that.

AndrewEQ: When I was testing on my lower end Marshmallow, after compiling with new uno, gradle, ndk, etc. the app could build, install but not run. I noticed that in the latest ndk the toolchains folder is a lot less, could it be that?

AndrewEQ: Oh right, this is the error I was getting when compiling with latest NDK on fuse 1.9: https://github.com/fuse-open/uno/issues/109

This was after I updated my android to latest everything.

AndrewEQ: "We're not going to keep compatibility with older Uno versions anyway, so this is only interesting for people who carry their own patches to old releases, I think. And they can carry this patch as well, no?" I think its interesting for all the devs that want to support a bigger market share. From KitKat to Nougat its 77%, with Oreo, that makes 96.2%!

kusma: Look, I think you're interpreting this a bit wrong. We're not going to drop compatibility with old devices, at least not intentionally. So I suppose that means fuse-open/uno#109 needs to be fixed before the next release.

What I'm talking about is requiring new *Uno versions* for new fuselibs versions. Uno and Fuselibs are versioned together and released together, and there's no plans of changing this.

AndrewEQ: Ah, my bad, yeah, its like we need 2 branches for both of them, "legacy" and "latest"?

Do you think we could fix push notifications before the next release? (https://developers.google.com/cloud-messaging/android/android-migrate-fcm)

kusma: No. I don't have any plans of maintaining old versions, at least. The goal is to keep compatibility, as far as I'm concerned. Perhaps we'll need to hide it behind some switch to please the app store, but if so then that's fine by me.

I think @cbaggers is working on trying to get that merged, no?

AndrewEQ: Hehe, I don't think anyone wants to maintain old versions lol, so I'm with you on that, just needs to run at a minimum until they have totally been made obsolete. Yeah, I think a switch like "--legacy=true" to compile with latest older libs, etc. would be great.
 
Last he said was him workin on updating his firebase version of push notifications, not sure if that meant integrating it with/replacing the android version of fuselibs.

kusma: @mortend Said in fuse-open/uno#109 that he tested on older android versions and things work fine. So it doesn't look like we regressed here in the first place to me...

AndrewEQ: So does fuse 1.9 compile fine with latest fuselibs maps?

mortend: The maven repository will be added in Uno 1.10 which is not yet released, and fuselibs is currently using Uno 1.9.

If latest fuselibs maps don't work without this patch - maybe we should merge this patch while waiting for Uno 1.10, and revert it again once 1.10 is in place...

Just to clarify on Android versions: Uno 1.10 upgrades and conforms to latest Android SDK and NDK, but this does not imply breaking builds for older devices. The minimum Android version required to run Fuse apps is Android 4.1 (API level 16), same as before.

AndrewEQ: @mortend cool, will test on my marshmallows when 1.10 is out, an idea on an eta? 
@cbaggers does/will 1.10 include a move from GCM to FCM in terms of push notifications? :D

mortend:
#1239 upgrades to Uno 1.10.0-rc1 which makes this PR redundant since the google repo is now already included. Closing this. :)

AndrewEQ:
yeah, no sweat :)

PR #1222, opened 2018-09-29
Fix iOS DatePicker in WhileTrue bug

cbaggers:
The actual issue here was simple. The Min & Max were set to the Value,
which meant there were no valid dates so iOS greyed them out.

There is probably a secondary issue of why this issue didn't manifest without
the inclusion of the WhileTrue, but that is for a separate PR.

This fixes:
- https://github.com/fuse-open/fuselibs/issues/1218
- https://forums.fusetools.com/t/datepicker-wont-allow-date-change-when-inside-whiletrue-on-ios/7295

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+### Fuse.Controls.DatePicker
+- Fixed bug which caused dates not to be selectable if the control was inside a WhileTrue on iOS
+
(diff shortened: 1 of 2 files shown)


cbaggers:
This change also matches the code in the android impl so I'm pretty confident about it being how it was meant to be.

PR #1224, opened 2018-09-29
fix invalid json in local notif payload on android

cbaggers:
Replace the old hacky way of making the result string with use of
JSONObject.

I'm not stoked with putting the helper function in
LocalNotificationReceiver however it saves adding extra java files to
the build for now when we have something that already works.

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.LocalNotifications/Android/Impl.uno
+++ b/Source/Fuse.LocalNotifications/Android/Impl.uno
@@ -68,7 +68,7 @@ namespace Fuse.LocalNotifications
                         String title = newIntent.getStringExtra("title");
                         String body = newIntent.getStringExtra("bbody");
                         String payload = newIntent.getStringExtra(@{ACTION});
-                        String result = "{ 'title': '" + title + "', 'body': '" + body + "', 'payload': '" + payload + "' }";
+						String result = com.fuse.LocalNotifications.LocalNotificationReceiver.MakePayloadString(title, body, payload);
                         @{NotificationRecieved(string):Call(result)};
                     }
                 },
@@ -128,7 +128,7 @@ namespace Fuse.LocalNotifications
 
             if (com.fuse.LocalNotifications.LocalNotificationReceiver.InForeground)
             {
-                String result = "{ 'title': '" + title + "', 'body': '" + body + "', 'payload': '" + payload + "' }";
+				String result = com.fuse.LocalNotifications.LocalNotificationReceiver.MakePayloadString(title, body, payload);
                 @{NotificationRecieved(string):Call(result)};
             } else {
                 Intent notificationIntent = new Intent(context, @(Activity.Package).@(Activity.Name).class);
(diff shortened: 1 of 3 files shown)


Source/Fuse.LocalNotifications/Android/Impl.uno
```diff
@@ -68,7 +68,7 @@ namespace Fuse.LocalNotifications
                         String title = newIntent.getStringExtra("title");
                         String body = newIntent.getStringExtra("bbody");
                         String payload = newIntent.getStringExtra(@{ACTION});
-                        String result = "{ 'title': '" + title + "', 'body': '" + body + "', 'payload': '" + payload + "' }";
+						String result = com.fuse.LocalNotifications.LocalNotificationReceiver.MakePayloadString(title, body, payload);
```
kusma: You're mixing indentation styles here. While you're using the one that the code-style mandates, This just gets hard to read. Could you either keep spaces, or start the PR with a commit that changes the indentation style for the whole file?

Source/Fuse.LocalNotifications/Android/Impl.uno
```diff
@@ -128,7 +128,7 @@ namespace Fuse.LocalNotifications
 
             if (com.fuse.LocalNotifications.LocalNotificationReceiver.InForeground)
             {
-                String result = "{ 'title': '" + title + "', 'body': '" + body + "', 'payload': '" + payload + "' }";
+				String result = com.fuse.LocalNotifications.LocalNotificationReceiver.MakePayloadString(title, body, payload);
```
kusma: Same here

Source/Fuse.LocalNotifications/Android/LocalNotificationReceiver.java
```diff
@@ -12,4 +13,19 @@ public void onReceive(Context context, Intent intent)
...
+	{
+		JSONObject jsonObj = new JSONObject();
+		try
+		{
+			jsonObj.put("title", (title == null ? "" : title));
+			jsonObj.put("body", (body == null ? "" : body));
+			jsonObj.put("payload", (payload == null ? "" : payload));
+		}
+		catch (Exception e)
+		{
```
kusma: Empty catch-block... really? If that's a sane thing to do, maybe add a comment to explain why?

Source/Fuse.LocalNotifications/Android/Impl.uno
cbaggers: I didnt do that because I remembered your hatred of whitespace change commits. Happy to do that though

Source/Fuse.LocalNotifications/Android/Impl.uno
cbaggers: same

Source/Fuse.LocalNotifications/Android/Impl.uno
cbaggers: I'll just make it conform with current style

Source/Fuse.LocalNotifications/Android/LocalNotificationReceiver.java
kusma: If an exception triggers, the body won't run fully, leaving us with a half-filled in jsonObj, no? That sounds bad to me...

kusma:
It's been almost a year without feedback from the author, I'm closing this.

@cbaggers: If you still want to pursue this, please reopen and address the feedback!

PR #1242, opened 2018-10-22
Fuse.Controls.CameraView: fix project references

mortend:
We should use project references when referencing projects from the same
repository. This avoids getting errors when building, such as:

    E0000: Multiple projects or packages with the name Fuse.ImageTools

The errors starts happening when building multiple times, when both the
original project and a built package version of it exist. After fixing the
project references, we can build multiple times without getting errors.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
```
Source/Fuse.Controls.CameraView/iOS/NativePhoto.uno(362): E3105: 'Fuse' does not contain type or namespace 'App'. Could you be missing a package reference?
```
Seems there's some reference missing?

mortend:
Whoops, it's fixed now.

PR #1245, opened 2018-11-06
Added IsLocationEnabled + Permission check for StartListening(Android)

AndrewEQ:
UX usage: GeoLocation.isLocationEnabled();

- iOS - Added isLocationEnabled - returns boolean
- Android - Added isLocationEnabled - returns boolean
- Android - Added permission check and request before getting location for StartListening

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


kusma:
So, it seems nobody is willing to review this one? Sadly, this is outside of my area of expertise, but unless someone points out something wrong, I could review the "technical" parts of the code and just assume the stuff I don't know is fine so we can move forwards?

@AndrewEQ: If that's OK with you, would you mind rebasing this on master? I don't think it would be a problem, I just want to see that things still build on a newer version :)

AndrewEQ:
@kusma sure thing 👍

Source/Fuse.GeoLocation/Android/AndroidLocationProvider.uno
```diff
@@ -173,17 +209,44 @@ namespace Fuse.GeoLocation
...
+				var permissions = new PlatformPermission[] 
+				{
+					Permissions.Android.INTERNET,
+					Permissions.Android.ACCESS_COARSE_LOCATION,
+					Permissions.Android.ACCESS_FINE_LOCATION
+				};
 				
-				if(IsNetworkEnabled(_locationManager))
-					RequestNetworkLocationUpdates(_locationManager, minimumReportInterval, desiredAccuracyInMeters, _updateListener);
+				Permissions.Request(permissions).Then(OnPermissionsResult, OnPermissionsError);
```
kusma: Why is this needed again? It's already done in Init?

AndrewEQ: Its because the user can disable geo-location outside of the app, after initially saying yes.

kusma: Right, makes sense. But perhaps this should be a helper-method, so we don't have to repeat ourselves, then?

AndrewEQ: There is a helper method: IsLocationEnabled()

Which is being used here to determine if it is enabled, you gotta do it on every "GPS call" as the user can turn off at anytime.

kusma: That's a different helper than what I'm proposing. I'm talking about these lines:
```java
var permissions = new PlatformPermission[] 
{
    Permissions.Android.INTERNET,
    Permissions.Android.ACCESS_COARSE_LOCATION,
    Permissions.Android.ACCESS_FINE_LOCATION
};
Permissions.Request(permissions).Then(OnPermissionsResult, OnPermissionsError);
```
...as these are exactly the same, and if we need to change the list of permissions we now have to update two locations.

AndrewEQ: oh I see, my bad, I actually can't think how to make that dynamic right now, ideas? 😄

mortend: Add the following helper method.

```
		void RequestPermissions()
		{
			var permissions = new PlatformPermission[] 
			{
				Permissions.Android.INTERNET,
				Permissions.Android.ACCESS_COARSE_LOCATION,
				Permissions.Android.ACCESS_FINE_LOCATION
			};
			
			Permissions.Request(permissions).Then(OnPermissionsResult, OnPermissionsError);
		}
```

... then replace the two identical permission-snippets in this class with calls to your new `RequestPermissions()` helper. :)

AndrewEQ: Oh I see, I thought @kusma was talking about having those permissions editable in like a .uxl or summin, will make the change soon...

kusma: Yup, that's exactly what I meant! Sorry if I was unclear :)

Source/Fuse.GeoLocation/iOS/IOSLocationProvider.uno
```diff
@@ -109,6 +109,33 @@ namespace Fuse.GeoLocation
...
+						break;
+					case kCLAuthorizationStatusAuthorizedAlways: return true;
+						break;
+					case kCLAuthorizationStatusAuthorizedWhenInUse: return true;
+						break;
+					case kCLAuthorizationStatusNotDetermined: return false;
+						break;
+					default: return false;
+						break;
+				}
```
kusma: You don't need breaks after the return-statements.
But, I think this whole check would be cleaner if written like this:
```java
switch ([CLLocationManager authorizationStatus]) {
case kCLAuthorizationStatusAuthorizedAlways:
case kCLAuthorizationStatusAuthorizedWhenInUse:
        return true;

default:
        return false;
}
```

Point is, you only need to distinguish between the true and false case, and since the default-case already covers false, we can get rid of all other false-cases. Then finally we can merge the two true-cases.

Source/Fuse.GeoLocation/LocationProvider.uno
```diff
@@ -83,6 +85,10 @@ namespace Fuse.GeoLocation
 			}
 		}
 
+		public bool IsLocationEnabled() {
```
kusma: style: please put the opening curly on its own line

Source/Fuse.GeoLocation/iOS/IOSLocationProvider.uno
AndrewEQ: I guess I was being a little verbose, just thought the next guy to read the cases could see the reasons but I guess its not important

Source/Fuse.GeoLocation/LocationProvider.uno
AndrewEQ: will do...

AndrewEQ:
Sure thing, this is improving my squashing skills 😄

Source/Fuse.GeoLocation/Fuse.GeoLocation.unoproj
```diff
@@ -2,6 +2,9 @@
   "Copyright": "Copyright (c) Fuse Open 2018",
   "Description": "Native Location",
   "Publisher": "Fuse Open",
+  "Package": {
```
mortend: One more thing: I removed the `Package` properties from project files earlier today, but your PR resurrects the property in this project file. Can you remove these three lines (basically revert `Fuse.GeoLocation.unoproj`) while at it?

AndrewEQ: Sure, why not...

kusma:
Thanks a lot for bearing with us here :)

PR #1261, opened 2019-04-09
Add TypeScript Definitions

mortend:
This adds a Definitions file for TypeScript, helpful for code-completion and as documentation.

This file is automatically picked up by TypeScript editors when Fuselibs is installed to `node_modules/` via `npm install` or `npm link`. Tested in Visual Studio Code.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -84,7 +84,6 @@ declare module "FuseJS/Base64" {
             console.log(Base64.encodeUtf8("Foo � bar")); //LOG: Rm9vIMKpIGJhcg==
     */
     function encodeUtf8(value: string): string;
-
 }
 
 /**
@@ -171,7 +170,6 @@ declare module "FuseJS/Bundle" {
         > Warning: This call will block until the operation is finished. If you are reading large amounts of data, use read() instead.
     */
     function readSync(filename: string): string;
-
 }
 
 /**
@@ -334,7 +332,6 @@ declare module "FuseJS/Lifecycle" {
         * `"stateChanged"` - Triggered when the app's lifecycle state has changed.
     */
     function on(event: Event, callback: () => void): void;
-
 }
 
 /**
@@ -380,7 +377,6 @@ declare module "FuseJS/Camera" {
         The image capture view is user-configurable on Android.
     */
     function takePicture(desiredWidth?: number, desiredHeight?: number): Promise<any>;
-
 }
 
 /**
@@ -450,7 +446,6 @@ declare module "FuseJS/CameraRoll" {
         Starts an OS-specific image picker view (user-configurable on Android).
     */
     function getImage(): Promise<any>;
-
 }
 
 /**
@@ -974,7 +969,6 @@ declare module "FuseJS/FileSystem" {
             FileSystem.writeTextToFileSync("myfile.txt", "Hello buddy");
     */
     function writeTextToFileSync(filename: string, text: string): void;
-
 }
 
 /**
@@ -1164,7 +1158,6 @@ declare module "FuseJS/GeoLocation" {
         See [the GeoLocation module](api:fuse/geolocation/geolocation) for an example.
     */
     function stopListening(): void;
-
 }
 
 /**
@@ -1300,7 +1293,6 @@ declare module "FuseJS/ImageTools" {
                 .then(function(newImage) { console.log("Path of resized image is " + newImage.path); });
     */
     function resize(image: any, options: any): Promise<any>;
-
 }
 
 /**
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -0,0 +1,2151 @@
...
+                    },
+                    body: JSON.stringify({
+                        registration_ids: [regID],
+                        data: {
+                            notification: {
+                                alert: {
+                                    title: 'Well would ya look at that!',
+                                    body: 'Hello from some other app'
+                                }
+
```
kusma: Most other methods have documentation, why doesn't this?

kusma: Most other methods have documentation, why doesn't this?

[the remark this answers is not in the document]
kusma: Right, I now read what the reason is... But I think we should rename the method instead (but keep a compatibility shim around for backwards-compatibility)...

kusma:
> The biggest blocker to fully automate the generation of this, is that the source dataset is hosted on a different repository. (See [mortend/docs@f55cd5d](https://github.com/mortend/docs/commit/f55cd5dd133a7f59145914c152b76cd056399683).)

Can't we check out the docs repo on CI somewhere, and reach it from there?

Anyway, the whole concept of our documentation (both dataset and generator) being scattered between the docs-repo and the fuselibs/uno repos, is a major source of pain when doing anything docs-related. I have a vague plan to try to fix this: merge the docs generator into uno, and make a proper "uno doc" command that builds html directly from source code + md-files. Then we can generate docs for each uno and fuselibs release on CI, and upload these to github releases as well as deploying them somewhere for hosting. And to get there, I also need to move all docs into each repo

It would mean that our docs becomes split between uno and fuselibs, but I think that's OK...

index.d.ts
kusma: I'm not sure I understand. We have a comment saying "Copy a video to the camera roll." for the CopyVideoToCameraRoll method. Isn't that enough?

mortend: The former doc-comment only contain doc-attributes and no other text, so the documentation string (found in api-docs JSON files) is empty/missing, which caused my generator to output this method without having documentation.

mortend: It looks like that comment is missing the `@scriptmethod` attribute which is required to be picked up by my generator.

mortend:
What you're describing sounds like a better solution, though involving more work than I had in mind for this project. When we get a proper "uno doc" command, it will probably be helpful in generating this file in a better way. I guess I was looking for a quicker way to add TypeScript support.

I currently work with local-built fuselibs most of the time, and checking out the docs repo on CI will not work when using `npm link` on local-builds. Locally, I don't think we want to check out the docs repo, build C# code, apply patches etc, during build.

index.d.ts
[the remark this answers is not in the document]
mortend: I can look into renaming those methods.

kusma:
> I currently work with local-built fuselibs most of the time, and checking out the docs repo on CI will not work when using `npm link` on local-builds. Locally, I don't think we want to check out the docs repo, build C# code, apply patches etc, during build.

Right, good point.

mortend:
Rebased and applied changes from #1262. Also dropped the commit that comments out the `delete` methods.

mortend:
If we don't want to add this file to this git repo, I suppose another alternative is to publish it as a separate NPM package, `fusejs-typescript` or something...

mortend:
Whoops, I just made a blunder while merging `release-1.11` into `master`... I have fixed `master` now, but it looks like GitHub now thinks this PR has been merged... Sorry about this! :/

mortend:
Regarding TypeScript Definitions, I now think I'll go for the separate package approach and create a new `fusejs-types` repo for this.

`fusejs-types` can live its own life independent from `fuselibs`, and more easily be updated on demand. It'll be less of a problem if Definitions get out of sync with current documentation, and we can put the generator that made this file in the same repo (instead of in `docs`).

Maybe in the future, when we have better uno doc-tooling, we'll be better equipped to keep things in sync/automated, and can move the Definitions file back to `fuselibs`. Thank you for the feedback.

kusma:
I don't think it's a huge deal, it's just some extra maintenance and risk of accidentally stale documentation. We can merge it to master if it's worth it.

mortend:
In general, I think it's better to distribute definitions together with their implementations, so I feel at `fuselibs` is where they ultimately belong.

Maybe I'll open a new PR with this again. It looks like this PR is now locked forever and can't be reopened.

PR #1273, opened 2019-05-22
Add disk cache functionality to image

ichan-mb:
Add disk cache functionality to the image so that the next time we display an image it will no longer be downloaded from the network.

This PR contains:
- [ ] Changelog
- [x] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Elements/Resources/HttpImageSource.uno
+++ b/Source/Fuse.Elements/Resources/HttpImageSource.uno
@@ -2,6 +2,7 @@ using Uno;
 using Uno.Graphics;
 using Uno.Collections;
 using Uno.UX;
+using Uno.IO;
 using Fuse.Drawing;
 using Fuse.Resources.Exif;
 
@@ -20,6 +21,7 @@ namespace Fuse.Resources
 				<HttpImageSource Url="https://upload.wikimedia.org/wikipedia/commons/0/06/Kitten_in_Rizal_Park%2C_Manila.jpg" />
 			</Image>
 
+		To cache the image to the disk, you can add `DiskCache` attribute and set it to `true` so that the next time we display an image it will no longer be downloaded from the network but use from disk instead.
 	*/
 	public sealed class HttpImageSource : ImageSource
 	{
@@ -34,7 +36,7 @@ namespace Fuse.Resources
 				if(value == null || value == "" )
 					return;
 
-				_proxy.Attach( HttpImageSourceCache.GetUrl( value ) );
+				_proxy.Attach( HttpImageSourceCache.GetUrl( value, DiskCache ) );
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


AndrewEQ:
Very cool @ichan-mb! Is there a way to get the total size of the disk cache, and a function to clear the disk cache?

ichan-mb:
There is a function to clear individual cached image in the form of Image Scripting Class [here](https://github.com/fuse-open/fuselibs/pull/1273/commits/8af50e8eca9857d3556c5f1b545659e0a3bca6ba#diff-28bfc2810b3f8f0824efefbcece1c2e3R16) so you can remove a single cache image in Javascript. But it doesn't have a function to clear all cached image. 

I still wonder where should that functions (clearing all disk cache and getting the total size of the disk cache) be stored. I'm thinking in the FuseJS ImageTools module, but is that making sense?

AndrewEQ:
Ok cool, hmm, [ImageTools](https://fuseopen.com/docs/fuse/imagetools/imagetools.html) is around image manipulation, whereas the images disk cache is around "disk manipulation", what do you think of [Storage](https://fuseopen.com/docs/fuse/storage/storagemodule.html)?

...although, I guess if one thinks of doing something to do with images, then ImageTools makes sense whereas I would think about Storage as "storing files", probably wouldn't relate it to image disk cache, so my vote in the end, at least for now would be with ImageTools lol

Source/Fuse.Elements/Resources/HttpImageSource.uno
```diff
@@ -178,7 +226,21 @@ namespace Fuse.Resources
 			{
 				try
 				{
-					_tex = TextureLoader.ByteArrayToTexture2DContentType(_data, _contentType);
+					if (_data == null && _contentType == null)
+					{
+						_data = File.ReadAllBytes(_filename);
+						_tex = TextureLoader.ByteArrayToTexture2DFilename(_data, _filename);
```
kusma: I don't think this is going to work. `_filename` doesn't contain any extension, and will just end up [throwing an exception](https://github.com/fuse-open/fuselibs/blob/9c9ab807f1c6ea0a6a3256940b543d41b8f033fb/Source/Experimental.TextureLoader/TextureLoader.uno#L99)...

Source/Fuse.Elements/Resources/HttpImageSource.uno
```diff
@@ -68,12 +70,28 @@ namespace Fuse.Resources
...
+		/** Determines whether we use the disk cache to store downloaded images so that the next time we display an image it will no longer be downloaded from the network. Default is false. */
+		public bool DiskCache { get { return _diskCache; } set { _diskCache = value; } }
+
+		public void ClearCache()
+		{
+			if (Url != "")
+			{
+				string urlHascode = Path.Combine(Directory.GetUserDirectory(UserDirectory.Data), Url.GetHashCode().ToString());
+				string filename = urlHascode + ".jpg";
+				if (File.Exists(filename)) File.Delete(filename);
```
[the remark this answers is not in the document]
ichan-mb: Okay

Source/Fuse.Elements/Resources/HttpImageSource.uno
```diff
@@ -152,22 +181,41 @@ namespace Fuse.Resources
...
 				_contentType = ct;
 
-			_orientation = ExifData.FromByteArray(data).Orientation;
+			new BackgroundLoad(data, _filename, _contentType, SuccessCallback, FailureCallback);
+		}
 
-			new BackgroundLoad(data, _contentType, SuccessCallback, FailureCallback);
+		bool FindFileFromUrl(string url, out string filename)
+		{
+			string urlHascode = Path.Combine(Directory.GetUserDirectory(UserDirectory.Data), Url.GetHashCode().ToString());
```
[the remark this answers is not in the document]
ichan-mb: Yes I Agree, fix is underway

Source/Fuse.Elements/Resources/HttpImageSource.uno
ichan-mb: It will work because of this block statement only be executed by [this statement](https://github.com/fuse-open/fuselibs/pull/1273/files/c51dd1a1c3467cbb66b3bb1eebe359d8a075716f#diff-7c9578067052e57d861466f1e69e34a3R136).  And [that statement](https://github.com/fuse-open/fuselibs/pull/1273/files/c51dd1a1c3467cbb66b3bb1eebe359d8a075716f#diff-7c9578067052e57d861466f1e69e34a3R136) only be called when `FindFileFromUrl` method return true

kusma: Ah, I see, thanks for clarifying. I don't think this is very clear from the code, though. Perhaps we could use a differently named variable than `_filename`, when it's not actually the full filename? For instance `_filenameBase` or something like that?

I also think this code could use some tests that can be used to verify that corner-cases like this work...

kusma: (note that the last part is "could"; I don't want to force you to do this if you're pressed on time or something)

ichan-mb: Yeah I have to refactor some variable and method name to better readability

ichan-mb:
Thanks!

PR #1277, opened 2019-06-17
Fix critical bugs on DiskCache

### LINKED ISSUE #1276: Normal ImageFill doesn't load image
@ichan-mb, @kusma 

**Describe the bug**
My image from the web was not loading, just says that it couldn't load the .jpg.

**To Reproduce**

```
<ImageFill StretchMode="UniformToFill" WrapMode="ClampToEdge" MemoryPolicy="UnloadUnused">
  <MultiDensityImageSource>
    <HttpImageSource Density="1" Url="{image1}" />
    <HttpImageSource Density="2" Url="{image2}" />
    <HttpImageSource Density="3" Url="{image3}" />
  </MultiDensityImageSource>
</ImageFill>
```

**Expected behavior**
Image loads and displays

**Smartphone (please complete the following information):**
 - Device: iPhoneSE
 - OS: 12.3

**Additional context**
When I rollback the disk cache commits, the images load and display correctly.

ichan-mb:
This is a fix for when using `HttpImageSource` without `DiskCache` property, the image can't even be loaded. This resolve #1276 

And this also corrects when we don't use DiskCache, images that have been loaded from the network are not saved to disk

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Elements/Resources/HttpImageSource.uno
+++ b/Source/Fuse.Elements/Resources/HttpImageSource.uno
@@ -146,9 +146,9 @@ namespace Fuse.Resources
 			try
 			{
 				_loading = true;
-				if (_diskCache && IsFileCacheExist(out _filenameBase, out _contentType))
+				if (IsFileCacheExist(out _filenameBase, out _contentType) && _diskCache)
 				{
-					new BackgroundLoad(null, _filenameBase, _contentType, SuccessCallback, FailureCallback);
+					new BackgroundLoad(null, _filenameBase, _contentType, _diskCache, SuccessCallback, FailureCallback);
 				}
 				else
 				{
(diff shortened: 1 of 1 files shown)


Source/Fuse.Elements/Resources/HttpImageSource.uno
```diff
@@ -146,9 +146,9 @@ namespace Fuse.Resources
 			try
 			{
 				_loading = true;
-				if (_diskCache && IsFileCacheExist(out _filenameBase, out _contentType))
```
ichan-mb: This is the root cause, `IsFileCacheExist` will not be called when `_diskCache` variable is `false`. Silly me😩

PR #1281, opened 2019-06-24
Prepare for new Uno

mortend:
Some build adjustments due to changes in the new Uno version.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/AllTests.unoproj
+++ b/Source/AllTests.unoproj
@@ -1,4 +1,5 @@
 {
+  "BuildCondition": "UNO_TEST",
   "Packages": [
     "Fuse",
     "FuseJS",
--- a/Source/AllTests.unoproj
+++ b/Source/AllTests.unoproj
@@ -1,9 +1,5 @@
 {
   "BuildCondition": "UNO_TEST",
-  "Packages": [
-    "Fuse",
-    "FuseJS",
-  ],
   "Projects": [
     "**Test.unoproj"
   ],
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,6 @@ mono: 5.4.1
 
 env:
   global:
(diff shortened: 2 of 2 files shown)


mortend:
AppVeyor :/

```
Since you ran multiple projects, here is a summary:
From 35 projects, ran 1213 tests, 3 failed.
  Failed:  Fuse.Navigation.Test.RouterTest.RelativeNest
  Failed:  Fuse.Navigation.Test.RouterTest.RelativeNestModify
  Failed:  Fuse.Navigation.Test.ActivatedTest.Deep
```

Source/AllTests.unoproj
```diff
@@ -1,4 +1,5 @@
 {
+  "BuildCondition": "UNO_TEST",
```
kusma: Hmm, this makes it sound like uno#197 is a regression, and that similar issues might be triggered in other repos...

Source/AllTests.unoproj
```diff
@@ -1,9 +1,5 @@
 {
   "BuildCondition": "UNO_TEST",
-  "Packages": [
-    "Fuse",
-    "FuseJS",
-  ],
```
kusma: Nice :+1:

mortend:
Yeah, it's a little bit iffy... But will do more good than harm I think...

A better solution might be to introduce project types (Library, Test, App), but that's more work and will also require changing project files in repos. `uno doctor` will only care about Library projects, `uno test` about Test, and so on. We can keep backwards compatibility and give warnings on projects missing a project type. Might implement this later.

kusma:
Yeah, let's see if it turns out to be a problem or not.

Source/AllTests.unoproj
mortend: The regression would be that the new version of `uno doctor` will find and build more projects. Shouldn't be a blocker if it happens though.

PR #1283, opened 2019-06-27
Use --run-local when running tests

mortend:
If we're lucky, this might help with instability on AppVeyor.

It worked on the first try when running on my personal AppVeyor at least.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend:
Woohoo, it worked two times in a row.

I've been thinking for a long time to remove the Web Server stuff in our test runner. Since we removed WebGL support, I don't think it's needed anymore, and that it only complicates things for us.

PR #1287, opened 2019-08-23
Fuse.Charting: drop experimental attributes

mortend:
This should remove the following notice from generated pages at https://fuseopen.com/docs.

    This entity is experimental and might be changed or removed in a future release.

Because Fuse.Animations.AttractorConfig is commonly used in Fuse.Charting code, we
also drop the experimental attribute from this class.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Animations/Attract.uno
+++ b/Source/Fuse.Animations/Attract.uno
@@ -14,7 +14,6 @@ namespace Fuse.Animations
 		A single `AttractorConfig` can be used for multiple `attract` expressions.
 		
 		@see @Attract
-		@experimental
 	*/
 	public class AttractorConfig : DestinationMotionConfig
 	{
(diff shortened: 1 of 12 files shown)


Source/Fuse.Animations/Attract.uno
```diff
@@ -14,7 +14,6 @@ namespace Fuse.Animations
 		A single `AttractorConfig` can be used for multiple `attract` expressions.
 		
 		@see @Attract
-		@experimental
```
kusma: This one isn't in Fuse.Charting, did you mean to include it?

mortend: `AttractorConfig` is used a lot in Fuse.Charting code so I think we should drop `@experimental` here too.

kusma: Ok. Perhaps you could include it in the commit message?

mortend: Fixed.

PR #1290, opened 2019-08-31
iOS 13 DarkMode

AndrewEQ:
Oh so dark and lovely :smile_cat:

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.Native/Fuse.Controls.Native.unoproj
+++ b/Source/Fuse.Controls.Native/Fuse.Controls.Native.unoproj
@@ -25,6 +25,7 @@
     "../Fuse.Elements/Fuse.Elements.unoproj",
     "../Fuse.Nodes/Fuse.Nodes.unoproj",
     "../Fuse.Common/Fuse.Common.unoproj",
+    "../Fuse.Scripting/Fuse.Scripting.unoproj",
   ],
   "Includes": [
     "*",
@@ -34,6 +35,8 @@
     "iOS/Helpers.mm:ObjCSource:iOS",
     "iOS/CanvasViewGroup.h:ObjCHeader:iOS",
     "iOS/CanvasViewGroup.mm:ObjCSource:iOS",
+    "iOS/ContainerView.h:ObjCHeader:iOS",
+    "iOS/ContainerView.mm:ObjCSource:iOS",
     "Android/Java/ScrollEventHandler.java:Java:Android",
     "Android/Java/FuseScrollView.java:Java:Android",
     "Android/Java/HorizontalScrollView.java:Java:Android",
--- a/Source/Fuse.Controls.Native/iOS/ContainerView.mm
+++ b/Source/Fuse.Controls.Native/iOS/ContainerView.mm
@@ -1,0 +1,50 @@
+#import "ContainerView.h"
+#import "uObjC.Foreign.h"
+@{Fuse.Controls.Native.iOS.DarkMode:IncludeDirective}
+
+@implementation ContainerView
+
+- (void) viewDidLoad {
+	#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
+	if (SYSTEM_VERSION_GREATER_THAN_OR_EQUAL_TO(@"13.0")) {
+
+		switch(self.traitCollection.userInterfaceStyle) {
+			case UIUserInterfaceStyleUnspecified: 
+				@{Fuse.Controls.Native.iOS.DarkMode.changeDarkMode(string):Call(@"Unspecified")};
+				break;
+			case UIUserInterfaceStyleLight: 
+				@{Fuse.Controls.Native.iOS.DarkMode.changeDarkMode(string):Call(@"Light")};
+				break;
+			case UIUserInterfaceStyleDark: 
+				@{Fuse.Controls.Native.iOS.DarkMode.changeDarkMode(string):Call(@"Dark")};
+				break;
+		}
+	}
+	#endif
+}
+
+- (void) traitCollectionDidChange: (UITraitCollection *) previousTraitCollection {
+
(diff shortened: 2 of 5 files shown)


ichan-mb:
Yey! so cool @AndrewEQ, but I think this feature should exist for Android too, so we keep cross-platform availability in Fuselibs. Refer to this for Android Q: https://developer.android.com/preview/features/darktheme#configuration_changes

AndrewEQ:
@ichan-mb yeah, would have done it but I didn't have any device with Q to test on, do you? 😸

AndrewEQ:
@mortend Travis failed cos its using an older version of XCode (9.2), I was testing on version 11 beta 6

mortend:
@AndrewEQ Maybe we can use the preprocessor to only include iOS 13/Xcode 11 specific code when available, and fallback to printing a warning in the log or something like that when building on older versions?

Similar to what we do in https://github.com/fuse-open/fuselibs/blob/master/Source/Fuse.Platform/iOS/SystemUI.uno#L256

AndrewEQ:
@mortend ok thanks, I'll have a look at it doing it like that

AndrewEQ:
@mortend the only problem I found was trying to exclude the separate objective-c files, and so older XCode would complain about the new lines of code as the framework doesn't exist.

AndrewEQ:
seems the only way forward, is forward lol, unless we can have an older and newer XCode check?

kusma:
@AndrewEQ: I don't follow... @mortend did indeed suggest to make the code XCode version dependent, and he showed an example of how to do that...

AndrewEQ:
@kusma that code example is for code written inside uno, how do you exclude whole objective-c files?

mortend:
@AndrewEQ you can also use `#ifdef..#endif` inside Obj-C files to exclude all or parts of the code. I can probably help if you're stuck.

mortend:
@AndrewEQ I don't have iOS 13 myself yet, but I think this will work: https://github.com/mortend/fuselibs/commit/49b78871c70f2dd68f7c74e6f15f6ce3b0bc5fb0

AndrewEQ:
Ok updated and trying again...

Source/Fuse.Controls.Native/iOS/ContainerView.mm
```diff
@@ -0,0 +1,49 @@
...
+				@{Fuse.Controls.Native.iOS.DarkMode.changeDarkMode(string):Call(@"Dark")};
+				break;
+		}
+	}
+	#endif
+}
+
+- (void) traitCollectionDidChange: (UITraitCollection *) previousTraitCollection {
+
+	#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: It's probably a good idea to always invoke the super class, so I think the `#if` should be placed after invoking `[super traitCollectionDidChange: previousTraitCollection]`.

Source/Fuse.Controls.Native/iOS/GraphicsView.uno
```diff
@@ -49,7 +50,10 @@ namespace Fuse.Controls.Native.iOS
 		[Foreign(Language.ObjC)]
 		static ObjC.Object CreateContainer()
 		@{
-			UIView* view = [[UIView alloc] init];
+			ContainerView* view = [[ContainerView alloc] init];
+			#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: Since we already protect for iOS 13 inside the `viewDidLoad` method, protecting again here is redundant and may be confusing to future readers. We should probably drop the `#if...#endif` stuff here.

Source/Fuse.Controls.Native/Fuse.Controls.Native.unoproj
```diff
@@ -2,9 +2,6 @@
   "Copyright": "Copyright (c) Fuse Open 2018",
   "Description": "Abstraction over platform-specific native controls",
   "Publisher": "Fuse Open",
-  "Package": {
```
mortend: Whoops, this happened again. Maybe rebase? :)

Source/Fuse.Controls.Native/iOS/GraphicsView.uno
AndrewEQ: Oh, I just thought of not calling a function to save CPU? 😄

Source/Fuse.Controls.Native/iOS/ContainerView.mm
AndrewEQ: Cool, will do!

AndrewEQ: Done!

Source/Fuse.Controls.Native/iOS/GraphicsView.uno
AndrewEQ: Done!

Source/Fuse.Controls.Native/Fuse.Controls.Native.unoproj
AndrewEQ: Done!

Source/Fuse.Controls.Native/iOS/GraphicsView.uno
mortend: Nice thought, but I doubt you're actually saving much CPU and this code is probably not very performance critical. Premature optimization is the root of all evil. ;)

AndrewEQ:
Woohoo!

mortend:
Regarding Android Q, just thought I'd mention that it's possible to target `x86` and test Android Q using the Android Emulator in Android Studio.

```
  "Android": {
    "Architectures": {
      "Debug": ["x86"]
    }
  },
```

AndrewEQ:
Oh cool, I'll look into it soon

PR #1291, opened 2019-08-31
iOS 13 SystemFont Changes + Previous System Font Issues

AndrewEQ:
- added iOS13 way of doing SystemFont
- fixed older system font issues with font weight/style
- added reasonable fallbacks for missing font weights for older system font way

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Common/Internal/SystemFont.uno
+++ b/Source/Fuse.Common/Internal/SystemFont.uno
@@ -11,6 +11,13 @@ namespace Fuse.Internal
 		public readonly int Index;
 		public readonly IEnumerable<string> Styles;
 
+		//iOS13+ - for default systemfont
+		extern(IOS) internal bool IsCustomFont = true;
+		extern(IOS) internal Fuse.SystemFont.Style FontStyle; //normal/italic
+		extern(IOS) internal Fuse.SystemFont.Weight FontWeight; //thin/light/regular/bold/etc
+		extern(IOS) internal Fuse.SystemFont.Design FontDesign; //default/monospaced/rounded/serif
+		extern(IOS) internal int FontSize; //12/16/etc
+
 		/** Create a `FontFaceDescriptor` that selects the face by matching the its string. */
 		public FontFaceDescriptor(FileSource fileSource, IEnumerable<string> styles)
 		{
@@ -51,6 +58,15 @@ namespace Fuse.Internal
 				hash = hash * 23 + s.GetHashCode();
 			}
 			return hash;
+		}
+
+		extern(IOS) internal void SetFontAttributes(Fuse.SystemFont.Style fontStyle, Fuse.SystemFont.Weight fontWeight, Fuse.SystemFont.Design fontDesign, int fontSize, bool isCustomFont)
+		{
+			FontStyle = fontStyle;
+			FontWeight = fontWeight;
+			FontDesign = fontDesign;
+			FontSize = fontSize;
+			IsCustomFont = isCustomFont;
 		}
 	}
 
--- a/Source/Fuse.Common/Internal/iOSSystemFont.uno
+++ b/Source/Fuse.Common/Internal/iOSSystemFont.uno
@@ -14,16 +14,30 @@ namespace Fuse.Internal
 		static HashSet<string> _families;
 		static Dictionary<string, string> _fontPaths;
 		static readonly string[] _weightNames = new string[]
-			{ "UltraLight", "Thin", "Light", "Regular", "Medium", "SemiBold", "Bold", "Heavy", "Black" };
+			{ "UltraLight", "Thin", "Light", "Regular", "Medium", "Semibold", "Bold", "Heavy", "Black" };
 
 		public static List<FontFaceDescriptor> Default
 		{
 			get
 			{
-				var descriptor = GetDefaultUIFontDescriptor();
-				var descriptors = GetFallbackUIFontDescriptors(descriptor);
-				var fontNames = GetFontNamesFromUIFontDescriptors(descriptors);
-				return GetDescriptorsFromFontNames(fontNames);
+				if (Fuse.iOSDevice.OperatingSystemVersion.Major >= 13) 
+				{
+					var path = "/System/Library/Fonts/Core/AppleSystemUIFont";
+					var fontName = ".AppleSystemUIFont";
+
+					var ffd = new FontFaceDescriptor(new SystemFileSource(path), GetFontNameStyles(fontName));
+					ffd.SetFontAttributes(Fuse.SystemFont.Style.Normal,Fuse.SystemFont.Weight.Normal,Fuse.SystemFont.Design.Default, 16, false);
+					var result = new List<FontFaceDescriptor>();
+					result.Add(ffd);
+					return result;
+				} 
+				else 
+				{
+					var descriptor = GetDefaultUIFontDescriptor();
+					var descriptors = GetFallbackUIFontDescriptors(descriptor);
+					var fontNames = GetFontNamesFromUIFontDescriptors(descriptors);
+					return GetDescriptorsFromFontNames(fontNames);
+				}
 			}
 		}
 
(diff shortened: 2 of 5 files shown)


AndrewEQ:
This PR is also failing the same way as https://github.com/fuse-open/fuselibs/pull/1290, because the CI is not using latest XCode which includes iOS13 frameworks.

AndrewEQ:
ok, updated with iOS 13 checks, if it passes, will probably still need someone to test on XCode version less than 11

AndrewEQ:
oh, the test would be simple like being able to build an iOS app and the default font still works as well as a custom defined one.

mortend:
I can test this on Xcode 9.4 and iOS 12.4 later tonight.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -14,16 +14,34 @@ namespace Fuse.Internal
...
 
 		public static List<FontFaceDescriptor> Default
 		{
 			get
 			{
-				var descriptor = GetDefaultUIFontDescriptor();
-				var descriptors = GetFallbackUIFontDescriptors(descriptor);
-				var fontNames = GetFontNamesFromUIFontDescriptors(descriptors);
-				return GetDescriptorsFromFontNames(fontNames);
+				#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: These are C preprocessor macros and won't work in Uno code. Since we run-time check for iOS 13 on the line below, I think we can just remove the `#if..#endif` stuff here.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -14,16 +14,34 @@ namespace Fuse.Internal
...
+				if (Fuse.iOSDevice.OperatingSystemVersion.Major >= 13) {
```
mortend: Curly braces should go on separate lines in Uno code. :)

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -14,16 +14,34 @@ namespace Fuse.Internal
...
+					return result;
+				} else {
+				#endif
+
+					var descriptor = GetDefaultUIFontDescriptor();
+					var descriptors = GetFallbackUIFontDescriptors(descriptor);
+					var fontNames = GetFontNamesFromUIFontDescriptors(descriptors);
+					return GetDescriptorsFromFontNames(fontNames);
+
+				#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: More C macros.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -51,18 +69,71 @@ namespace Fuse.Internal
...
 		{
-			var weightIndex = Math.Clamp((int)weight, 0, _weightNames.Length - 1);
-			var descriptor = GetMatchingFontDescriptor(family, style == Fuse.SystemFont.Style.Italic, _weightNames[weightIndex]);
-			if (descriptor == null)
-			{
-				return Default;
-			}
-			else
-			{
+			#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: More C macros.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -51,18 +69,71 @@ namespace Fuse.Internal
...
+
+			#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
 			}
+			#endif
+		}
+
+		static ObjC.Object GetFallbackUIFontDescriptorsWeight(ObjC.Object descriptor, string family, string weightName, bool isItalic) 
+		{
+			var fontName = (!isItalic) ? GetDescriptorFontName(descriptor) : GetDescriptorName(descriptor);
+			if (weightName == "UltraLight" && isItalic == false && (DoesFontWeightExist(family, fontName, weightName) == false)) {
```
mortend: Curly braces should go on separate lines, but in this case all curly braces can be removed because it's all single statements.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -103,6 +174,7 @@ namespace Fuse.Internal
 					if (i > start + 1)
 					{
 						var substr = fontName.Substring(start, i - start);
+						// debug_log("style: " + substr);
```
mortend: Remove leftover `debug_log`.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -329,6 +433,110 @@ namespace Fuse.Internal
...
+			UIFontDescriptor* descriptor = [UIFont systemFontOfSize: fontSize weight:selectedWeight].fontDescriptor;
+
+			//normal by default or italic
+			if ([fontStyle isEqualToString:@"italic"]) {
+				descriptor = [descriptor fontDescriptorWithSymbolicTraits:
+					[descriptor symbolicTraits] | UIFontDescriptorTraitItalic];
+			}
+
+			//design
+			#if defined(__IPHONE_13_0) && __IPHONE_OS_VERSION_MAX_ALLOWED >= __IPHONE_13_0
```
mortend: This is foreign Obj-C code, so it's fine to use the C macros here.

Source/Fuse.Controls.Native/iOS/FontCache.uno
```diff
@@ -37,12 +37,23 @@ namespace Fuse.Controls.Native.iOS
 			var path = GetOptionalPath(descriptor.FileSource);
 			if (path != null)
 			{
-				var uifontdescriptor
+				if (descriptor.IsCustomFont) {
```
mortend: More curly braces that should go on separate lines.

Source/Fuse.Common/Internal/SystemFont.uno
```diff
@@ -11,6 +11,13 @@ namespace Fuse.Internal
 		public readonly int Index;
 		public readonly IEnumerable<string> Styles;
 
+		//iOS13+ - for default systemfont
+		public bool IsCustomFont = true;
```
mortend: Are these specific for iOS? If so, we can do `extern(IOS) bool IsCustomFont = ...`.

Also, can we drop `public`? Or make `internal` instead?

Source/Fuse.Common/SystemFont.uno
```diff
@@ -42,7 +42,7 @@ namespace Fuse
 			Light,
 			Normal,
 			Medium,
-			SemiBold,
+			Semibold,
```
mortend: Here we're renaming a publicly visible entity, which is a breaking change.

That's fine if we properly deprecate the old entity by adding the following at the end of the enum:
```
[Obsolete]
SemiBold = Semibold
```

Then potential users of the old code will get a warning in their build log, instead of build errors.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -329,6 +433,110 @@ namespace Fuse.Internal
 			throw new Exception("iOSSystemFont: No matching style in " + fileName);
 		}
 
+
+
+		public static ObjC.Object GetMatchingSystemFontDescriptor(Fuse.SystemFont.Style fontStyle, Fuse.SystemFont.Weight fontWeight, Fuse.SystemFont.Design fontDesign, int fontSize)
```
mortend: Drop `public` or make `internal`?

Source/Fuse.Common/Internal/SystemFont.uno
AndrewEQ: `extern(iOS)` breaks compilation, was able to make it internal

Source/Fuse.Common/SystemFont.uno
AndrewEQ: Like so right?
```
                public enum Weight
		{
			UltraLight,
			Thin,
			Light,
			Normal,
			Medium,
			Semibold,
			Bold,
			Heavy,
			Black,

			[Obsolete]
			SemiBold = Semibold
		}
```

mortend: Yeah :)

Source/Fuse.Common/Internal/iOSSystemFont.uno
AndrewEQ: was able to make internal

AndrewEQ:
Ok, sorted.

Source/Fuse.Common/Internal/SystemFont.uno
```diff
@@ -11,6 +11,13 @@ namespace Fuse.Internal
 		public readonly int Index;
 		public readonly IEnumerable<string> Styles;
 
+		//iOS13+ - for default systemfont
+		internal bool IsCustomFont = true;
```
mortend: I think the iOS specific fields can be `extern(IOS) internal bool IsCustomFont ...` if we also make the SetFontAttributes method `extern(IOS) internal void SetFontAttributes(...)`.

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -120,21 +184,40 @@ namespace Fuse.Internal
...
+		static string GetDescriptorName(ObjC.Object descriptor)
 		@{
 			return ((UIFontDescriptor*)descriptor).fontAttributes[UIFontDescriptorNameAttribute];
 		@}
 
 		[Foreign(Language.ObjC)]
 		static string GetDefaultFontFamily()
 		@{
-			return [UIFont systemFontOfSize: 12].familyName;
+			return [UIFont systemFontOfSize: 16].familyName;
```
mortend: It's unclear to me why we suddenly change font size from 12 to 16...

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -120,21 +184,40 @@ namespace Fuse.Internal
...
+		@}
+
+		[Foreign(Language.ObjC)]
+		static ObjC.Object GetDefaultFontDescriptor(string weight, bool isItalic)
+		@{
+			NSDictionary* attributes
+				= \@{ UIFontDescriptorFamilyAttribute: [UIFont systemFontOfSize: 12].familyName };
```
mortend: ... but here the font size is 12 again..?

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -329,6 +425,110 @@ namespace Fuse.Internal
...
 
+
+
+		internal static ObjC.Object GetMatchingSystemFontDescriptor(Fuse.SystemFont.Style fontStyle, Fuse.SystemFont.Weight fontWeight, Fuse.SystemFont.Design fontDesign, int fontSize)
+		{
+			string selectedFontStyle = "normal";
+			string selectedFontWeight = "normal";
+			string selectedFontDesign = "default";
+
+			switch(fontStyle) {
```
mortend: Code style nits (three occurrences in this method):
- Add space between `switch` and `(`
- Move curly braces to their own lines

Source/Fuse.Common/Internal/iOSSystemFont.uno
```diff
@@ -329,6 +425,110 @@ namespace Fuse.Internal
...
+			return getSystemFontDesign(selectedFontStyle, selectedFontWeight, selectedFontDesign, fontSize);
+		}
+
+
+		[Foreign(Language.ObjC)]
+		static ObjC.Object getSystemFontDesign(string fontStyle, string fontWeight, string fontDesign, int fontSize)
+		@{
+			//weight
+			UIFontWeight selectedWeight;
+			if ([fontWeight isEqualToString:@"ultralight"]) {
```
mortend: Code style nitpicking, but IMO all curly braces in this method should be removed (because of single-statement), or moved to their own lines.

Source/Fuse.Common/Internal/SystemFont.uno
AndrewEQ: I get this when using `extern(IOS) ...`:
```
fuselibs/Source/Fuse.Common/Internal/SystemFont.uno(69.4): E3102: There is nothing named IsCustomFont accessible in this scope. Are you missing a package reference?
```

AndrewEQ: works for the SetFontAttributes method

Source/Fuse.Common/Internal/iOSSystemFont.uno
AndrewEQ: I standardised it at 16 (believe me you won't see the default size get bigger cos the defaults always been 16 😄)

AndrewEQ:
Sorted.

Source/Fuse.Common/Internal/SystemFont.uno
mortend: Line 69 is inside the SetFontAttributes method, so if you do `extern(IOS)` on both the method and these fields, it should build successfully.

AndrewEQ: ok, sorted.

Source/Fuse.Common/Internal/iOSSystemFont.uno
mortend: I see it now. ;)

AndrewEQ:
Shweet!

PR #1296, opened 2019-09-11
Fuse.Models: upgrade zone.js to v0.9.1

mortend:
This upgrades zone.js to the most recent version. Our current version hasn't been updated in almost two years, so it's probably a good idea to upgrade and get latest fixes and optimizations.

I've manually tested my own apps built using FuseJS Models, and things still work...

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend:
Zone.js isn't currently distributed standalone on NPM so for now it was a lot easier to replace the existing source-files. Maybe next time. :)

PR #1308, opened 2019-09-17
Fixed launcher threading - Fuse.iOS - OpenURL

AndrewEQ:
Fixed threading issue for the iOS url launcher

### DIFF
--- a/Source/Fuse.iOS/iOSInterop.uno
+++ b/Source/Fuse.iOS/iOSInterop.uno
@@ -52,7 +52,9 @@ namespace Fuse.iOS.Bindings
 		[Foreign(Language.ObjC)]
 		public static extern(iOS) void LaunchUriiOS(string uri)
 		@{
-			[[UIApplication sharedApplication] openURL:[NSURL URLWithString:uri]];
+			dispatch_sync(dispatch_get_main_queue(), ^{
+				[[UIApplication sharedApplication] openURL:[NSURL URLWithString:uri]];
+			});
 		@}
 	}
 }


Source/Fuse.iOS/Fuse.iOS.unoproj
```diff
@@ -2,9 +2,6 @@
   "Copyright": "Copyright (c) Fuse Open 2018",
   "Description": "Native UI and bootstrapping for iOS",
   "Publisher": "Fuse Open",
-  "Package": {
-    "ProjectUrl": "https://fuseopen.com"
-  },
```
kusma: This is already like this on master, perhaps you could rebase on master to make this go away? :)

AndrewEQ:
oops, my bad

Source/Fuse.iOS/Fuse.iOS.unoproj
AndrewEQ: strange where that came from cos I originally rebased, anyhoo, sorted it out

AndrewEQ:
Cool, how can we get travis to run again?

kusma:
I believe it's already re-running, something seems to be wrong with the reporting:
https://travis-ci.org/fuse-open/fuselibs/builds/586046406

kusma:
@AndrewEQ: I don't know why it's stuck. The new build passed, but it's not tagged correctly. Probably a glitch due to the closing of the PR and poor timing.

Perhaps you could do a `git commit --amend --reset-author` to give the tip of the branch a new timestamp, and then re-push?

AndrewEQ:
ok, lemme try get out and give it a push

PR #1309, opened 2019-09-19
Android Q(29) - Dark Theme Support

AndrewEQ:
Added support for Android Dark Theme to our "Dark Mode" module: https://developer.android.com/guide/topics/ui/look-and-feel/darktheme

Needs Uno PR as well: https://github.com/fuse-open/uno/pull/236

### DIFF
--- a/Source/Fuse.Controls.Native/DarkMode.uno
+++ b/Source/Fuse.Controls.Native/DarkMode.uno
@@ -64,18 +64,19 @@ namespace Fuse.Controls.Native.iOS
 		"android.app.Activity")]
 	public class DarkMode : NativeEventEmitterModule
 	{
-		public static readonly DarkMode _instance;
+		internal static readonly DarkMode _instance;
 
 		public DarkMode(): base(true, "changed")
 		{
 			if (_instance != null) return;
 			Resource.SetGlobalKey(_instance = this, "FuseJS/DarkMode");
 
-			SetupAndroidListener();
+			if defined(ANDROID)
+				SetupAndroidListener();
 		}
 
 		[Foreign(Language.Java)]
-		public extern(Android) void SetupAndroidListener() 
+		extern(Android) void SetupAndroidListener() 
 		@{ 
 			@{checkForDarkThemeChange():Call()};
 
@@ -95,12 +96,8 @@ namespace Fuse.Controls.Native.iOS
 			});
 		@}
 		
-		public extern(!MOBILE) void SetupAndroidListener() @{ debug_log("SetupAndroidListener not supported on this platform."); @}
-
-		public extern(iOS) void SetupAndroidListener() @{ @}
-
 		[Foreign(Language.Java)]
-		public static extern(Android) void checkForDarkThemeChange()
+		static extern(Android) void checkForDarkThemeChange()
 		@{
 			switch (com.fuse.Activity.getRootActivity().getResources().getConfiguration().uiMode & android.content.res.Configuration.UI_MODE_NIGHT_MASK) {
 				case android.content.res.Configuration.UI_MODE_NIGHT_YES:
@@ -117,7 +114,7 @@ namespace Fuse.Controls.Native.iOS
 		@}
 
 
-		public static void changeDarkMode(string modeValue) {
+		internal static void changeDarkMode(string modeValue) {
 
 			if (DarkMode._instance == null) {
 				new DarkMode();
--- a/Source/Fuse.Controls.Native/DarkMode.uno
+++ b/Source/Fuse.Controls.Native/DarkMode.uno
@@ -58,6 +58,10 @@ namespace Fuse.Controls.Native.iOS
 		</App>
 		```
 	*/ 
+	[ForeignInclude(Language.Java,
+		"java.lang.Runnable",
+		"android.content.res.Configuration",
+		"android.app.Activity")]
 	public class DarkMode : NativeEventEmitterModule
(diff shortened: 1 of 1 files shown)


Source/Fuse.Controls.Native/DarkMode.uno
```diff
@@ -58,6 +58,10 @@ namespace Fuse.Controls.Native.iOS
 		</App>
 		```
 	*/ 
+	[ForeignInclude(Language.Java,
+		"java.lang.Runnable",
+		"android.content.res.Configuration",
+		"android.app.Activity")]
 	public class DarkMode : NativeEventEmitterModule
```
mortend: This code is from the other PR, but on a second thought, I think we should drop `public` here because we don't want users to use this class directly.

Source/Fuse.Controls.Native/DarkMode.uno
```diff
@@ -66,20 +70,59 @@ namespace Fuse.Controls.Native.iOS
 		{
 			if (_instance != null) return;
 			Resource.SetGlobalKey(_instance = this, "FuseJS/DarkMode");
+
+			SetupAndroidListener();
```
mortend: Here we can do
```
if defined(ANDROID)
    SetupAndroidListener();
```

After that we can drop the following dummy variants of this method for `iOS` and `!MOBILE`.

Source/Fuse.Controls.Native/DarkMode.uno
```diff
@@ -66,20 +70,59 @@ namespace Fuse.Controls.Native.iOS
...
+				}
+			});
+		@}
+		
+		public extern(!MOBILE) void SetupAndroidListener() @{ debug_log("SetupAndroidListener not supported on this platform."); @}
+
+		public extern(iOS) void SetupAndroidListener() @{ @}
+
+		[Foreign(Language.Java)]
+		public static extern(Android) void checkForDarkThemeChange()
```
mortend: Drop `public`.

Source/Fuse.Controls.Native/DarkMode.uno
```diff
@@ -66,20 +70,59 @@ namespace Fuse.Controls.Native.iOS
...
+					break; 
+				case android.content.res.Configuration.UI_MODE_NIGHT_UNDEFINED:
+					@{changeDarkMode(string):Call("Unspecified")};
+					break; 
+			}
+
+		@}
+
+
 		public static void changeDarkMode(string modeValue) {
```
mortend: This is from other PR, but maybe change to `internal` instead of `public`.

Source/Fuse.Controls.Native/DarkMode.uno
```diff
@@ -66,20 +70,59 @@ namespace Fuse.Controls.Native.iOS
...
+
+			SetupAndroidListener();
 		}
 
-		/* 
-			Example call for when implementing Android version:
-			@{Fuse.Controls.Native.iOS.DarkMode.changeDarkMode(string):Call(@"Dark")};
-		*/
+		[Foreign(Language.Java)]
+		public extern(Android) void SetupAndroidListener()
```
mortend: Drop `public`.

Source/Fuse.Controls.Native/DarkMode.uno
AndrewEQ: I needed it public to be able to call the function changeDarkMode function from external Objective-C file

Source/Fuse.Controls.Native/DarkMode.uno
AndrewEQ: Ok cool...

Source/Fuse.Controls.Native/DarkMode.uno
AndrewEQ: aight

Source/Fuse.Controls.Native/DarkMode.uno
AndrewEQ: aight

Source/Fuse.Controls.Native/DarkMode.uno
AndrewEQ: aight

Source/Fuse.Controls.Native/DarkMode.uno
mortend: In theory that should still work without public, or we have found a bug in the compiler I think. If you get compiler errors without public it's probably fine to keep it public.

Source/Fuse.Controls.Native/DarkMode.uno
```diff
@@ -58,6 +58,10 @@ namespace Fuse.Controls.Native.iOS
 		</App>
 		```
 	*/ 
+	[ForeignInclude(Language.Java,
+		"java.lang.Runnable",
+		"android.content.res.Configuration",
+		"android.app.Activity")]
 	public class DarkMode : NativeEventEmitterModule
 	{
 		public static readonly DarkMode _instance;
```
mortend: One more `public` that should be dropped, or make `internal` if it's used outside this class.

Source/Fuse.Controls.Native/DarkMode.uno
AndrewEQ: yeah, I had compiler errors of it not finding the class

mortend: I see, then public it is. :)

PR #1310, opened 2019-09-20
FuseJS/UserSettings

ichan-mb:
`FuseJS/UserSettings` module provides key-value pairs mechanism to store and retrieve primitive data types (string, number, boolean) as well as an array and a JSON object.

You can use this module to store information such as configuration data, application states, etc.

This PR contains:
- [x] Changelog
- [x] Documentation
- [ ] Tests


Source/Fuse.Storage/AndroidUserSettings.uno
```diff
@@ -0,0 +1,124 @@
...
+		public string GetStringValue(string key)
+		{
+			return GetStringValue(_handle, key);
+		}
+
+		[Foreign(Language.Java)]
+		string GetStringValue(Java.Object handle, string key)
+		@{
+			return ((SharedPreferences)handle).getString(key, "");
+		@}
```
kusma: OK, so here you have a separate method just so you can easily pass _handle. This gets a bit more tedious to read than it has to, IMO. A pattern I've used in the past is to have an internal static class, called something like `SharedPreferencesHelpers`, and putting all the foreign code on that. This reduces context-switching a bit when reading the code. You can see [examples of this](https://github.com/fuse-open/uno/blob/master/lib/Uno.Threading/Implementation/Posix/PthreadHelpers.uno) in our threading code.

Another alternative is to drop the Uno-wrappers, and use `@{AndroidUserSettings:Of(_this)._handle:Get()}` to get _handle instead of passing it in.

A final option is to simply just move all the helpers to the top of the source-file and call it the day :)

This is all largely taste-based, so I leave it up to you, just wanted to make sure you're aware of these options.

ichan-mb: Ok I'll try to simplifly it by removing unnecessary uno wrapper. thanks

Source/Fuse.Storage/UserSettingsImpl.uno
```diff
@@ -0,0 +1,269 @@
...
+		void SetStringValue(string key, string value);
+
+		void SetNumberValue(string key, double value);
+
+		void SetBooleanValue(string key, bool value);
+
+		void Remove(string key);
+
+		void Clear();
+	}
```
kusma: So, I'm not a huge fan of using interfaces to hide platform differences. The reason is that interfaces have a relatively large performance penalty (lots of vtable work), and we don't have a proper devirtualizer in Uno.

So I think it's better to do stuff like

```C#
if defined(iOS)
    iOSImpl.DoStuff();
else if defined(Android)
    AndroidImpl.DoStuff();
else
    build_error;
```

Yes, it's uglier, and yes, it's just papering over a less than ideal compiler, but I doubt that last part will be fixed anytime soon, and the performance problem just keeps adding up everywhere if we don't work actively to avoid it.

mortend: It should be fairly easy to change this code to `abstract class` instead of `interface`, since there's no need for multiple inheritance here.

On `abstract class` all function pointers are known at compile-time and performance is good (comparable to calling a function via pointer in native C). Calling methods on `interface` will involve first looking-up the vtable, which adds a little more run-time overhead. In this case, it's essentially one type-check and return a pointer to the first vtable, since only one interface is used. `interface` performance in Uno isn't a disaster I think, but `abstract class` is faster, and plain non-virtual methods (as kusma suggests) are fastest (by a only small margin I think, and more ugly...).

kusma: I think the overhead depends on how often you'd call the code. For something like this, I'd expect quite a few calls in common cases, so I think we should consider keeping the overhead as low as possible.

Abstract class has pretty bad cache-behavior. These kinds of things are also often the reason why OOP programs tend to be more sluggish than other languages, as things like this adds up all over the software-stack. If the mind-set is to do the known-performant thing by default, you're going to get much less overhead-buildup.

mortend: This is all true, but when looking at the bigger picture we see that the code already involves a bunch of overhead we have little control over:

1. The underlying implementations are invoking platform-specific Java or Obj-C APIs, where I'm fairly confident both languages already involve more overhead regarding method calls than is being introduced by the Uno-layer.

2. And ultimately, clients will interact with this code via the JavaScript-engine, adding even more run-time overhead.

I'd argue that even if we make the Uno code as fast as possible, the win will be relatively very small compared to the overhead in 1 and 2, and we can't avoid bad cache-behavior.

So, I just wanted to suggest an alternative solution that may possibly fix 80% of the problem by 20% of the engineering-effort, simply switching from `interface` to `abstract class`. That said, the solution you're suggesting is definitely the fastest one.

kusma: Right. Yeah, perhaps the bridging into the other languages will dominate the performance...

ichan-mb: Wow, there's so many information and insight with this discussion. Thank you, I'll try to fix the code using the fastest approach mention by @kusma

Source/Fuse.Storage/UserSettingsModule.uno
```diff
@@ -261,7 +299,12 @@ namespace Fuse.Storage
 		*/
 		object Clear(Context c, object[] args)
 		{
-			_userSetting.Clear();
+			if defined(Android)
+					AndroidUserSettingsImpl.Clear();
+				else if defined(iOS)
+					IOSUserSettingsImpl.Clear();
+				else
+					DesktopUserSettingsImpl.GetInstance().Clear();
```
kusma: I think most of this hunk accidentally got indented a bit too deep :)

Source/Fuse.Storage/Json.uno
```diff
@@ -0,0 +1,108 @@
...
+				get { return _array.Length; }
+			}
+
+			public object this[int index]
+			{
+				get { return _array[index]; }
+			}
+		}
+	}
+}
```
kusma: What makes this different from [Fuse.Json](https://github.com/fuse-open/fuselibs/blob/master/Source/Fuse.Common/Json.Parse.uno#L7)? Why can't we just reuse that one?

Source/Fuse.Storage/UserSettingsModule.uno
ichan-mb: ops! fix underway.

Source/Fuse.Storage/Json.uno
ichan-mb: Yeah, it is same, I created new Json.uno in `Fuse.Storage` because I cannot access the `parse` method in `Json.Parse.uno` in `Fuse.Common` because its modifier is `internal`.  I'm eager to change the method modifier in [Json.Parse.uno](https://github.com/fuse-open/fuselibs/blob/5f128ad3d6577b4b5faf8dc4c80452f76ea151ff/Source/Fuse.Common/Json.Parse.uno#L13) to `public` but I'm afraid. If it's OK to change its modifier, I will use that class instead and remove `Json.uno` on this PR

kusma: Right. I think it's just a matter of adding this project to the [InternalsVisibleTo-list in Fuse.Common.unoproj](https://github.com/fuse-open/fuselibs/blob/5f128ad3d6577b4b5faf8dc4c80452f76ea151ff/Source/Fuse.Common/Fuse.Common.unoproj#L5), and you should be able to use it without copying.

mortend: Adding to `InternalsVisibleTo` is a less drastic change because it's not changing public interfaces, but I'm for making the `Parse` method `public` instead, as @ichan-mb is eager to do. That seems easier and a bit cleaner to me, unless doing this ends up with more build errors because you'll need more internal stuff...

I see the `Parse` method is already documented and sounds like it may be useful for others to find...

In any case, this is up to you guys, just wanted to chime in with my opinion.

ichan-mb: Oh wow!, a new knowledge for me about `InternalVisibleTo` I noted that. Yeah I think I will make the `Parse` method `public` since I see `Stringify` method is `public` too [here](https://github.com/fuse-open/fuselibs/blob/5f128ad3d6577b4b5faf8dc4c80452f76ea151ff/Source/Fuse.Common/Json.uno#L15). It will make a complete pair of Json Function. Hope doing that doesn't make build errors

Nicero:
Very userful module! Please let me suggest you to add also _append_ and _remove_ for the various methods. In my opinion that would make the module even more effective.

Taking your example:
`userSettings.putArray('preferences', ['Technology', 'Cars', 'Foods']);`
add 'Music':
`userSettings.appendArray('preferences', ['Music']);`
result:
`preferences = ['Technology', 'Cars', 'Foods', 'Music'];`

Same with .`removeArray`.

This should be done also for `.putString` (_.appendString_, _.removeString_) and `.putObject` (_.appendObject_, _.removeObject_)

kusma:
> Very userful module! Please let me suggest you to add also _append_ and _remove_ for the various methods [...]

Patches welcome!

Nicero:
@kusma 
Sorry, but I'm not good enough nor in Javascript, nor in Uno or C to allow me to edit a single byte of code of Fuse libs!

kusma:
> @kusma
> Sorry, but I'm not good enough nor in Javascript, nor in Uno or C to allow me to edit a single byte of code of Fuse libs! If I could do it I would be happy to help Fuse development.

People tend to under-estimate themselves. Give it a try, maybe you'll figure it out? If not, ask for some help figuring it out! People tend to prioritize scratching their own itches over others...

PR #1314, opened 2019-10-14
iOS Push Notifications - IsRegisteredForRemoteNotifications Update

AndrewEQ:
- updated check to include current user setting in iOS 13

### DIFF
--- a/Source/Fuse.PushNotifications/iOS/Impl.uno
+++ b/Source/Fuse.PushNotifications/iOS/Impl.uno
@@ -163,10 +163,18 @@ namespace Fuse.PushNotifications
 		internal static bool IsRegisteredForRemoteNotifications()
 		@{
 			__block bool isReg = false;
+
+			__block bool isNotificationsSettingsEnabled = false;
+
 			dispatch_sync(dispatch_get_main_queue(), ^{
-			  isReg = [[UIApplication sharedApplication] isRegisteredForRemoteNotifications];
+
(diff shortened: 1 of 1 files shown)


Source/Fuse.PushNotifications/iOS/Impl.uno
```diff
@@ -163,10 +163,18 @@ namespace Fuse.PushNotifications
...
+
+			__block bool isNotificationsSettingsEnabled = false;
+
 			dispatch_sync(dispatch_get_main_queue(), ^{
-			  isReg = [[UIApplication sharedApplication] isRegisteredForRemoteNotifications];
+			  
+				isReg = [[UIApplication sharedApplication] isRegisteredForRemoteNotifications];
+				
+				[[UNUserNotificationCenter currentNotificationCenter] getNotificationSettingsWithCompletionHandler:^(UNNotificationSettings * _Nonnull settings) {
+				    isNotificationsSettingsEnabled = (settings.alertStyle != UNAlertStyleNone);
```
mortend: This line is indented by mixed tabs and spaces.

AndrewEQ: Its ok, I'll sort out the whitespace...

AndrewEQ:
Yeah, it basically allows us to know whether they've turned on push notifications; either by the user not enabling it when first asked or when they purposely turn it off in the app settings. When the user does this, we can change the interface accordingly to say something like "Hey! Notifications are currently off, if you would like to receive them, turn them back on!", and present a button to go the app settings.

PR #1321, opened 2019-11-04
Graceful trigger fail

AndrewEQ:
Added graceful fail to trigger when its not rooted else it crashes the app with the exception.

### DIFF
--- a/Source/Fuse.Triggers/Trigger.uno
+++ b/Source/Fuse.Triggers/Trigger.uno
@@ -342,7 +342,11 @@ namespace Fuse.Triggers
 			if (!_isStarted)
 			{
 				if (!IsRootingStarted)
-					throw new Exception( "Trigger started prior to being rooted: " + this );
+				{
+					Fuse.Diagnostics.UserError("Warning: Trigger.uno - Trigger started prior to being rooted: ", this );
+					return;
+				}
+
 				_isStarted = true;
 				UseContent = true;
 				PlayActions(TriggerWhen.Start);


Source/Fuse.Triggers/Trigger.uno
```diff
@@ -341,8 +341,11 @@ namespace Fuse.Triggers
 		{
 			if (!_isStarted)
 			{
-				if (!IsRootingStarted)
-					throw new Exception( "Trigger started prior to being rooted: " + this );
+				if (!IsRootingStarted){
+					debug_log("Warning: Trigger.uno - Trigger started prior to being rooted: " + this);
+					return;
+				}
```
kusma: Hmm, this just hides the symptoms... Why do you want to do that rather than fixing the actual problem?

mortend: I think we should call `Fuse.Diagnostics.UserWarning(message, object)` to report the warning instead.

https://github.com/fuse-open/fuselibs/search?q=UserWarning&unscoped_q=UserWarning

AndrewEQ: Tried fixing the problem but it randomly occurs, don't know why things are not rooted at the time of triggering the animation, nor how to make sure that it is rooted for sure before triggering the animation.

mortend: @kusma I don't think this have to be a fatal error, so reporting the problem via the Diagnostics API is maybe better than crashing the app.

AndrewEQ: ...switched it to Fuse.Diagnostics.UserError()

Source/Fuse.Triggers/Trigger.uno
```diff
@@ -341,8 +341,11 @@ namespace Fuse.Triggers
 		{
 			if (!_isStarted)
 			{
-				if (!IsRootingStarted)
-					throw new Exception( "Trigger started prior to being rooted: " + this );
+				if (!IsRootingStarted){
+					Fuse.Diagnostics.UserError("Warning: Trigger.uno - Trigger started prior to being rooted: ", this );
```
mortend:
```suggestion
					Fuse.Diagnostics.UserError("Trigger started prior to being rooted", this);
```

Source/Fuse.Triggers/Trigger.uno
```diff
@@ -341,8 +341,11 @@ namespace Fuse.Triggers
 		{
 			if (!_isStarted)
 			{
-				if (!IsRootingStarted)
-					throw new Exception( "Trigger started prior to being rooted: " + this );
+				if (!IsRootingStarted){
```
mortend: Code-style: `{` that should've been on its own line.

AndrewEQ: sorted

PR #1337, opened 2019-12-24
String function expressions

ichan-mb:
Added more string functions :
- `trim` function
- `indexOf` function
- `substring` function
- `replace` function
- `insert` function
- `split` function
- `startsWith` function
- `endsWith` function

This PR contains:
- [x] Changelog
- [ ] Documentation
- [ ] Tests


mortend:
Also, #1337! 🍾

PR #1404, opened 2021-03-23
Update ModuleInstance.uno

AndrewEQ:
Fixes an edge use case where navigation possibly reloads a page which tries to re-add a module again with the same key in the dictionary, thereby throwing a dictionary error which crashes the app: https://github.com/fuse-open/uno/blob/b9bd80402a775f3572e1bf84e32b28b2706de971/lib/UnoCore/Source/Uno/Collections/Dictionary.uno#L358


mortend (approved):

mortend:
- [x] https://travis-ci.org/github/fuse-open/fuselibs/builds/764069917

PR #1414, opened 2021-05-01
TextEditRenderer: enable horizontal scrolling only on lower version of android

ichan-mb:
This is a workaround fix for the issue: https://github.com/fuse-open/fuselibs/issues/1407. When we set alignment to the center or right on the recent version of Android, on lost focus the text won't visible.
related issue: 

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend (approved):

mortend:
- [x] https://travis-ci.org/github/fuse-open/fuselibs/builds/769182640

PR #1428, opened 2021-10-06
Single finger PanGesture

ichan-mb:
Currently, `PanGesture` (part of `TransformGesture`) to translate an `Element` only works with two-finger and that is not intuitive. With this PR we can do a `PanGesture` with both one or two fingers.

### Example:

```
<App>
    <JavaScript>
        module.exports = {
            url : 'https://images.unsplash.com/photo-1633493894963-3dfe3149fcdb?ixid=MnwxMjA3fDB8MHxwaG90by1wYWdlfHx8fGVufDB8fHx8&ixlib=rb-1.2.1&auto=format&fit=crop&w=987&q=80'
        }
    </JavaScript>
    <ClientPanel Color="White">
        <Panel HitTestMode="LocalBounds">
            <Image Url="{url}">
                <InteractiveTransform ux:Name="ImageTrans"/>
            </Image>
            <ZoomGesture Target="ImageTrans"/>
            <PanGesture Target="ImageTrans"/>
            <RotateGesture Target="ImageTrans"/>
        </Panel>
    </ClientPanel>
</App>
```

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Gestures/Internal/TwoFinger.uno
+++ b/Source/Fuse.Gestures/Internal/TwoFinger.uno
@@ -29,6 +29,23 @@ namespace Fuse.Gestures.Internal
 		}
 
 		bool _allowKeyControl = true;
+		bool _allowSingleTouch = false;
+
+		/* used for pan gesture to enable translating element using just one finger */
+		public bool SingleTouch
+		{
+			get
+			{
+				return _allowSingleTouch;
+			}
+			set
+			{
+				if (value != _allowSingleTouch)
+				{
+					_allowSingleTouch = value;
+				}
+			}
+		}
 
 		static readonly PropertyHandle _property = Fuse.Properties.CreateHandle();
 		static public TwoFinger Attach(Visual n)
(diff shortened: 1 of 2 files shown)


ichan-mb:
Hopefully, it will not cause any trouble, just for notice for anyone who read this PR, if you are getting into trouble because of this PR, there's a bool property in the `PanGesture`  to turn if this behavior, it called `SingleTouch` and set it to `false`

> 2.0 is right around the corner, so this is probably the last time we can allow changing defaults like this
Yes, I agree, and I am so happy about this upcoming new v2.0 🎉

mortend:
> Maybe would be better make Single Touch property false by default? In that case if someone need such behaviour, it can enable it.

Good point.

My gut feeling is with @ichan-mb I think, that one finger seems more intuitive in most cases.

Do we know about any existing apps that might break because of this, or concrete cases where one finger is a bad idea?

Anagrr:
> Good point.
> 
> My gut feeling is still with @ichan-mb I think, that one finger seems more intuitive in most cases.
> 
> Do we know about any existing apps that might break because of this, or concrete cases where one finger is a bad idea?

Ok, agree

PR #1431, opened 2021-11-03
Squashed commit of the global markdown adjustments

Anagrr:
This PR contains:
- [ ] Changelog
- [x] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Alerts/Triggers.uno
+++ b/Source/Fuse.Alerts/Triggers.uno
@@ -29,28 +29,29 @@ namespace Fuse
 	public delegate void AlertHandler(object sender, AlertArgs args);
 
 	/**
-		This is trigger action for showing native alert dialog with a single button. Only available on iOS or Android
+This is trigger action for showing native alert dialog with a single button. Only available on iOS or Android
 
-		## Example
+## Example
 
-			The following example shows how to use it:
-
-				<JavaScript>
-					module.exports = {
-						handler: function(data) {
-							if (data.buttonLabel == 'Yes'){
-								console.log("yes button clicked")
-							}
-						}
-					};
-				</JavaScript>
-				<Panel>
-					<Button Text="Display Alert" Alignment="Center">
-						<Clicked>
-							<ShowAlert Message="Hello world!" OkLabelButton="Yes" Handler="{handler}"/>
-						</Clicked>
-					</Button>
-				</Panel>
+The following example shows how to use it:
+'''javascript
+<JavaScript>
+	module.exports = {
+		handler: function(data) {
+			if (data.buttonLabel == 'Yes'){
+				console.log("yes button clicked")
+			}
+		}
+	};
+</JavaScript>
+<Panel>
+	<Button Text="Display Alert" Alignment="Center">
+		<Clicked>
+			<ShowAlert Message="Hello world!" OkLabelButton="Yes" Handler="{handler}"/>
+		</Clicked>
+	</Button>
+</Panel>
+```
 	*/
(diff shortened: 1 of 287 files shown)


Source/Fuse.Alerts/Triggers.uno
```diff
@@ -29,28 +29,29 @@ namespace Fuse
 	public delegate void AlertHandler(object sender, AlertArgs args);
 
 	/**
-		This is trigger action for showing native alert dialog with a single button. Only available on iOS or Android
+This is trigger action for showing native alert dialog with a single button. Only available on iOS or Android
```
mortend, line 32: This comment and other comments should be indented inside the `/**` with two tabs like before

Source/Fuse.Alerts/Triggers.uno
```diff
@@ -133,30 +134,31 @@ namespace Fuse
 	}
 
 	/**
-		This is trigger action for showing an ok/cancel dialog. Only available on iOS or Android
+This is trigger action for showing an ok/cancel dialog. Only available on iOS or Android
```
mortend, line 137: Two tabs

Source/Fuse.Android/StatusBarConfig.uno
```diff
@@ -31,27 +31,28 @@ namespace Fuse.Android
 	}
 
 	/**
-		Configures the appearance of the status bar on *Android*.
+Configures the appearance of the status bar on *Android*.
```
mortend, line 34: Two tabs

Source/Fuse.Auth/BiometricModule.uno
```diff
@@ -14,7 +14,7 @@ namespace Fuse
 		## Example
 
 			The following example shows how to use it:
-
+			```javascript
```
mortend, line 17: `xml` is better here I think?

Source/Fuse.Charting/PlotAxis.uno
```diff
@@ -153,7 +154,7 @@ namespace Fuse.Charting
 		# Example
 
 		The following example places labels on the y-axis of a bar chart.
-
+		```javascript
```
mortend, line 157: `xml`?

Source/Fuse.Charting/PlotBar.uno
```diff
@@ -27,7 +27,7 @@ namespace Fuse.Charting
 		# Example
 
 		The following example draws a red bar chart using the `PlotBar` element and @Rectangle
-
+		```javascript
```
mortend, line 30: `xml`?

mortend (changes requested):
Nice work :D

Source/Fuse.Auth/BiometricModule.uno
Anagrr: In these specific cases, JS is better for visual highlight, because we have more (or almost equal) code lines then the markup tags. In case of use `xml` we lost the js highlight

mortend: Cool, that's a good point. In that case it looks like writing `js` is more popular than `javascript`.

Anagrr: They are equal for the highlighting engine. We can use both

Source/Fuse.Alerts/Triggers.uno
mortend: This still needs two tabs.

mortend (commented):
A few more tabs please. :)

mortend (approved):

PR #1433, opened 2021-11-08
iOS: fix MetalAngle compile errors

### LINKED ISSUE #1432: Could not compile iOS project when -DMetal flag is set
**Describe the bug**
On the `Uno v2.0.0` has the ability to pass `-DMetal` to compile ios project using Metal backend instead of using OpenGL. but it gives errors and is unable to compile properly. I suspect we have to modify the `GraphicsView` class 

**To Reproduce**
Create a simple/empty project. and then compile with: `uno build -DMetal`

**Expected behavior**
The project should compile with no error 

**Screenshots**

image

mortend:
This fixes compile errors and enables building using -DMETAL.

The -DMETAL flag was introduced in https://github.com/fuse-open/uno/pull/396.

Apps are crashing so it looks like more work is needed to support
MetalANGLE in Fuselibs.

Closes #1432

### DIFF
--- a/Source/Fuse.Common/GraphicsWorker.uno
+++ b/Source/Fuse.Common/GraphicsWorker.uno
@@ -7,6 +7,7 @@ using Fuse;
 namespace Fuse
 {
 	[extern(ANDROID) Require("Source.Include", "Uno/Graphics/GLHelper.h")]
+	[extern(IOS && METAL) Require("Source.Include", "MetalANGLE/MGLKit.h")]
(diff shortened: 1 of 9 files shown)


ichan-mb:
With this PR, now we can compile the iOS project with `-DMetal` without any errors.. 👍

ichan-mb:
But when I test using simple app just showing a rectangle and run the application, it only show blank with a black screen

mortend:
Cool, fixing compile errors is a start. I will look more into the other problems later.

PR #1441, opened 2022-04-09
Fix memory leak on glass effect

ichan-mb:
Fixed memory leak where the framebuffer after the capture region operation is not released

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend (approved):
Nice :)

PR #1443, opened 2022-04-23
Change implementation for checking device orientation

ichan-mb:
Currently on iOS polling mechanism to check device orientation is using an old API that is deprecated and when I observe it takes more CPU usage. 

This PR it replaces the `[[UIApplication sharedApplication] statusBarOrientation]` API with the `[[UIDevice currentDevice] orientation]` and can reduce CPU usage up to 1% when the app is in an idle state. And this PR is also fixed some bad indentation

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend (approved):
Nice :)

PR #1445, opened 2022-05-15
added several newlines

Anagrr:
This should fix the markup issue in the docs. The docs for site need to be re-generated and published

This PR contains:
- [ ] Changelog
- [X] Documentation
- [ ] Tests


mortend (approved):
Thank you @Anagrr!

PR #1454, opened 2022-11-11
Device orientation for iOS 16 compatibility

ichan-mb:
According to the iOS 16 Release note:

"Attempting to set an orientation on UIDevice via setValue:forKey: isn’t supported and no longer works."

This is a fix using the new API introduced in iOS 16


mortend (approved):
Nice fix :)

PR #1459, opened 2023-01-15
add Snapping to the ScrollView

ichan-mb:
Scroll snapping allows you to lock the viewport to a certain location after a user has finished scrolling. You can use the `SnapAlignment` property to control the lock position. Three lock positions are available: ' Start `Center`, and `End`.

The setup that works now is with a `StackPanel` (Horizontal or Vertical)
```
<ScrollView LayoutMode="PreserveVisual" Height="400">
	<StackPanel>
		<Each Count="100" Reuse="Frame" >
			<Panel Color="#AAA">
				<Text Value="Data {= index() }"/>
			</Panel>
		</Each>
	</StackPanel>
	<ScrollViewSnap SnapAlignment="Center" />
</ScrollView>
```
It's required to use `LayoutMode="PreserveVisual"` on the `ScrollView` and the child element in the `StackPanel` has an equal size

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests

### DIFF
--- a/Source/Fuse.Controls.ScrollView/ScrollViewSnap.uno
+++ b/Source/Fuse.Controls.ScrollView/ScrollViewSnap.uno
@@ -1,0 +1,379 @@
+using Uno;
+using Uno.UX;
+
+using Fuse.Elements;
+using Fuse.Controls;
+
+namespace Fuse.Controls
+{
+	/**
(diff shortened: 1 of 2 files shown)


mortend (approved):
Cool :)

Looks good to me!

PR #1462, opened 2023-01-30
Fix desktop implementation of UserSettings Module

ichan-mb:
Currently broken on desktop (NET) because:
- There is no `Json.Stringify` for `Dictionary` object (the implementation is using `Dictionary` object and serialize it to json)
- Shared output file between Apps / Project, so it will overwrite the output between Apps/Project

This will fix those problems.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend:
The build is failing on AppVeyor, but it looks like a glitch/false negative.

Can you try force-pushing the branch to trigger a new build?

```
git commit --amend
git push --force
```

ichan-mb:
seems OK now..

PR #1496, opened 2024-06-15
Fix Android CameraView not showing on Android 13 or later because of permissions

ichan-mb:
This PR also fix memory leak on `<Video>` component reported on android Logcat because FileDescriptor is not closed.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend (approved):
Nice!

PR #1503, opened 2024-11-23
iOS MapView fix memory leaks 

ichan-mb:
Memory usage continue to reside even though MapView has been removed from the UI Tree, This is because something was missed when destroying the object.

This PR contains:
- [ ] Changelog
- [ ] Documentation
- [ ] Tests


mortend (approved):
Nice fix!

